PR #60, opened 2022-09-15
Update openapi generator to 6.1.0

jld3103:


jld3103:
upgrading the minor version breaks, let's build our own openapi generator that is not a complete mess

PR #104, opened 2022-10-24
ci: Fix F-Droid nightlies

jld3103:


eighthave:
This worked for me on GitHub Actions:

```yaml
      - name: fdroid nightly
        run: |
          sudo add-apt-repository ppa:fdroid/fdroidserver
          sudo apt-get update
          sudo apt-get install apksigner fdroidserver python3-pip --no-install-recommends
          sudo apt-get purge fdroidserver
          pip3 install https://gitlab.com/eighthave/fdroidserver/-/archive/index-v2-nightly/fdroidserver.tar.gz
          export DEBUG_KEYSTORE=${{ secrets.DEBUG_KEYSTORE }}
          which fdroid
          fdroid nightly --archive-older 10
```

provokateurin:
I previously had issues at some point with the plain ubuntu image because the apksigner was too old. I didn't really investigate what was going back then. I switched to uploading the build artifacts in the mean time and hope to just publish the app on F-Droid. So I don't want to put any work into this because it will be obsolete soon. Thanks for your heads up :)

eighthave:
I recommend using the `ubuntu-22.04` image, which is about to become the 
default.  Nightly builds are still useful even if you've already published to 
releases on f-droid.org.  They provide an easy way to test builds after each push.

Issue #188, opened 2023-01-30
Flutter web compatibility


adil192:
I'm getting this exception when creating a NextcloudClient on Flutter web:
<details>
<summary>Error: Unsupported operation: Platform._version</summary>

```python
Error: Unsupported operation: Platform._version
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 266:49      throw_
dart-sdk/lib/_internal/js_dev_runtime/patch/io_patch.dart 289:5                   _version
dart-sdk/lib/io/platform_impl.dart 122:32                                         get version
dart-sdk/lib/io/platform.dart 67:37                                               get _version
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/operations.dart 872:8   get
dart-sdk/lib/io/platform.dart 232:32                                              get version
dart-sdk/lib/_http/http_impl.dart 3769:26                                         _getHttpVersion
dart-sdk/lib/_http/http_impl.dart 2554:23                                         new
dart-sdk/lib/_http/http.dart 1240:14                                              new
packages/nextcloud/src/nextcloud.openapi.dart 109:38                              new
packages/nextcloud/src/client.dart 15:8                                           new
packages/saber/pages/nextcloud/login.dart 80:30                                   _tryLogin
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54                runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 123:5                _async
packages/saber/pages/nextcloud/login.dart 79:25                                   [_tryLogin]
packages/saber/components/nextcloud/login_group.dart 108:28                       _login
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 84:54                runBody
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 123:5                _async
packages/saber/components/nextcloud/login_group.dart 96:14                        [_login]
packages/flutter/src/material/ink_well.dart 1096:21                               handleTap
packages/flutter/src/gestures/recognizer.dart 253:24                              invokeCallback
packages/flutter/src/gestures/tap.dart 627:11                                     handleTapUp
packages/flutter/src/gestures/tap.dart 306:5                                      [_checkUp]
packages/flutter/src/gestures/tap.dart 239:7                                      handlePrimaryPointer
packages/flutter/src/gestures/recognizer.dart 615:9                               handleEvent
[...]

PR #190, opened 2023-02-06
Add web support

provokateurin:
Fixes https://github.com/provokateurin/nextcloud-neon/issues/188


Leptopoda (approved):
I can't really test this but LGTM

Leptopoda:
Aren't we still hitting the CORS problems?

In this case you might want to remove the `fixe #188` so the issue won't get closed automatically

provokateurin:
Yes we will run into CORS problems, but that is nothing Neon can fix. This is entirely on Nextcloud. I'd say the issue is closed, since it is possible from our side, just not from the server side.

provokateurin:
And it is possible to use some of the APIs, but they need to have CORS manually enabled which most apps don't do (but there are some that do it).

PR #201, opened 2023-02-27
fix _toDartName

Leptopoda:
When the name provided to `_toDartName` starts with a `_isNonAlphaNumericString` the `uppercaseFirstCharacter` is not respected.

https://github.com/provokateurin/nextcloud-neon/blob/443c23d167b9eea4d01c18e39d5eabd413276757/packages/dynamite/lib/src/openapi_builder.dart#L972-L978

here the problem is that the position of the first character is hard coded to the first character of the input name thus the `uppercaseFirstCharacter` is ignored if the input starts with a special character.

The method is inspirerd by the `_toCamelCase` as can be found on the built_value branch and in the built_value package.

### DIFF
--- a/packages/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/lib/src/openapi_builder.dart
@@ -975,37 +975,27 @@ class OpenAPIBuilder implements Builder {
 String _clientName(final String tag) => '${_toDartName(tag, uppercaseFirstCharacter: true)}Client';
 
 String _toDartName(
-  final String input, {
+  final String name, {
   final bool uppercaseFirstCharacter = false,
 }) {
-  final result = StringBuffer();
-
-  final parts = input.split('');
-  for (var i = 0; i < parts.length; i++) {
-    var char = parts[i];
-    final prevChar = i > 0 ? parts[i - 1] : null;
+  var result = '';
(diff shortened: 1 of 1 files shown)


packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1239,13 +1226,15 @@ TypeResult resolveObject(
...
+                      schema.required?.contains(propertyName),
+                      null,
+                    );
+
                     b
                       ..name = _toDartName(propertyName)
                       ..type = refer(
-                        _makeNullable(
-                          result.name,
-                          !(schema.required ?? []).contains(propertyName),
```
provokateurin, line 1247: This ! inverts the contains statement, but you removed it

provokateurin (changes requested):

packages/dynamite/lib/src/openapi_builder.dart
Leptopoda: Ok I must have been utterly incompetent when cherry picking this onto its separate branch.

This PR should only include the changes to `_toDartName` and not the other experimentations I did :)

provokateurin: Ah no problem!

PR #207, opened 2023-03-10
add vscode formatting settings

Leptopoda:
As I use both auto format on save and fairly often format using a keyboard shortcut it makes sense to configure the formatting options.

makes developing using vscode easier.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,4 @@
+{
+  "dart.lineLength": 120,
+  "yaml.format.singleQuote": true
+}


provokateurin:
You need to sign your commits.

Leptopoda:
They should signed. the gpg key just isn't added to GH yet as I changes some stuff around

But you're right I'll add it :)

Leptopoda:
Ok I don't know what happened.
Looks like when starting to use a subkey you have to delete the old key and re add it to GH even though the public key is still the same.

Nevertheless it's now marked as verified :)

PR #210, opened 2023-03-11
use https for initializing git submodules

Leptopoda:
This unifies the git submodule to always use https instead of git.

I think most people using ssh already have a redirect setup so we should settle on https as the default.


provokateurin:
Makes perfect sense. Thank you

PR #224, opened 2023-03-24
fix contentString with a base object contentSchema

Leptopoda:
Without this we could generate annotations like:
```dart
  @JsonKey(
    fromJson: String.fromJsonString,
    toJson: String.toJsonString,
  )
```

### DIFF
--- a/packages/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/lib/src/openapi_builder.dart
@@ -1265,13 +1265,28 @@ TypeResult resolveObject(
                       ..docs.addAll(_descriptionToDocs(propertySchema.description));
                     final hasDifferentName = _toDartName(propertyName) != propertyName;
                     final isContentString = propertySchema.isContentString;
-                    final isContentStringArray = isContentString && result is TypeResultList;
                     final hasExtraJsonKeyValues =
                         extraJsonKeyValues != null && extraJsonKeyValues.containsKey(propertyName);
-                    if (hasDifferentName || isContentString || isContentStringArray || hasExtraJsonKeyValues) {
+
+                    final arguments = <String, Expression>{
+                      if (hasDifferentName) ...{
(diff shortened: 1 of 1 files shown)


packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1268,7 +1268,9 @@ TypeResult resolveObject(
                     final isContentStringArray = isContentString && result is TypeResultList;
                     final hasExtraJsonKeyValues =
                         extraJsonKeyValues != null && extraJsonKeyValues.containsKey(propertyName);
-                    if (hasDifferentName || isContentString || isContentStringArray || hasExtraJsonKeyValues) {
+                    if (hasDifferentName ||
+                        result is! TypeResultBase && (isContentString || isContentStringArray) ||
```
provokateurin, line 1272: You need to put parenthesis around `result is! TypeResultBase && (isContentString || isContentStringArray)`

provokateurin, line 1272: Only checking for `TypeResultBase` is not enough, since `TypeResultEnum` is also just a string and `TypeResultList` is also valid in this case. I think you should invert the condition and only check against `TypeResultMap` and `TypeResultObject`.

provokateurin: Although `TypeResultMap` would have the same problems with `.fromJsonString` and `.toJsonString`, right? But it should definitely work with that... No idea how to fix that

provokateurin (changes requested):
I think it would make more sense to throw an exception instead of silently ignoring it

Leptopoda:
I think that throwing an exception is a bit hard.
the generated code does work it's just not that nice to not decode it.

maybe printing a message/warning is better in that case?

provokateurin:
Yeah we can also do that

provokateurin (approved):

PR #228, opened 2023-04-01
fix problems in the spec

### LINKED ISSUE #193: [BUG:] Nullability error in the generated spec UserStatusStatus.clearAt

Leptopoda:
fixes #193 

as requested in https://github.com/provokateurin/nextcloud-neon/pull/194#issuecomment-1492937977


specs/core.json
```diff
@@ -1311,4 +1317,4 @@
       }
     }
   }
-}
```
provokateurin, line 1314: Newline

specs/notifications.json
```diff
@@ -509,4 +499,4 @@
       }
     }
   }
-}
+}
```
provokateurin, line 502: Newline

specs/user_status.json
```diff
@@ -192,11 +191,7 @@
                 "$ref": "#/components/schemas/OCSMeta"
               },
               "data": {
-                "anyOf": [
-                  {
-                    "$ref": "#/components/schemas/Status"
-                  }
-                ]
+                "$ref": "#/components/schemas/Status"
```
provokateurin, line 194: This might fail, since an empty array can be returned here if the user has no status. Maybe just go for the `[oneOf: [array, Status]]` thing?

specs/user_status.json
```diff
@@ -245,11 +240,7 @@
                 "$ref": "#/components/schemas/OCSMeta"
               },
               "data": {
-                "anyOf": [
-                  {
-                    "$ref": "#/components/schemas/PublicStatus"
-                  }
-                ]
+                "$ref": "#/components/schemas/PublicStatus"
```
provokateurin, line 243: Same as above

specs/user_status.json
```diff
@@ -612,4 +601,4 @@
       }
     }
   }
-}
+}
```
provokateurin, line 604: Newline

provokateurin (changes requested):

specs/core.json
Leptopoda: Bad formatter settings :)

PR #229, opened 2023-04-02
Relax oneOf assertion

Leptopoda:
Also adding an assertion for anyOf :)

### DIFF
--- a/packages/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/lib/src/openapi_builder.dart
@@ -1549,7 +1549,10 @@ TypeResult resolveType(
                               ],
                             ],
                             if (schema.oneOf != null) ...[
-                              "assert([${fields.values.join(',')}].where((final x) => x != null).length == 1, 'Need oneOf for \$data');",
+                              "assert([${fields.values.join(',')}].where((final x) => x != null).length >= 1, 'Need oneOf for \$data');",
+                            ],
+                            if (schema.anyOf != null) ...[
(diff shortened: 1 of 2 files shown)


packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1549,7 +1549,10 @@ TypeResult resolveType(
                               ],
                             ],
                             if (schema.oneOf != null) ...[
-                              "assert([${fields.values.join(',')}].where((final x) => x != null).length == 1, 'Need oneOf for \$data');",
+                              "assert([${fields.values.join(',')}].where((final x) => x != null).length >= 1, 'Need oneOf for \$data');",
+                            ],
+                            if (schema.anyOf != null) ...[
+                              "assert([${fields.values.join(',')}].where((final x) => x != null).length >= 1,  'Need anyOf for \$data');",
```
provokateurin, line 1555: I thought I remembered that `anyOf` means any number of matches, even if it is zero, but I can't find it in the spec :woman_shrugging:

provokateurin (approved):

packages/dynamite/lib/src/openapi_builder.dart
Leptopoda: Nope any of has at least one match.

We already talked about that
https://github.com/provokateurin/nextcloud-neon/issues/193#issuecomment-1430138395

provokateurin: Alright, I was looking at https://spec.openapis.org/oas/latest.html which is the official spec (swagger is theoretically exactly the same, but they sometimes have better explanations or examples).

PR #237, opened 2023-04-07
Feature/built value split

Leptopoda:
the `nextcloud.openapi.g.dart` is not included as the build will currently fail :)

But at least it should be fairly easy to review


Leptopoda:
Yeah tests are abviously expected to fail

packages/dynamite/pubspec.yaml
```diff
@@ -6,6 +6,7 @@ environment:
 
 dependencies:
   build: ^2.3.1
+  built_value: ^8.4.4
```
provokateurin, line 9: This isn't used, is it?

provokateurin (changes requested):

packages/dynamite/pubspec.yaml
Leptopoda: I thought I need it for the build.yaml file but turns out I don't

Leptopoda:
Also I noticed the imports are missing in the generated file.

Leptopoda:
I also removed the json_serializable import as we are not depending on them anymore.

I didn't bother keeping it for now :)

packages/nextcloud/lib/src/nextcloud.openapi.g.dart
provokateurin, line 1: Maybe remove this file in a separate commit before doing any changes? Idk when you will add it back

provokateurin (approved):

packages/nextcloud/lib/src/nextcloud.openapi.g.dart
Leptopoda: Adding the imports lets the generator run them so It is back in my latest push.

It got auto removed as the generator failed without the imports.

PR #240, opened 2023-04-09
Refactor packages

provokateurin:


Leptopoda (approved):
LGTM

but could we put `neon_*` into it's own folder?
I plan to introduce a few `dynamite_*` packages (thinking of externalizing the ContentString<T> to make unit testing a bit easier and so we don't need to generate it every time) and I think we would clutter up `packages` quite a bit.

provokateurin:
I thought about that and I think it makes sense at some point. Do you have a suggestion on how to structure them (you can already add your `dynamite_*` packages to the structure)?

Leptopoda:
There aren't any packages yet but as mentioned on matrix I plan to add some common serializers that we can use (UnixTime, ISO8601Duration [I think builtValue already has one]) and externalizing the mentioned ContentString.

I'ts just for for the future and

provokateurin:
I understand, I was just asking if you had a better structure in mind that will also fit your package ideas.

Leptopoda:
I don't really know :(

But I know that I get quite confused having them all in the folder structure.

provokateurin:
Ok how about this:
```
dynamite/
  dynamite/ -> currently packages/dynamite/
  ... -> your new packages
neon/
  neon/ -> currently packages/neon/
  files/ -> currently packages/neon_files/
  ...
packages/
  app/
  ...
```
I would keep `dynamite` in `packages/` until you add your packages.
The package names will stay as they are right now.

Leptopoda:
sounds reasonable

provokateurin:
I wonder if mono_repo will still pick up those packages :thinking:

Leptopoda:
we could also do something like:
```
packages
├── app
├── dynamte
│   └── packages
│       └── contentString
└── neon
    └── packages
        ├── files
        └── news
```

While that might make sense for dynamite I don't really like it for neon as neon_files isn't really a package needed by neon itself.

provokateurin:
Hm I fear we run into too many sub directories. For dynamite it looks good, but for neon it's too much. How about:
```
packages
├── app
├── dynamite
│   ├── dynamite
│   └── dynamite_contentstring
└── neon
    ├── neon
    ├── neon_files
    └── neon_news
```
I would keep the prefix in order to keep the main libs the first entries, otherwise they are just somewhere in between and not very visible.

provokateurin:
Now the rebase fun begins :/

Leptopoda:
It's only 10 lines for me :P

provokateurin:
If it was only renames it would be easy, but the code split requires me to add new architecture parts so that the code can go into the sub packages but still load in the lib.

PR #241, opened 2023-04-09
Feature/built value split

Leptopoda:
This one is quite big


provokateurin:
Changes look good and most tests are passing now, but would it be possible to fix the tests in this PR already or is it too big and needs a separate PR?

Leptopoda:
I'm not home rn and would need to check to be sure but the only thing missing for the tests should be the someOfs stuff. It would just be https://github.com/provokateurin/nextcloud-neon/commit/3ea11e7c3d4122a3d15ee59024ce9e10e239b446

I can sqash them together if you want

PR #245, opened 2023-04-10
Feature/built value split

Leptopoda:
SomeOfs.

Tests are passing :tada: 
I also noticed I forgot to squash the default values into the last one so adding it here.

The ` as Object` for the deserialization is needed for a similar reason we needed to relax our assertions with the ofs. It's just some built_value stuff.


Leptopoda:
analyzer is complaining that neon isn't fixed yet

provokateurin (approved):

PR #246, opened 2023-04-11
migrate contentString

Leptopoda:
I've manually tested this with:
```json
"type": "string",
"application/json",
"contentSchema": {}
```
and
```json
"type": "string",
"application/json",
"contentSchema": {
  "type": "string"
}
```

and here our generated stuff for better readability:

```dart
abstract class ContentString<T> implements Built<ContentString<T>, ContentStringBuilder<T>> {
  const ContentString._();

  factory ContentString([void Function(ContentStringBuilder<T>)? b]) = _$ContentString<T>;

  /// decoded contentString
  T get content;
  static ContentString fromJson(Object json) => serializers.deserializeWith(serializer, json)!;
  Map<String, dynamic> toJson() => serializers.serializeWith(serializer, this) as Map<String, dynamic>;
  @BuiltValueSerializer(custom: true)
  static Serializer<ContentString> get serializer => _$ContentStringSerializer();
}

class _$ContentStringSerializer implements PrimitiveSerializer<ContentString<Object?>> {
  const _$ContentStringSerializer();

  @override
  final Iterable<Type> types = const [ContentString, _$ContentString];

  @override
  final String wireName = 'ContentString';

  @override
  Object serialize(
    Serializers serializers,
    ContentString<Object?> object, {
    FullType specifiedType = FullType.unspecified,
  }) {
    final isUnderspecified = specifiedType.isUnspecified || specifiedType.parameters.isEmpty;
    if (!isUnderspecified) serializers.expectBuilder(specifiedType);
    final parameterT = isUnderspecified ? FullType.object : specifiedType.parameters[0];

    final result = serializers.serialize(object.content, specifiedType: parameterT);

    return json.encode("$result");
  }

  @override
  ContentString<Object?> deserialize(
    Serializers serializers,
    Object? serialized, {
    FullType specifiedType = FullType.unspecified,
  }) {
    final isUnderspecified = specifiedType.isUnspecified || specifiedType.parameters.isEmpty;
    if (!isUnderspecified) serializers.expectBuilder(specifiedType);
    final parameterT = isUnderspecified ? FullType.object : specifiedType.parameters[0];

    final result = isUnderspecified
        ? ContentStringBuilder<Object?>()
        : serializers.newBuilder(specifiedType) as ContentStringBuilder<Object?>;

    result.content = serializers.deserialize(json.decode(serialized as String), specifiedType: parameterT);

    return result.build();
  }
}
```

I think we could also putt the above into it's own lib that gets exported by dynamite so we don't need to generate it all the time and even better unit test it easier even though we currently don't make use of it.
But that's for some other time.

EDIT:
I also removed the fromJsonString as I didn't see any value in it but we can also keep it.

### DIFF
--- a/packages/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/lib/src/openapi_builder.dart
@@ -770,10 +770,9 @@ class OpenAPIBuilder implements Builder {
                               if (nullable) {
                                 code.write('if (${_toDartName(parameter.name)} != null) {');
                               }
-                              final isPlainList = result is TypeResultList && !result.fromContentString;
                               final value = result.encode(
                                 result.serialize(_toDartName(parameter.name)),
-                                onlyChildren: isPlainList && parameter.in_ == 'query',
+                                onlyChildren: result is TypeResultList && parameter.in_ == 'query',
                                 // Objects inside the query always have to be interpreted in some way
                                 mimeType: 'application/json',
                               );
@@ -788,7 +787,7 @@ class OpenAPIBuilder implements Builder {
                                   break;
                                 case 'query':
                                   code.write(
-                                    "queryParameters['${parameter.name}${isPlainList ? '[]' : ''}'] = $value;",
+                                    "queryParameters['${parameter.name}${result is TypeResultList ? '[]' : ''}'] = $value;",
                                   );
                                   break;
                                 case 'header':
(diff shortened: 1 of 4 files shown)


Leptopoda:
Ok that was a stupid mistake.
We should be thinking of also enabling static analysis for the generated files.

provokateurin:
Is there something wrong with the PR or is it ready to review?

Leptopoda:
Is ready to review

packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -770,7 +770,7 @@ class OpenAPIBuilder implements Builder {
                               if (nullable) {
                                 code.write('if (${_toDartName(parameter.name)} != null) {');
                               }
-                              final isPlainList = result is TypeResultList && !result.fromContentString;
+                              final isPlainList = result is TypeResultList;
```
provokateurin, line 773: This only checks if it is a list, can't it be removed/inlined?

provokateurin:
Very nice cleanup!

PR #247, opened 2023-04-14
Add dynamites output to static analysis to prevent mistakes

provokateurin:


packages/nextcloud/lib/src/nextcloud.openapi.dart
```diff
@@ -35,8 +38,8 @@ class NextcloudResponse<T, U> {
   String toString() => 'NextcloudResponse(data: $data, headers: $headers)';
 }
 
-class _Response {
-  _Response(
+class RawResponse {
```
Leptopoda, line 41: If wanted we can still hide it from the export.

packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1862,3 +1877,16 @@ TypeResult resolveType(
 
   throw Exception('Can not convert OpenAPI type "${schema.toJson()}" to a Dart type');
 }
+
+// ignore: avoid_positional_boolean_parameters
+int sortRequired(final bool a, final bool b) {
```
Leptopoda, line 1882: This took me a bit to understand.

Maybe going forward we should add doc comments to those stuff so others can understand easier.

provokateurin: Yeah, I agree. Dart sort methods are pretty weird to use imo.

packages/nextcloud/lib/src/nextcloud.openapi.dart
provokateurin: Yeah, but I don't think it matters that much.

packages/dynamite/lib/src/openapi_builder.dart
Leptopoda: We don't need to explain sort methods but mentioning it is one is enough.

My head expanded the `sortRequired` into `isSortRequired` and I then expected it have a boolean return type.

Just mentioning it is a sort method for required params would have be it.

provokateurin: Is `sortRequiredElements` good? It sorts both parameters and properties, so it has to be agnostic.

Leptopoda (approved):

PR #248, opened 2023-04-14
header serialization

Leptopoda:
last big step :)

### DIFF
--- a/packages/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/lib/src/openapi_builder.dart
@@ -879,20 +879,6 @@ class OpenAPIBuilder implements Builder {
                                 if (response.headers != null) {
                                   final identifier =
                                       '${tag != null ? _toDartName(tag, uppercaseFirstCharacter: true) : null}${_toDartName(operationId, uppercaseFirstCharacter: true)}Headers';
-                                  final headerParseFunctions = <String, String>{};
-                                  for (final headerName in response.headers!.keys) {
-                                    final functionIdentifier = '_${_toDartName('${identifier}Parse$headerName')}';
-                                    headerParseFunctions[headerName] = functionIdentifier;
-                                    final result = resolveType(
-                                      spec,
-                                      state,
-                                      identifier,
-                                      response.headers![headerName]!.schema!,
-                                    );
-                                    output.add(
-                                      '${result.name} $functionIdentifier(final Map data, final String key) => ${result.deserialize(result.decode('data[key]'))};',
-                                    );
-                                  }
                                   final result = resolveType(
                                     spec,
                                     state,
@@ -905,13 +891,6 @@ class OpenAPIBuilder implements Builder {
                                         },
                                       },
                                     ),
-                                    extraJsonKeyValues: {
-                                      for (final headerName in response.headers!.keys) ...{
-                                        headerName.toLowerCase(): {
-                                          'readValue': headerParseFunctions[headerName]!,
-                                        },
-                                      },
-                                    },
                                   );
                                   headersType = result.name;
                                   headersValue = result.deserialize('response.headers');
@@ -1423,9 +1402,8 @@ TypeResult resolveObject(
   final OpenAPI spec,
   final State state,
   final String identifier,
-  final Schema schema, {
-  required final Map<String, Map<String, String>>? extraJsonKeyValues,
-}) {
+  final Schema schema,
+) {
   if (!state.resolvedTypes.contains('${state.prefix}$identifier')) {
     state.resolvedTypes.add('${state.prefix}$identifier');
     state.registeredJsonObjects.add('${state.prefix}$identifier');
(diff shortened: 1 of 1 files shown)


provokateurin:
Can you generate some example code and put it here? I just want to see the output. It's too bad that response headers are not used anywhere right now (only in some experimental branches I have).

Leptopoda:
Ok diggin into it I think a rebase might have gone bad :(

I mean https://github.com/provokateurin/nextcloud-neon/pull/248/files#diff-1d39e7cbcf0dadc13301db10b6a3240d340ac9a017e4489d842320e9718514ebR1565 is never true.

Now looking into the old code it's not doing anything with the `json_serializable` `readValue` property (i't just doing a plain lookup at the given key and decodes the data into the needed type. Nothing json_serializable wouldn't do anyways).

If yes we only require the `wireName` property (similar to [`JsonKey.name`](https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/name.html).

Am I right or just overlooking something? I thought about it a bit when I wrote the code but don't see any benefit now :thinking:

provokateurin:
I think you are right. We should start unit testing the output of dynamite or dynamite itself :/

Leptopoda:
Ok I've simplified it as mentioned above although I can't test it :(

---

Just adding a header to the first operation I found:

```json
    "/index.php/login/v2": {
      "post": {
        "operationId": "init-login-flow",
        "tags": [
          "core"
        ],
        "security": [],
        "responses": {
          "200": {
            "description": "Login flow init",
            "headers": {
              "X-RateLimit-Limit": {
                "schema": {
                  "$ref": "#/components/schemas/LoginFlowInit"
                }
              }
            },
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/LoginFlowInit"
                }
              }
            }
          }
        }
      }
    }
```
Which will generate something like:
```dart
  Future<NextcloudResponse<NextcloudCoreLoginFlowInit, NextcloudCoreInitLoginFlowHeaders>> initLoginFlow() async {
    var path = '/index.php/login/v2';
    final queryParameters = <String, dynamic>{};
    final headers = <String, String>{
      'Accept': 'application/json',
    };
    Uint8List? body;
    final response = await rootClient.doRequest(
      'post',
      Uri(path: path, queryParameters: queryParameters.isNotEmpty ? queryParameters : null).toString(),
      headers,
      body,
    );
    if (response.statusCode == 200) {
      return NextcloudResponse<NextcloudCoreLoginFlowInit, NextcloudCoreInitLoginFlowHeaders>(
        NextcloudCoreLoginFlowInit.fromJson(json.decode(utf8.decode(response.body) as String) as Object),
        NextcloudCoreInitLoginFlowHeaders.fromJson(response.headers as Object),
      );
    }
    throw NextcloudApiException.fromResponse(response); // coverage:ignore-line
  }

//...

abstract class NextcloudCoreInitLoginFlowHeaders
    implements Built<NextcloudCoreInitLoginFlowHeaders, NextcloudCoreInitLoginFlowHeadersBuilder> {
  const NextcloudCoreInitLoginFlowHeaders._();

  factory NextcloudCoreInitLoginFlowHeaders([void Function(NextcloudCoreInitLoginFlowHeadersBuilder)? b]) =
      _$NextcloudCoreInitLoginFlowHeaders;

  static NextcloudCoreInitLoginFlowHeaders fromJson(Object json) => serializers.deserializeWith(serializer, json)!;
  Map<String, dynamic> toJson() => serializers.serializeWith(serializer, this) as Map<String, dynamic>;
  @BuiltValueField(wireName: 'x-ratelimit-limit')
  NextcloudCoreLoginFlowInit? get xRatelimitLimit;
  static Serializer<NextcloudCoreInitLoginFlowHeaders> get serializer => _$nextcloudCoreInitLoginFlowHeadersSerializer;
}
```

and a serializer 

```dart
class _$NextcloudCoreInitLoginFlowHeadersSerializer implements StructuredSerializer<NextcloudCoreInitLoginFlowHeaders> {
  @override
  final Iterable<Type> types = const [NextcloudCoreInitLoginFlowHeaders, _$NextcloudCoreInitLoginFlowHeaders];
  @override
  final String wireName = 'NextcloudCoreInitLoginFlowHeaders';

  @override
  Iterable<Object?> serialize(Serializers serializers, NextcloudCoreInitLoginFlowHeaders object,
      {FullType specifiedType = FullType.unspecified}) {
    final result = <Object?>[];
    Object? value;
    value = object.xRatelimitLimit;
    if (value != null) {
      result
        ..add('x-ratelimit-limit')
        ..add(serializers.serialize(value, specifiedType: const FullType(NextcloudCoreLoginFlowInit)));
    }
    return result;
  }

  @override
  NextcloudCoreInitLoginFlowHeaders deserialize(Serializers serializers, Iterable<Object?> serialized,
      {FullType specifiedType = FullType.unspecified}) {
    final result = NextcloudCoreInitLoginFlowHeadersBuilder();

    final iterator = serialized.iterator;
    while (iterator.moveNext()) {
      final key = iterator.current! as String;
      iterator.moveNext();
      final Object? value = iterator.current;
      switch (key) {
        case 'x-ratelimit-limit':
          result.xRatelimitLimit.replace(serializers.deserialize(value,
              specifiedType: const FullType(NextcloudCoreLoginFlowInit))! as NextcloudCoreLoginFlowInit);
          break;
      }
    }

    return result.build();
  }
}
```

wich basically just decodes the value at the given headername.

provokateurin:
The output looks good

provokateurin (approved):

provokateurin:
I really don't know why it was so complicated before haha

Leptopoda:
Can you give me an example response we would try to deserialize?

This would make implementing way easier.

provokateurin:
Yeah it's basically a content string, so we could internally treat it that way. Take `X-Chat-Last-Common-Read` from https://github.com/nextcloud/spreed/blob/master/lib/Controller/ChatController.php for example. It contains an integer and needs to be treated that way. See https://github.com/provokateurin/nextcloud-neon/blob/feature/spreed/specs/spreed.json for reference.

Leptopoda:
Thanks. I'll have a look :)

I'ts probably easier for now just to restore my old approach (custom serializer) than trying to leverage ContentString but we'll see.

provokateurin:
Whatever works best right now. It's not super important anyway, since headers are only used in that feature branch and I'm not working on it right now.

PR #252, opened 2023-04-17
migrate neon

Leptopoda:
last thing needed for us to land built_value.

I don't know if it makes sense to merge as we're gonna need to rebase it anyways :woman_shrugging:.

I've just changed the Types from `List` to `Iterable`. We could also keep the types and just cast it to a list (`.toList()`)


Leptopoda:
Also both cases don't require other dependencies as raised in:
https://github.com/provokateurin/nextcloud-neon/pull/194#issuecomment-1435369096

provokateurin (approved):

PR #254, opened 2023-04-17
migrate neon fix

Leptopoda:
Just tried to run neon and found #252 to be fairly bad.

I shouldn't have changed it last minute without testing (passing analysis != running app) :sweat_smile:


Leptopoda:
I've rebased this branch onto main and once this is in I'll sqash the two migrate neon commits and push it to #194

Issue #12, opened 2022-07-06
Cache files responses


provokateurin:
Right now the WebDAV API calls are not cached because they don't go through the OpenAPI generated clients. This needs some changes in the Nextcloud package to make it possible to serialize the responses.

Issue #13, opened 2022-07-06
Implement autostart


provokateurin:

provokateurin:
With https://pub.dev/packages/launch_at_startup it should be easy to implement

Issue #11, opened 2022-07-06
Notification for new unread articles


provokateurin:
This should probably implemented using a general approach where every app can add recurring tasks. On mobile platforms the scheduler behind it should be implemented using the native platform on mobile and on desktop just as a simple periodic timer.

Issue #10, opened 2022-07-06
Implement user status heartbeat


jld3103:
Depends on https://github.com/nextcloud/server/pull/32646 which should be released in Nextcloud 25.

PR #262, opened 2023-04-19
Make WebDav serializable

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/12

This drops support for custom props.


provokateurin:
There are some props that I haven't got working yet, so I left them out for now. Nothing in this repo relies on those props, so it's fine to drop them for now.

Leptopoda (changes requested):
First of all I need to clarify that I've never worked with webdav and never used the old api we exposed. I'm purely reviewing this coming from the dart side.

1. Is there a reason why WebDavFile has been removed from the client into neon? This abstraction might also be useful for users just wanting to use the dart lib without the entire neon stuff.

2. Have you tried running the changes in neon? while they work the performance is really bad :( Opening a folder will yield in frame times of 2.7sec wich is a bit higher than the maximum 16ms we can afford for 60fps :upside_down_face: Looks like mostly being the path getter of the file.
https://github.com/provokateurin/nextcloud-neon/pull/262/files#diff-7da1b4f44ceb1c98e1d775bf27739022a8ec07f27c388b246451f3f7b1291eabR32

In generall the entire WebDavFile interface isn't cached so the expensive code might even get run multiple times :/

It should be fairly reproducible but I've attached a profiling export
[dart_devtools_2023-04-22_12 26 20.249.json.txt](https://github.com/provokateurin/nextcloud-neon/files/11301131/dart_devtools_2023-04-22_12.26.20.249.json.txt)

provokateurin:
1. Yeah, I'll add it back and also try to add all the other props that weren't previously exposed in the WebDavFile
2. I ran it and it worked fine, but I didn't notice those lags because my emulator just lags by itself. Caching the path and props should be doable, even the individual prop values.

Thanks for the review!

Leptopoda:
I just digged a bit deeper and must admit that this is a worst case scenario as the folder I was opening contains around 670 files (I should really set an auto delete on my `.Contacts-Backup`).

I can also repro the same on master so we should probably deserialize this async (probably for another commit).
At least the times are roughly halved so caching might still make sense.

provokateurin:
Right, that's a lot of files. I only tested it on the dev docker container that has like 10-20 files. Yeah we should offload the serialization parts, I think using flutter's `compute` would be the right thing to use.

Leptopoda:
While this would help with reducing stuttering/blocked frames we would still need to wait the full second to open the folder :(

I was thinking of maybe doing this in batches as we scroll or just a normal stream (although the stream would again decode everything even if we don't scroll to the bottom so batches might be better).

provokateurin:
Hm that sounds complex, but like a good idea. I'd say that is something we can implement later, since the problem existed before this PR.

provokateurin:
I was able to reproduce your performance problems, although they weren't quite as dramatic running them on my linux machine. Anyway, caching the props makes it go away! I'll try to also add the offloading of the serialization to increase it even more.

provokateurin:
Using `compute` now, but I couldn't notice any difference.

Leptopoda (commented):
With the caching I see frame times of around 70ms for the same folder wich should be acceptable for now. Looks like they are even faster than the current master :+1: 

> although they weren't quite as dramatic running them on my linux machine.

I just ran it on my machine aswell and dam are those render times fast. I don't think it's even comparable :)

Leptopoda (approved):

provokateurin:
I'm going to add back the old props to WebDavFile and also add more of the props that weren't exposed until now.

Leptopoda:
https://github.com/provokateurin/nextcloud-neon/blob/fee5ad7beeac3588942eb5d89eaa9334e9fd8cb7/packages/nextcloud/lib/src/webdav/webdav.g.dart#L97
```
E/SurfaceSyncer(29860): Failed to find sync for id=2
I/flutter (29860): Null check operator used on a null value
I/flutter (29860): #0      _$WebDavResponseFromXmlElement (package:nextcloud/src/webdav/webdav.g.dart:97)
I/flutter (29860): #1      new WebDavResponse.fromXmlElement (package:nextcloud/src/webdav/webdav.dart:48)
I/flutter (29860): #2      _$WebDavMultistatusFromXmlElement.<anonymous closure> (package:nextcloud/src/webdav/webdav.g.dart:29)
I/flutter (29860): #3      MappedIterator.moveNext (dart:_internal/iterable.dart:393)
I/flutter (29860): #4      new _GrowableList._ofOther (dart:core-patch/growable_array.dart:202)
I/flutter (29860): #5      new _GrowableList.of (dart:core-patch/growable_array.dart:152)
I/flutter (29860): #6      new List.of (dart:core-patch/array_patch.dart:52)
I/flutter (29860): #7      Iterable.toList (dart:core/iterable.dart:470)
I/flutter (29860): #8      _$WebDavMultistatusFromXmlElement (package:nextcloud/src/webdav/webdav.g.dart:29)
I/flutter (29860): #9      new WebDavMultistatus.fromXmlElement (package:nextcloud/src/webdav/webdav.dart:34)
I/flutter (29860): #10     RequestManager.wrapWebDav.<anonymous closure> (package:neon/src/utils/request_manager.dart:45)
I/flutter (29860): #11     compute.<anonymous closure> (package:flutter/src/foundation/_isolates_io.dart:18)
I/flutter (29860): #12     _RemoteRunner._run (dart:isolate:954)
I/flutter (29860): #13     _delayEntrypointInvocation.<anonymous closure> (dart:isolate-patch/isolate_patch.dart:300)
I/flutter (29860): #14     _RawReceivePort._handleMessage (dart:isolate-patch/isolate_patch.dart:192)
I/flutter (29860)
````

provokateurin:
Oh interesting, I didn't know that is possible. Do you have an idea what file caused this or can you dump the raw xml response?

provokateurin:
Fixed it, responses without an href are filtered out when converting them to WebDavFiles. I would still be interested what caused this.

Leptopoda:
> Fixed it, responses without an href are filtered out when converting them to WebDavFiles. I would still be interested what caused this.

I have no Idea and was only able to repro once (while sadly no debugger was attached).

PR #269, opened 2023-04-21
Fix/built value/someofs types

Leptopoda:
The ContentString might get a bit bigger so I'll submit this separately.
This PR fixes someOfs discriminators and patches the built_value_generator to not be suffering the bug when generating with numbers.

Also do I have write acess to your built_value branch?
If so I'd force push a rebase onto the current master as can be found on my current built_value branch.


provokateurin:
I think you have access, GH really doesn't display me any permissions 🤷‍♀️ . Feel free to force push the branch

PR #276, opened 2023-04-23
sortbox: improve performance

Leptopoda:
https://github.com/Leptopoda/nextcloud-neon/blob/0cdd0c15da37cf30eb1dec556b31fd9e5f3730ea/packages/sort_box/lib/sort_box.dart#L24

Ideas for a better variable name are welcome :)

https://github.com/Leptopoda/nextcloud-neon/blob/0cdd0c15da37cf30eb1dec556b31fd9e5f3730ea/packages/sort_box/lib/sort_box.dart#L9

should we make the comparablegetter it's own typedef?
I don't think it's worth it but it went through my head so just asking.

---

I removed  `preventSecondarySort` from the api as I think it was only to make the recursiveness working but we can obviously still support it.

### DIFF
--- a/packages/sort_box/.gitignore
+++ b/packages/sort_box/.gitignore
@@ -28,3 +28,5 @@ migrate_working_dir/
 .dart_tool/
 .packages
 build/
+
+coverage/
--- a/packages/sort_box/lib/sort_box.dart
+++ b/packages/sort_box/lib/sort_box.dart
@@ -1,68 +1,44 @@
 // ignore_for_file: public_member_api_docs
 
+typedef ComparableGetter<T> = Comparable Function(T);
+
 class SortBox<T extends Enum, R> {
   SortBox(
     this._properties,
     this._secondaryBoxes,
   );
 
-  final Map<T, Comparable Function(R)> _properties;
+  final Map<T, ComparableGetter<R>> _properties;
   final Map<T, Box<T>> _secondaryBoxes;
 
-  List<R> sort(
-    final List<R> input,
-    final Box<T> box, {
-    final bool preventSecondarySort = false,
-  }) {
+  List<R> sort(final List<R> input, final Box<T> box) {
     if (input.length <= 1) {
       return input;
     }
 
-    final counters = <int, List<R>>{};
     final comparableGetter = _properties[box.property]!;
-
-    for (final item1 in input) {
-      var counter = 0;
-      for (final item2 in input) {
-        if (item2 != item1) {
-          final comparable1 = comparableGetter(item1);
-          final comparable2 = comparableGetter(item2);
-          counter += box.order == SortBoxOrder.descending
-              ? comparable1.compareTo(comparable2)
-              : comparable2.compareTo(comparable1);
-        }
-      }
-      if (counters[counter] == null) {
-        counters[counter] = [];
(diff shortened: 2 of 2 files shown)


provokateurin:
> should we make the comparablegetter it's own typedef?

Let's do it.

> I removed preventSecondarySort from the api as I think it was only to make the recursiveness working but we can obviously still support it.

Exactly, no need to keep it.

provokateurin (approved):

provokateurin:
Ah actually I think it was recursive because it would be able to sort on an infinite amount of levels, but tbh that is a pretty niche use-case and better performance is definitely worth more than that.

Leptopoda:
We could also make this work:
We would just call `_compare` recursively instead of the entire sort.

https://github.com/provokateurin/nextcloud-neon/blob/0cdd0c15da37cf30eb1dec556b31fd9e5f3730ea/packages/sort_box/lib/sort_box.dart#L35-L41

I mean we can chain as many `_compare` as we want. If they are equal go to the next one.

I'll have a look :)

provokateurin:
I really don't think we need it, but if you want to give it a shot then go for it.

Leptopoda:
Yeah in that case I'd leave it until we really need to.

makin compare recursive means we would probably need to move 
https://github.com/provokateurin/nextcloud-neon/blob/0cdd0c15da37cf30eb1dec556b31fd9e5f3730ea/packages/sort_box/lib/sort_box.dart#L17-L19

into it. They should all be in constant time but why should we do it on every compare.

provokateurin:
Ok, so you only want to add the typedef and then it's ready?

Leptopoda:
Still thinking about this...
I can't wrap my head around removing potentially useful features. I made a POC adding it in again (it works) but how is this ` be able to sort on an infinite amount of levels` currently possible. Aren't we always stopping at the second level by having `preventSecondarySort` set?

```dart
  List<R> sort(final List<R> input, final Box<T> box) {
    if (input.length <= 1) {
      return input;
    }

    final sorted = input..sort((final item1, final item2) => _compare(item1, item2, box));

    return sorted;
  }

  int _compare(
    final R item1,
    final R item2,
    final Box<T>? box, {
    final bool preventSecondarySort = false,
  }) {
    // Stop sorting when no adinitonal property is specified.
    if (box == null) {
      return 0;
    }

    final comparableGetter = _properties[box.property]!;

    final comparable1 = comparableGetter(item1);
    final comparable2 = comparableGetter(item2);

    final order =
        box.order == SortBoxOrder.ascending ? comparable1.compareTo(comparable2) : comparable2.compareTo(comparable1);

    final secondaryBox = _secondaryBoxes[box.property];

    if (order == 0 && !preventSecondarySort) {
      return _compare(item1, item2, secondaryBox, preventSecondarySort: true);
    }

    return order;
  }
```

provokateurin:
Oh hm you're right :sweat_smile:, so you don't actually remove any features...

Leptopoda:
now that we have the code keep or sweep?
We could add a maxLevels as an int and break when reaching it?

I mean we now have both versions xD

provokateurin:
No just keep the simple two-level sort, I think it's bad to have code that won't be used much.

PR #287, opened 2023-04-25
Fix nullable vs. required

provokateurin:
Previously `required` was used like `!nullable`. This is wrong behaviour, because a property could be required, but the value could be nullable. In Dart this makes no difference, but we have to still support it (latest generated specs do this correctly).


packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1180,7 +1196,18 @@ String _makeNullable(final String type, final bool nullable) => nullable && type
 
 String _toFieldName(final String dartName, final String type) => dartName == type ? '\$$dartName' : dartName;
 
-bool _isParameterNullable(final bool? required, final dynamic default_) => !(required ?? false) && default_ == null;
+bool _isDartParameterNullable(
+  final bool? required,
+  final bool? nullable,
+  final dynamic default_,
+) =>
+    (!(required ?? false) && default_ == null) || (nullable ?? false);
```
Leptopoda, line 1204: Could we call `_isDartParameterRequired` here?

packages/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1242,8 +1269,14 @@ TypeResult resolveObject(
           final sortedParameterKeys = schema.properties!.keys.toList()
             ..sort(
               (final a, final b) => sortRequiredElements(
-                (schema.required ?? []).contains(a) && schema.properties![a]!.default_ == null,
-                (schema.required ?? []).contains(b) && schema.properties![b]!.default_ == null,
+                _isDartParameterRequired(
+                  (schema.required ?? []).contains(a),
```
Leptopoda, line 1273: the receiver can take a nullable value so `schema.required?.contains(a)` might be cleaner.

Also everywhere else this is done

Leptopoda (changes requested):

Issue #284, opened 2023-04-24
LoginFlow QR code support


Leptopoda:
It has been a request for the CookBook app for a while and I made a working [POC](https://github.com/Teifun2/nextcloud-cookbook-flutter/pull/193/commits/feed63ea28dc46277ab8874739ba3504aa57f3ed) that works.

When we migrate to neon this feature would be lost so Ideally we should support the functionality here as well.

We basically only need to scan the code and decode it with something like 
```dart
Map<String, String> parseNCLoginQR(Uri uri) {
  return uri.path.split('&').map((e) {
    final parts = e.split(':');
    final key = parts[0].replaceFirst(RegExp('/'), '');
    parts.removeAt(0);
    return {key: parts.join(':')};
  }).fold({}, (p, e) => p..addAll(e));
}
```

I haven't found official documentation about the used format but it's fairly self explanatory so I made the above code.

I used the `qr_code_scanner` package that currently only supports iOS and Android so linux is out but I'd argue this is not a problem.

I could clean up my POC and contribute it into here. Any Ideas?

provokateurin:
Sounds good! Does the QR code already contain the server URL?

Leptopoda:
Yes it does. It looks like:
`nc://login/user:$USERID&password:$APPPASSWORD&server:$URL`

The Urls also does contain the schmeme thus a split at `:` can't be done that easily.
The snippet above does decode it into:
```dart
{
 'user': 'admin',
 'password': 'superSecret',
 'server': 'https://example.com',
}
```

provokateurin:
So it would also make sense to register the android app for the `nc://` scheme, so that it would trigger when the user scans the QR code in their camera app, right?

Leptopoda:
Haven't thought about this :eyes: 
But it should be possible.

Either my OS doesn't support so or the official NC app doesn't do that for some reasons (I'd need to look at the source).

provokateurin:
Do you want to implement this? Then please assign yourself :)

Leptopoda:
I coul'd do it but that would slip to next week (or even a bit further).
I'll first finish the ContentString/built_value stuff and don't have Time this weekend. Also I'd like to tackle my experiments with the ListView performance.

So feel free to do it if you have the time otherwise you can assign me.

provokateurin:
Ok then I'll do it. I was interested in it, but I didn't want to take it away from you :)

provokateurin:
[...]

PR #291, opened 2023-04-26
Add QR-Code login

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/284


Leptopoda:
Looks like you're using `mobile_scanner` just out of curiosity does this mean we are depending on play services for this to work or do you plan using the bundled version?

Also is the bundled version going to be approved by f-droid? I couldn't find anything about it and I somehow doubt that everything of mlkit is open source :woman_shrugging:

provokateurin:
Oh I didn't think about that. I looked at qr_code_scanner but it said it's deprecated, that's why I went with this lib. If it uses proprietary stuff then that's a no-go and we need to find something else.

Leptopoda:
Ok according to google:
https://developers.google.com/ml-kit/vision/barcode-scanning/code-scanner
> The implementation resides entirely within Google Play services, so there is no significant impact to the size of your app. 
Looks like it's proprietary google stuff :(
Searching the fdroid forum revealed an issue of another flutter dev failing with the the `mobile_scanner` package.

> I looked at qr_code_scanner but it said it's deprecated

I didn't noticed that. I only goofed around with it a couple of times and it seemed quite ok to use.

provokateurin:
Meh ok, then maybe qr_code_scanner it is. I wonder if there is pure dart library for scanning images for qr codes, because then we could just use the camera plugin :thinking: .

provokateurin:
https://pub.dev/packages/qr_code_dart_scan Seems to do that and it uses the camera plugin, but it also hasn't been updated in a while. That library depends on https://github.com/shirne/zxing-dart for scanning. Then there is https://pub.dev/packages/zxing_scanner advertised by the zxing dart implementation.

I think in general it's nice to use a dart implementation because it won't bloat the app size as much and is easier for cross platform. I would go with the https://pub.dev/packages/zxing_scanner library since that seems to be the most up-to-date and developed by the same person that develops the zxing dart implementation.

provokateurin:
zxing_scanner works, but the resolutions is really bad somehow and the speed is also not great. Let's go with qr_code_scanner for now :/

provokateurin:
Somehow qr_code_scanner never gives me a result, same as this report: https://github.com/juliuscanute/qr_code_scanner/issues/356. You said your implementation works, can you check if this PR also works? Then it might just be my emulator or something else that breaks it.

Leptopoda:
What was the reason not to use the mentioned zxing_dart package?

provokateurin:
I think I will go with https://pub.dev/packages/flutter_zxing as it uses the same zxing implementation for all platforms. Hopefully it's smooth and fast, but I'll have to test that first.

provokateurin:
Will implement it on top of https://github.com/provokateurin/nextcloud-neon/pull/452

packages/neon/neon/lib/src/pages/login_qrcode.dart
```diff
@@ -0,0 +1,105 @@
...
+            password: widget.password,
+          ),
+        ),
+      );
+    });
+  }
+
+  @override
+  Widget build(final BuildContext context) => const SizedBox();
+}
```
provokateurin, line 94: I couldn't find a better way than having this intermediate page, because in the router it's only possible to return a location and not execute any logic. @Leptopoda do you know if there is any better method for doing this?

provokateurin:
I couldn't get it working with my QR scanner app (https://f-droid.org/en/packages/com.secuso.privacyFriendlyCodeScanner/) but when triggering the intent via the command line it works. I assume my QR scanner just forces to open in the URL in the browser.

packages/neon/neon/lib/src/pages/login_qrcode.dart
```diff
@@ -0,0 +1,105 @@
...
+              }
+              final match = parseLoginQrcodeURL(url);
+              if (match == null) {
+                throw InvalidQrcodeException();
+              }
+              await processLoginQrcode(context, match);
+            } catch (e, s) {
+              if (_lastErrorURL != url) {
+                debugPrint(e.toString());
+                debugPrint(s.toString());
```
Leptopoda, line 47: `debugPrintStack()` ?

packages/neon/neon/lib/src/utils/login_qrcode.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:meta/meta.dart';
+
+final _loginQrcodeUrlRegex = RegExp(r'^nc://login/user:(.*)&password:(.*)&server:(.*)$');
+final _loginQrcodePathRegex = RegExp(r'^/user:(.*)&password:(.*)&server:(.*)$');
```
Leptopoda, line 4: when `parseLoginQrcodeURL` is moved into the class those should probably too

packages/neon/neon/lib/src/pages/login_qrcode.dart
Leptopoda: You can add a redirect callback on a per route basis.
This would be the `LoginQrcodeIntermediateRoute`:

```dart
@immutable
class LoginQrcodeIntermediateRoute extends GoRouteData {
  const LoginQrcodeIntermediateRoute({
    required this.serverURL,
    required this.loginName,
    required this.password,
  });

  final String serverURL;
  final String loginName;
  final String password;

  @override
  Widget build(final BuildContext context, final GoRouterState state) => const SizedBox();

  @override
  FutureOr<String?> redirect(final BuildContext context, final GoRouterState state) async {
    final result = await LoginCheckServerStatusRoute(serverURL: serverURL).push<bool>(context);
    return LoginCheckAccountRoute(
      serverURL: serverURL,
      loginName: loginName,
      password: password,
    ).location;
  }
}
```

Note that the checkstatus page won't work like this and needs something different :)
I just copy pasted it from the widget.

packages/neon/neon/lib/src/pages/login_qrcode.dart
```diff
@@ -0,0 +1,105 @@
...
+        ),
+      );
+    });
+  }
+
+  @override
+  Widget build(final BuildContext context) => const SizedBox();
+}
+
+Future processLoginQrcode(final BuildContext context, final LoginQrcode qrcode) async {
```
Leptopoda, line 96: If we manage to make the redirection work in the route this should be moved to into the remaining `LoginQrcodePage`

packages/neon/neon/lib/src/router.dart
```diff
@@ -29,6 +31,15 @@ class AppRouter extends GoRouter {
...
             final account = accountsBloc.activeAccount.valueOrNull;
 
+            final loginQrcode = parseLoginQrcodeURL(state.location);
+            if (loginQrcode != null) {
+              return LoginQrcodeIntermediateRoute(
+                serverURL: loginQrcode.server,
+                loginName: loginQrcode.user,
+                password: loginQrcode.password,
+              ).location;
+            }
```
Leptopoda, line 41: I assume this is where the intet get's catched?

packages/neon/neon/lib/src/router.dart
```diff
@@ -105,6 +116,14 @@ class HomeRoute extends GoRouteData {
       path: 'flow/:serverURL',
       name: 'loginFlow',
     ),
+    TypedGoRoute<LoginQrcodeRoute>(
+      path: 'qrcode',
+      name: 'loginQrcode',
+    ),
+    TypedGoRoute<LoginQrcodeIntermediateRoute>(
+      path: 'qrcode/intermediate/:serverURL/:loginName/:password',
```
Leptopoda, line 124: wouldn't it make sense to just call this the same as the qr code encodes?

THis way we wouldn't need the global redirect (wich AFAICT gets executed on every navigation)

packages/neon/neon/lib/src/pages/login_qrcode.dart
Leptopoda: I'm sorry. Looks like you already tried this.
But I think all the logic should be in the routing and not this intermediate page.

Maybe the `LoginQrcodeIntermediateRoute` can actually build the `LoginCheckServerStatusPage`. This page could be adapted to also take the `LoginQrcode` directly (separate named cunstructor?) and if the full qrcode is present (rather than only the server) it will redirect differently when clicking next

packages/neon/neon/lib/src/pages/login_qrcode.dart
```diff
@@ -0,0 +1,105 @@
...
+  State<LoginQrcodePage> createState() => _LoginQrcodePageState();
+}
+
+class _LoginQrcodePageState extends State<LoginQrcodePage> {
+  String? _lastErrorURL;
+
+  @override
+  Widget build(final BuildContext context) => Scaffold(
+        appBar: AppBar(),
+        body: ReaderWidget(
```
Leptopoda, line 25: We should probably also specify `codeFormat: Format.qrCode` to avoid scanning wrong formats in the first place

packages/neon/neon/lib/src/pages/login_qrcode.dart
provokateurin: > Maybe the LoginQrcodeIntermediateRoute can actually build the LoginCheckServerStatusPage

I don't know if I like this. I already tried something similar but it doesn't really improve the overall structure or logic.

I was thinking to unify the LoginCheckServerStatusPage and LoginCheckAccountPage and just let it check the relevant info based on the parameters. Then we can just process the whole QR code on one page without having any additional redirects etc. We could still keep them as separate routes or unify them too, not sure if that improves anything. What do you think?

packages/neon/neon/lib/src/pages/login_qrcode.dart
provokateurin: This is how it's done all over the app. I'll make a follow up to fix it everywhere

packages/neon/neon/lib/src/router.dart
provokateurin: Exactly. Sadly it doesn't contain the full URI so we have to do this matching. The whole `nc://login` part is removed

packages/neon/neon/lib/src/router.dart
provokateurin: As mentioned above that doesn't work because the URI is stripped. It would just be one big parameter which really sucks. Not a problem of our architecture but of the structure of the QR code login URIs.

packages/neon/neon/lib/src/pages/login_qrcode.dart
Leptopoda: I don't have anything against multiple routes for the same page.
This might even be preferable as the default BackButton would redirect to the correct page.

I don't know how much unifying them makes sense.
I mean there is a reason they are separated in our flow diagram (https://github.com/provokateurin/nextcloud-neon/blob/main/docs/login.md)
I'm also not 100% conviced by the current approach of pushing the LoginCheckServerStatusRoute as that isn't the declarative approach go_router enables us to use (It's generally quite easy to forget a).

Maybe something like this makes sense?
```dart
@TypedGoRoute<LoginRoute>(
  path: '/login',
  name: 'login',
  routes: [
    TypedGoRoute<LoginQrcodeRoute>(
      path: 'qrcode',
      name: 'loginQrcode',
      routes: [
        TypedGoRoute<LoginQrcodeCheckServerStatusRoute>( // LoginCheckServerStatusPage
          path: 'check/server',
        ),
      ],
    ),
    TypedGoRoute<LoginFlowRoute>(
      path: 'flow/:serverURL',
      name: 'loginFlow',
      routes: [
        TypedGoRoute<LoginCheckServerStatusRoute>( // LoginCheckServerStatusPage
          path: 'check/server',
        ),
      ],
    ),
    TypedGoRoute<LoginCheckAccountRoute>(
      path: 'check/account',
      name: 'checkAccount',
    ),
  ],
)
```

packages/neon/neon/lib/src/router.dart
Leptopoda: can we think about #144 nevertheless?
I don't like the `/` between the parameters.

We can also do all at once with the mentioned issue.

Leptopoda:
I think we should reconsider the LoginPage design.
It looks a bit bad having content separated by text.

---

Maybe we can move the `Login using the server creds` into the textbox hint and the `https://...` into it's decoration.
And the Qr description is already handled by the tooltip of the icon button.

Finally maybe move the qr button below the text field as I personally prefer that design.

---

For comparison:
Screenshot_20230709-211358_Nextcloud

provokateurin:
I agree, that design looks better :)

Leptopoda:
How do you like this?
It cleans up the routes and navigation but adds some boilerplate to the routes.
It's all declarative though :)

Leptopoda:
This looks odd. An error should be shown in the sccount tile (not 100% related to this PR but the qr flow is the only one this can happen)

---



An invalid url will let the progress indicators spin forever.
Also some design cleanups needed (not related to this pr though)

packages/neon/neon/lib/src/pages/login_check_server_status.dart
```diff
@@ -47,35 +58,49 @@ class _LoginCheckServerStatusPageState extends State<LoginCheckServerStatusPage>
...
+  void _onContinue() {
+    if (widget.loginName != null) {
+      LoginCheckAccountRoute(
+        serverUrl: widget.serverURL,
+        loginName: widget.loginName!,
+        password: widget.password!,
+      ).pushReplacement(context);
+    } else {
+      LoginFlowRoute(serverUrl: widget.serverURL).pushReplacement(context);
+    }
```
provokateurin, line 101: This makes so much sense, somehow I never thought about it

packages/neon/neon/lib/src/router.dart
```diff
@@ -29,19 +33,17 @@ class AppRouter extends GoRouter {
           navigatorKey: navigatorKey,
           initialLocation: const HomeRoute().location,
           redirect: (final context, final state) {
-            final account = accountsBloc.activeAccount.valueOrNull;
-
             final loginQrcode = LoginQrcode.tryParse(state.location);
-            if (loginQrcode != null) {
-              return LoginQrcodeIntermediateRoute(
-                serverURL: loginQrcode.server,
+            if (state.path == const LoginRoute().location && loginQrcode != null) {
```
provokateurin, line 37: Why is this limited to when the user is already on the login page? They could have one account logged in and be on the home page and then just scan a login qr code.

provokateurin (commented):
The changes look good. Haven't tested it yet. Instead of the `.qrCode` constructors I'd like to have something more generic since it can technically also used without a QR code. I'd suggest `.withCredentials` or something similar.

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -2,20 +2,34 @@ import 'dart:convert';
...
 import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
 import 'package:nextcloud/nextcloud.dart';
 
 part 'account.g.dart';
 
+/// Credentials interface
+@internal
+@immutable
+abstract interface class Credentials {
```
Leptopoda, line 14: Ideally we'd use this interface everywhere we have a `withCredentials` constructor but sadly go_router wants the base types themselves for the routes. Meaning we'd do something like Credentials -> basetypes -> Credentials wich is way more boilerplate than it already needs to be.

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -2,20 +2,34 @@ import 'dart:convert';
...
+  /// Url of the server
+  abstract final String serverURL;
+
+  /// Username
+  abstract final String username;
+
+  /// Password
+  abstract final String? password;
```
Leptopoda, line 22: Does this need to be nullable?

I don't see any reason for it (no code seems to break with it being changed but I might be wrong)

Leptopoda (approved):

Leptopoda:
This shouldn't be an approval just a comment :)

packages/neon/neon/lib/src/router.dart
```diff
@@ -102,81 +129,222 @@ class HomeRoute extends GoRouteData {
...
+    required super.loginName,
+    required super.password,
+  });
+
+  @override
+  String get serverUrl => super.serverUrl;
+  @override
+  String get loginName => super.loginName;
+  @override
+  String get password => super.password;
```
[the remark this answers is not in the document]
provokateurin: I guess they just don't check if the names are the same.

packages/neon/neon/lib/src/models/account.dart
provokateurin: At some point we will have guest accounts without passwords :)

packages/neon/neon/lib/src/models/account.dart
provokateurin: Is this anything that can be fixed in this PR?

Leptopoda: It can but I don't like all the boilerplate.
Especially with the redirects it get's worse:

Credential -> string -> credential  -> string -> Credential -> basetypes.

With the three steps being Login*Route, AddAccount*Route and *Page.

packages/neon/neon/lib/models.dart
```diff
@@ -1,3 +1,3 @@
-export 'package:neon/src/models/account.dart';
+export 'package:neon/src/models/account.dart' hide Credentials, LoginQrcode;
```
provokateurin, line 1: Just mark them as internal?

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -76,3 +93,106 @@ extension AccountFind on Iterable<Account> {
...
+      );
+    }
+
+    throw const FormatException();
+  }
+
+  /// Creates a new `LoginQrcode` object by parsing a url string.
+  ///
+  /// Returns `null` if the [url] string is not valid as a LoginQrcode or a
+  /// valid path.
```
provokateurin, line 167:
```suggestion
  /// Returns `null` if the [url] string is not valid as a LoginQrcode.
```

packages/neon/neon/lib/src/pages/login_qrcode.dart
```diff
@@ -0,0 +1,63 @@
...
+          },
+        ),
+      );
+
+  void _processLoginQrcode(final BuildContext context, final LoginQrcode qrcode) {
+    LoginCheckServerStatusRoute.withCredentials(
+      serverUrl: qrcode.serverURL,
+      loginName: qrcode.username,
+      password: qrcode.password,
+    ).pushReplacement(context);
```
provokateurin, line 61: Can be inlined as it is only used once

packages/neon/neon/lib/models.dart
Leptopoda: They already are :)

But AFAIK the annotation is not hiding it and just produces a lint so we notice if it's accidentally exported.
https://dart.dev/tools/diagnostic-messages#invalid_export_of_internal_element

provokateurin: I see, but is it really a problem if someone absolutely wants to import it? I mean there is probably no good reason, but still

Leptopoda: We'd need to provide api stability for it :thinking: 

But we can export it if you want.

provokateurin: Let's just keep it and if any one ever needs it they can export it

provokateurin:
@Leptopoda one more formal approval?

packages/neon/neon/lib/src/pages/login.dart
```diff
@@ -89,11 +104,10 @@ class _LoginPageState extends State<LoginPage> {
                         hintText: 'https://...',
                       ),
                       keyboardType: TextInputType.url,
-                      initialValue: widget.serverURL,
                       validator: (final input) => validateHttpUrl(context, input),
-                      onFieldSubmitted: (final input) async {
+                      onFieldSubmitted: (final input) {
                         if (_formKey.currentState!.validate()) {
-                          await _beginLoginFlow(input);
+                          _beginLoginFlow(input);
```
Leptopoda, line 110: We should probably also inline this

Leptopoda (approved):
I'll push this small thing and it's good to merge

Leptopoda:
@provokateurin ready :tada:

provokateurin:
Of course random CI failure :upside_down_face:

PR #295, opened 2023-05-06
neon: use Size class for sizes

Leptopoda:
I think this simplifies the interfaces of some widgets. What do you think?

I quite like the ability to define squares and having things like `size.shortestSide`


packages/neon/neon_files/lib/widgets/file_preview.dart
```diff
@@ -17,18 +16,19 @@ class FilePreview extends StatelessWidget {
...
   final FilesBloc bloc;
   final FileDetails details;
-  final int width;
-  final int height;
+  final Size size;
   final Color? color;
   final BorderRadius? borderRadius;
   final bool withBackground;
 
+  int get width => size.width.toInt();
```
Leptopoda, line 24: we could make this private :thinking:

packages/neon/neon_news/lib/widgets/feed_icon.dart
```diff
@@ -1,33 +1,32 @@
 part of '../neon_news.dart';
 
 class NewsFeedIcon extends StatelessWidget {
-  const NewsFeedIcon({
+  NewsFeedIcon({
     required this.feed,
-    this.size = 48,
+    this.dimension = 48,
```
provokateurin, line 6: I would keep this named size. The flutter Icon class also calls it size

provokateurin (changes requested):

packages/neon/neon_news/lib/widgets/feed_icon.dart
Leptopoda: Bit it's called dimension when talking about a square.
https://api.flutter.dev/flutter/dart-ui/Size/Size.square.html

provokateurin (approved):

Issue #300, opened 2023-05-07
neon: better acessibility 


Leptopoda:
1. for screenreaders:
- We currently have a lot of IconButtons without any tooltip.
- some lists may contain invisible a11y nodes (investigation needed)

2. rtl layout
while a lot of things are multi directional by default our drawer/quickbar implementation isn't.

---

Fortunately I'm not in need for a screen reader and I only know ltr languages/layouts so I'm not the best person for this but it's a starting point.

provokateurin:
I would love to work on this, but I have very little experience with these things.

provokateurin:
I found https://docs.flutter.dev/accessibility-and-localization/accessibility which seems to be a nice introduction to the topic regarding flutter.

Leptopoda:
I've already started to rewrite the drawer for #294.

> I found https://docs.flutter.dev/accessibility-and-localization/accessibility which seems to be a nice introduction to the topic regarding flutter.

I've read this a dozen times over the last few years but would still only consider to know the tip of the iceberg.
The talk under https://docs.flutter.dev/accessibility-and-localization/accessibility#building-with-accessibility-in-mind is interesting.

> I hope that it will be possible to 'lint' accessibility issues.

You can test it in unit test :woman_shrugging:

Leptopoda:
I think a great starting point would be to add tooltips to every icon button.
we don't need to wrap it in one like:
https://github.com/provokateurin/nextcloud-neon/blob/a7be31ff2632d02eea7c19afbfb8e6edf2ea473d/packages/neon/neon/lib/src/pages/home.dart#L361

as IconButton already has support for it:
https://api.flutter.dev/flutter/material/IconButton/IconButton.html

provokateurin:
I think the next steps could be adding widget tests for everything (which we should do anyway) and then adding the automatic accessibility checks.

Leptopoda:
Our `ProgressIndicators` should also include a semantic label.
Same for our `NeonValidationTile`

Leptopoda:
Should we close this one?
#2113 Should help a lot and I don't think we can do anything more rn.

provokateurin:
There is probably a lot we could improve, but having this very broad issue doesn't help much with that.

PR #302, opened 2023-05-07
Add tooltips to all IconButtons

provokateurin:
https://github.com/provokateurin/nextcloud-neon/issues/300

### DIFF
--- a/packages/neon/neon/lib/src/pages/account_settings.dart
+++ b/packages/neon/neon/lib/src/pages/account_settings.dart
@@ -33,6 +33,7 @@ class AccountSettingsPage extends StatelessWidget {
                   Navigator.of(context).pop();
                 }
               },
+              tooltip: AppLocalizations.of(context).accountOptionsRemove,
               icon: const Icon(MdiIcons.delete),
             ),
             IconButton(
@@ -46,6 +47,7 @@ class AccountSettingsPage extends StatelessWidget {
                   await _options.reset();
                 }
               },
+              tooltip: AppLocalizations.of(context).settingsResetFor(_name),
               icon: const Icon(MdiIcons.cogRefresh),
             ),
           ],
--- a/packages/neon/neon/lib/src/pages/home.dart
+++ b/packages/neon/neon/lib/src/pages/home.dart
@@ -245,16 +245,14 @@ class _HomePageState extends State<HomePage> {
                                                             margin: const EdgeInsets.symmetric(
                                                               vertical: 5,
                                                             ),
-                                                            child: Tooltip(
-                                                              message: account.client.humanReadableID,
-                                                              child: IconButton(
-                                                                onPressed: () {
-                                                                  _accountsBloc.setActiveAccount(account);
-                                                                },
-                                                                icon: IntrinsicHeight(
-                                                                  child: NeonAccountAvatar(
-                                                                    account: account,
-                                                                  ),
+                                                            child: IconButton(
+                                                              onPressed: () {
+                                                                _accountsBloc.setActiveAccount(account);
+                                                              },
+                                                              tooltip: account.client.humanReadableID,
+                                                              icon: IntrinsicHeight(
+                                                                child: NeonAccountAvatar(
+                                                                  account: account,
                                                                 ),
                                                               ),
                                                             ),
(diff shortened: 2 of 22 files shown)


packages/neon/neon/lib/src/pages/account_settings.dart
```diff
@@ -46,6 +47,7 @@ class AccountSettingsPage extends StatelessWidget {
                   await _options.reset();
                 }
               },
+              tooltip: AppLocalizations.of(context).settingsResetFor(_name),
```
Leptopoda, line 50: This string ends with a question mark. I wouldn't expect that in this case.

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -556,6 +550,12 @@ class _HomePageState extends State<HomePage> {
...
                                           },
+                                          tooltip: AppLocalizations.of(context).settingsAccount,
+                                          icon: IntrinsicWidth(
+                                            child: NeonAccountAvatar(
+                                              account: account,
+                                            ),
+                                          ),
                                         ),
                                       ],
                                     ),
```
Leptopoda, line 561: Really can't wait to rebase my restructure :/

packages/neon/neon_news/lib/pages/article.dart
```diff
@@ -137,12 +143,14 @@ class _NewsArticlePageState extends State<NewsArticlePage> {
                       mode: LaunchMode.externalApplication,
                     );
                   },
+                  tooltip: AppLocalizations.of(context).newsArticleOpenLink,
```
Leptopoda, line 146: maybe `Open in browser` is better

packages/neon/neon_news/lib/widgets/feeds_view.dart
```diff
@@ -90,6 +89,8 @@ class NewsFeedsView extends StatelessWidget {
                     ),
                   );
                 },
+                tooltip: AppLocalizations.of(context).newsShowFeedErrorMessage,
```
Leptopoda, line 92: Huh what's this.

An IconButton with a text child?
A text will automatically be added to the a11y tree so we might have both the tooltip and the text.

Maybe this shouldn't be an IconButton altogether :thinking:

packages/neon/neon_notes/lib/pages/note.dart
```diff
@@ -104,6 +101,12 @@ class _NotesNotePageState extends State<NotesNotePage> {
                     _titleFocusNode.unfocus();
                   }
                 },
+                tooltip: _showEditor
+                    ? AppLocalizations.of(context).notesHideEditor
```
Leptopoda, line 105: Maybe the strings could be something like `edit note` and `preview changes`

Leptopoda (changes requested):

Leptopoda:
Just stumbled across `NeonException` and noticed that there might be some IconButton behavior that's harder to find

https://github.com/provokateurin/nextcloud-neon/blob/ae383f95b34dd38da1edcc897b403df964dce61d/packages/neon/neon/lib/src/widgets/exception.dart#L51-L60

Leptopoda:
FloatingActionButton is another candidate.

similar to my review above the text shouldn't just say 'add' as this is not very helpful to people not seeing the screen. It should probably describe the actions like 'add news feed' or 'create note'

packages/neon/neon_news/lib/widgets/feeds_view.dart
provokateurin: I'd leave it as is right now since I'm only trying to add the tooltips for all iconbuttons and not fix everything :)

Leptopoda:
> Just stumbled across NeonException and noticed that there might be some IconButton behavior that's harder to find

inkwell still needs attention; I think the only ones are:
https://github.com/provokateurin/nextcloud-neon/blob/d3519b3bbf271461b3a2fea41a6984bcb440e13a/packages/neon/neon_files/lib/widgets/browser_view.dart#L173-L179

https://github.com/provokateurin/nextcloud-neon/blob/d3519b3bbf271461b3a2fea41a6984bcb440e13a/packages/neon/neon/lib/src/widgets/exception.dart#L51-L60

and when option.isEmpty

https://github.com/provokateurin/nextcloud-neon/blob/d3519b3bbf271461b3a2fea41a6984bcb440e13a/packages/neon/neon_notes/lib/widgets/category_select.dart#L74-L89

The first two could be made IconButtons althought that would also change things like padding (so caution).
For the last one I'd add the `onTap` callBack to the listTile itself and conditionally wrap it in a `ToolTip` (might make sens to externalize the option.isEmpty into a variable). Also while being there the builder can be removed as we use Listview.builder wich already wraps the children in their own builder (I can see why this should be for a separate PR though).

Leptopoda (approved):
LGTM
I'm sure that people using screen readers on a daily basis will still find a few edge cases but it's better than before :+1: 

I quite liked the old design of the path_bar in neon_files
Screenshot_20230518_095101

The TextButton will be colored while the home icon and the separators aren't.
Also the Home icon looks a bit to big.

But that might be personal preference.

provokateurin:
I can also make it look more like it was before. I don't have anything against this new design either 🤷‍♀️

Leptopoda:
Let's keep it as is.
I probably just need getting used to it.

Using a TextButton (with a highlight and color) might show the user that it is trappable and not just a nice UI element.

PR #309, opened 2023-05-11
Stop pinning intl

provokateurin:
This broke CI when the new flutter version was released (due to missing version pinning for flutter in CI) and it also makes upgrading annoying, so we just drop it.


Leptopoda (approved):
While I'm not a fan of not pinning the version at all I see why it's needed here.

Especially with the intl 0.18 & flutter 3.7 problems that existed

Leptopoda:
Now we are hitting CI issues with not having updated to 3.10 yet :/

provokateurin:
I will have to put all three PRs together as this is a total mess right now...

PR #308, opened 2023-05-10
Update dependencies

provokateurin:
Depends on https://github.com/creativecreatorormaybenot/wakelock/pull/195

PR #310, opened 2023-05-11
Fix and update dependencies and flutter

provokateurin:
Replaces https://github.com/provokateurin/nextcloud-neon/pull/309 and https://github.com/provokateurin/nextcloud-neon/pull/308 since it was a total mess with the new flutter version...
Depends on https://github.com/provokateurin/nextcloud-neon/pull/311

### DIFF
--- a/packages/nextcloud/pubspec.yaml
+++ b/packages/nextcloud/pubspec.yaml
@@ -5,7 +5,7 @@ environment:
   sdk: '>=2.19.0 <3.0.0'
 
 dependencies:
-  cookie_jar: ^3.0.1
+  cookie_jar: ^4.0.2
   crypto: ^3.0.2
   crypton: ^2.0.5
   intl: ^0.17.0
@@ -24,6 +24,6 @@ dev_dependencies:
     git:
       url: https://github.com/stack11/dart_nit_picking
       ref: 0b2ee0d
-  process_run: ^0.12.5+2
+  process_run: ^0.13.0
   test: ^1.16.0
   xml_serializable: ^2.2.2
--- a/packages/nextcloud/pubspec.yaml
+++ b/packages/nextcloud/pubspec.yaml
@@ -2,13 +2,13 @@ name: nextcloud
 version: 1.0.0
 
 environment:
-  sdk: '>=2.19.0 <3.0.0'
+  sdk: '>=3.0.0 <4.0.0'
 
 dependencies:
   cookie_jar: ^4.0.2
   crypto: ^3.0.2
   crypton: ^2.0.5
-  intl: ^0.17.0
+  intl: ^0.18.0
   json_annotation: ^4.7.0
   version: ^3.0.2
   xml: ^6.1.0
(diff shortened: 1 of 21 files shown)


packages/nextcloud/pubspec.yaml
```diff
@@ -2,13 +2,13 @@ name: nextcloud
...
-  sdk: '>=2.19.0 <3.0.0'
+  sdk: '>=3.0.0 <4.0.0'
 
 dependencies:
-  cookie_jar: ^3.0.1
+  cookie_jar: ^4.0.2
   crypto: ^3.0.2
   crypton: ^2.0.5
-  intl: ^0.17.0
+  intl: any
```
Leptopoda, line 11: Is there a reason we still need this now?
I think pinning it to 0.18.0 might be good.

provokateurin: If another flutter app would try to import the package but still uses flutter <3.10 (requiring intl 0.17) then it would fail. I think the best is to unpin it, as much as I hate doing it.

provokateurin: That's the magic of unreleased software hehe

packages/neon/neon/lib/src/app.dart
```diff
@@ -266,7 +266,7 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, tray.Tra
   }
 
   @override
-  Widget build(final BuildContext context) => StreamBuilder<Brightness>(
+  Widget build(final BuildContext context) => StreamBuilder<Brightness?>(
```
Leptopoda, line 269: This breaks the app that the NativeSplash never gets removed.
the blow `!platformBrightnessSnapshot.hasData` internally checks wether `platformBrightnessSnapshot.data != null` thus we have a problem here

Leptopoda (changes requested):

packages/neon/neon/lib/src/app.dart
Leptopoda: I already stumbled over this another time and IMO we should just generate a `darkTheme` and a `lightTheme` and let MaterialApp decide the brightness for us.

---

A bit unrelated I once played with making `nextcloudTheme` a ThemeExtension as this would enable us to use the usual Theme.of(context).extesion<CustomExtenstion>(); to get our theme everywhere we have our beloved BuidContext. Sadly this just added more boilerplate than was worth for me to really submit.

provokateurin: Sounds good. I forgot to check this change, I knew that it could break things.

Leptopoda: I could make a change that we land before this PR as the theme changes isn't related to the update.

provokateurin: Sure, I'd appreciate that. We would have to merge it with failing CI though :)

provokateurin:
Since our CI is currently broken I'd be ok with merging it while still having the git dependency. @Leptopoda ok with you too?

Leptopoda (approved):

PR #311, opened 2023-05-11
neon: cleanup theming

Leptopoda:
let the MaterialApp decide wich theme to use for a given brightness

---

I think it's a nice little cleanup :)
you can also just cherrypick them onto #310 if you don't want to merge with failing CI


packages/neon/neon/lib/src/app.dart
```diff
@@ -260,42 +242,53 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, tray.Tra
...
+              builder: (final context, final activeAccountSnapshot) {
+                if (themeMode == null || themeOLEDAsDark == null) {
+                  return Container();
+                }
+
+                FlutterNativeSplash.remove();
+                return ValueListenableBuilder(
+                  valueListenable: themeNotifier,
+                  builder: (final context, final nextcloudTheme, final child) {
+                    final routerDelegate = AppRouter(
```
provokateurin, line 267: Do we actually want to rebuild the navigator?

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -48,7 +46,7 @@ class _HomePageState extends State<HomePage> {
 
     _capabilitiesBloc.capabilities.listen((final result) async {
       if (result.data != null) {
-        widget.onThemeChanged(result.data!.capabilities.theming);
+        themeNotifier.value = result.data!.capabilities.theming;
```
provokateurin, line 49: I'd say this should go into the bloc. No need to only run in the widget.

provokateurin (changes requested):

packages/neon/neon/lib/src/pages/home.dart
Leptopoda: good idea :)

provokateurin: Could you fix this soon? Then we can merge this and afterwards fix the dependencies and flutter.

Leptopoda: Oh yeah one sec

provokateurin:
Yep, that's the right place.

Leptopoda:
Never mind again. It is working I was just logged in with another account that doesn't have theming :)

This should be good to go.

provokateurin:
Ok, let me check it locally and play CI :smile:

Leptopoda:
Ok I had fixed the null value for the theme with my last force push; But it's not in the commit  :thinking:

provokateurin:
Huh weird, but I already fixed it :)

Issue #317, opened 2023-05-13
neon:  Theme switching when changing accounts


Leptopoda:
Moving the themeNotifier into the bloc made the theme not switch on account switching.
Looks like the capabilities are only fetched once per app start per account.

Having multiple accounts will enable the correct theme on the first account checkout but not on subsequent.

---

I won't have much time and am not that familiar with our blocs yet so feel free to tackle this.

provokateurin:
There are still problems. When I open the settings weird things start to happen and I randomly get the wrong theme etc.

PR #318, opened 2023-05-13
neon: Fix theme switching when switching accounts

provokateurin:
Fixes https://github.com/provokateurin/nextcloud-neon/issues/317

Also noticed the capabilities were dispose in the home page which was wrong.

### DIFF
--- a/packages/neon/neon/lib/src/utils/bloc.dart
+++ b/packages/neon/neon/lib/src/utils/bloc.dart
@@ -40,3 +40,17 @@ abstract class InteractiveBloc extends Bloc {
     }
   }
 }
+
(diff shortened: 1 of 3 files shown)


packages/neon/neon/lib/src/utils/bloc.dart
```diff
@@ -40,3 +40,17 @@ abstract class InteractiveBloc extends Bloc {
     }
   }
 }
+
+extension DisposeableIterableBloc on Iterable<Bloc> {
+  void dispose() {
```
Leptopoda, line 45: Do we want to call this `disposeAll()` :thinking:

Leptopoda (approved):
But looks good :)

Issue #321, opened 2023-05-13
Broken built_value changes


provokateurin:
The following parameter
```json
{
    "name": "subadmin",
    "in": "query",
    "description": "Groups where the user is subadmin",
    "schema": {
        "type": "array",
        "items": {
            "type": "string"
        },
        "default": []
    }
},
```
generates this code
```dart
final BuiltList<String> subadmin = const [],
```
which is not valid.

Leptopoda:
Ok just had a look and the problem is that we need the default value (as a const) in the parameter list of an operation.

I think it is valid here to just make the parmeter a plain `List` so the caller would need to pas `builtListValue.toList()`.

Leptopoda:
https://github.com/provokateurin/nextcloud-neon/blob/6531b2ada799a158c58afeeb11dea168512bdb7c/packages/dynamite/dynamite/lib/src/type_result/object.dart#L21

Hard coded variable name :grimacing:

I'll make a few cleanups (I might get to it today but can't promise anything before Tuesday)

provokateurin:
I pushed the openapi-automatic branch with the broken code. You can use it check your fix.

provokateurin:
I'm pretty sure https://github.com/provokateurin/nextcloud-neon/blob/6531b2ada799a158c58afeeb11dea168512bdb7c/packages/dynamite/dynamite/lib/src/type_result/map.dart#L13 should be `MapBuilder`.

provokateurin:
I get `Bad state: No serializer for '_$ContentString<BuiltList<int>>'.` for
```json
                    {
                        "name": "shareTypes",
                        "in": "query",
                        "description": "Types of shares to search for",
                        "schema": {
                            "type": "string",
                            "contentMediaType": "application/json",
                            "contentSchema": {
                                "type": "array",
                                "default": null,
                                "items": {
                                    "type": "integer",
                                    "format": "int64"
                                },
                                "description": "Types of shares to search for"
                            }
                        }
                    },
```
although the serilizers contain
```dart
      ..addBuilderFactory(
        const FullType(ContentString, [
          FullType(BuiltList, [FullType(int)])
        ]),
        ContentString<BuiltList<int>>.new,
      )
```

Leptopoda:
> I get Bad state: No serializer for '_$ContentString<BuiltList<int>>'. for

haven't had the time to look into it  but did this happen using one of `serializeNextcloud/deserializeNextcloud`?

[...]

PR #324, opened 2023-05-17
built_value followup

Leptopoda:
fixes: #321 

For now just tested on master and openapi-automatic.
Still needs testing on the spreed branch and some cleanups.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -726,7 +726,7 @@ class OpenAPIBuilder implements Builder {
                                 ),
                                 parameter.schema!,
                                 nullable: dartParameterNullable,
-                              );
+                              ).dartType;
 
                               state.resolvedTypeCombinations.add(result);
 
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -930,6 +930,8 @@ class OpenAPIBuilder implements Builder {
                                       ),
                                       mediaType.schema!,
                                     );
+                                    state.resolvedTypeCombinations.add(result);
+
                                     if (mimeType == '*/*' || mimeType.startsWith('image/')) {
                                       dataType = 'Uint8List';
                                       dataValue = 'response.body';
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -779,7 +779,7 @@ class OpenAPIBuilder implements Builder {
                                 code.write('if (${_toDartName(parameter.name)} != null) {');
                               }
                               final value = result.encode(
-                                result.serialize(_toDartName(parameter.name)),
+                                _toDartName(parameter.name),
                                 onlyChildren: result is TypeResultList && parameter.in_ == 'query',
                                 // Objects inside the query always have to be interpreted in some way
                                 mimeType: 'application/json',
@@ -858,7 +858,7 @@ class OpenAPIBuilder implements Builder {
                                       code.write('if ($parameterName != null) {');
                                     }
                                     code.write(
-                                      'body = Uint8List.fromList(utf8.encode(${result.encode(result.serialize(parameterName), mimeType: mimeType)}));',
+                                      'body = Uint8List.fromList(utf8.encode(${result.encode(parameterName, mimeType: mimeType)}));',
                                     );
                                     if (dartParameterNullable) {
                                       code.write('}');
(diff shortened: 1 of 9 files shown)


packages/dynamite/dynamite/lib/src/type_result/type_result.dart
```diff
@@ -46,17 +46,42 @@ abstract class TypeResult {
...
+  /// Serializes the variable named [object].
+  ///
+  /// The serialized result is an [Object]?
+  String serialize(final String object) => 'jsonSerializers.serialize($object, specifiedType: const $fullType)';
 
-  String deserialize(final String object, {final bool toBuilder = false});
+  /// Deserializes the variable named [object].
+  ///
+  /// The serialized result will be of [name].
+  String deserialize(final String object) =>
```
Leptopoda, line 57: Ideally I'd like to call decode automatically just like encode does with serialize.

packages/nextcloud/lib/src/nextcloud.openapi.dart
```diff
@@ -301,16 +316,17 @@ class NextcloudCoreClient {
       queryParameters['file'] = file;
     }
     if (x != 32) {
-      queryParameters['x'] = x.toString();
+      queryParameters['x'] = json.encode(jsonSerializers.serialize(x, specifiedType: const FullType(int)));
```
provokateurin, line 319: Not sure if this really makes sense... Parameter encoding isn't correct right now: https://swagger.io/docs/specification/describing-parameters/#schema-vs-content. We neither support style and explode nor content.

packages/dynamite/dynamite/lib/src/type_result/base.dart
```diff
@@ -9,24 +9,17 @@ class TypeResultBase extends TypeResult {
...
-  String deserialize(final String object, {final bool toBuilder = false}) {
-    if (name == 'JsonObject') {
-      return 'JsonObject($object)${nullable ? '?' : ''}';
+  }) {
+    if (name == 'String') {
+      return object;
     }
 
-    return '($object as $name ${nullable ? '?' : ''})';
+    return super.encode(object, mimeType: mimeType);
```
provokateurin, line 22: See my comment below.

provokateurin:
I'm not convinced of all the changes. For base types we don't need to run the json serializers and we need to be more careful with serializing parameters as that is just hardcoded right now.

packages/nextcloud/lib/src/nextcloud.openapi.dart
Leptopoda: I don't get your point. Could you explain more?
I don't see any behavior change to what we already did.

```dart
final int x = 3;
var serialized = jsonSerializers.serialize(x, specifiedType: const FullType(int)); //this is still and int with value 3
var encoded = json.encode(serialized); // this is a String with the value '3'

print(x == serialized); // true
print(encoded.runtimeType); // String
print(encoded == x.toString()); // true
```

Yes the generated code looks more complex but we can unify both serialization and encoding for all TypeResults (making it less error prone).

---

I agree that we could leave encode and serialize separated (thus that encode won't serialize for you).

provokateurin: Sure it is the same, but it adds unnecessary work. I think the previous serialization/encoding for base types was just fine and did the least amount of work.

Leptopoda: I prefer sharing the serialization much as possible :/

Is it ok for you if we share the serializer approach for all objects (Enum, Map, List and Object) but keep the old implementation for Base types?

provokateurin: Yes! That's what I wanted. Only the base types should be different, everything else should be the same.

packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -1786,7 +1796,11 @@ TypeResult resolveType(
                             ],
                           ],
                           'try {',
-                          'result.${fields[result.name]!} = ${result.deserialize('data', toBuilder: true)};',
+                          if (result is TypeResultBase || result is TypeResultEnum) ...[
+                            'result._${fields[result.name]!} = ${result.deserialize('data')};',
+                          ] else ...[
+                            'result._${fields[result.name]!} = ${result.deserialize('data')}.toBuilder();',
+                          ],
```
provokateurin, line 1803: Is the new `_` intended? I think this breaks the code...

provokateurin: But CI passes which confuses me

Leptopoda: I can check again but it's intended.
When we use it without the _ the setter will be called initializing the private variable with an empty builder.
This is so one could do `result.field.replace(value)` instead of `result._field = value.toBuilder()`.
The problem is that building the result will also build the instantiated builder but as it might be empty (because the catch case was triggered) it might fail with required fields not being set.

Does this make sense?

---

I'll check again tomorrow as I might have overlooked something.

provokateurin: I think I understand. I just didn't know that built_value has these private setters. Thanks for the explanation!

provokateurin (approved):

Issue #322, opened 2023-05-16
Expose the currently supported Nextcloud version


adil192:
Following #226 the `nextcloud` package only supports Nextcloud 26.
I'm adding a check to my app to tell the user about this, but it would be better if the current supported Nextcloud version was stored in your repo rather than mine.

## Static field for current supported Nextcloud version
It would also be nice if you stored the current supported Nextcloud version, so I could do something like this:
```dart
print('Only Nextcloud ${NextcloudCoreServerStatus.supportedVersion} is supported');
```
rather than hardcoding in 'Only Nextcloud 26'.

## My current code
```dart
final NextcloudCoreServerStatus status;
try {
  status = await client.core.getStatus();
} catch (e) {
  throw NcLoginFailure();
}
if (!status.version.startsWith('26.')) { // <- here
  throw NcUnsupportedFailure('Only Nextcloud 26 is supported!');
}
```

## Proposed code 1
```dart
final NextcloudCoreServerStatus status;
try {
  status = await client.core.getStatus();
} catch (e) {
  throw NcLoginFailure();
}
// here [versionSupportedByPackage] is a boolean for whether [status.version] is 26.*
if (!status.versionSupportedByPackage) {
  throw NcUnsupportedFailure('Only Nextcloud 26 is supported!');
}
```

## Proposed code 2 (preferred)
```dart
final NextcloudCoreServerStatus status;
try {
  status = await client.core.getStatus();
} catch (e) {
  throw NcLoginFailure();
}
// here [versionSupportedByPackage] is a string (which would currently be '26')
if (!status.version.startsWith(NextcloudCoreServerStatus.versionSupportedByPackage)) {
  throw NcUnsupportedFailure('Only Nextcloud ${NextcloudCoreServerStatus.versionSupportedByPackage} is supported!');
}
```

provokateurin:
There already is a check for the supported versions (you need to check core for the server version): https://github.com/provokateurin/nextcloud-neon/blob/main/packages/nextcloud/lib/src/version_supported.dart
We could extend it to also expose the supported version so that it can be told to the user. Does that work for you?

adil192:
Yes that's perfect, thank you

PR #325, opened 2023-05-17
Expose supported versions

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/322

Already making use of new dart features :)

@adil192 does this interface work for you?

### DIFF
--- a/packages/neon/neon/lib/src/pages/home.dart
+++ b/packages/neon/neon/lib/src/pages/home.dart
@@ -60,30 +60,22 @@ class _HomePageState extends State<HomePage> {
             ...appsResult.data!.map((final a) => a.id),
           ]) {
             try {
-              bool? supported;
-              switch (id) {
-                case 'core':
-                  supported = await widget.account.client.core.isSupported(result.data);
-                  break;
-                case 'news':
-                  supported = await widget.account.client.news.isSupported();
-                  break;
-                case 'notes':
-                  supported = await widget.account.client.notes.isSupported(result.data);
-                  break;
+              final (supported, _) = switch (id) {
+                'core' => await widget.account.client.core.isSupported(result.data),
+                'news' => await widget.account.client.news.isSupported(),
+                'notes' => await widget.account.client.notes.isSupported(result.data),
+                _ => (true, null),
+              };
+              if (supported || !mounted) {
+                return;
               }
-              if (!(supported ?? true)) {
-                if (!mounted) {
-                  return;
-                }
-                var name = AppLocalizations.of(context).appImplementationName(id);
-                if (name == '') {
-                  name = id;
-                }
-                await _showProblem(
-                  AppLocalizations.of(context).errorUnsupportedVersion(name),
-                );
+              var name = AppLocalizations.of(context).appImplementationName(id);
+              if (name == '') {
+                name = id;
               }
+              await _showProblem(
+                AppLocalizations.of(context).errorUnsupportedVersion(name),
+              );
             } catch (e, s) {
               debugPrint(e.toString());
               debugPrint(s.toString());
--- a/packages/nextcloud/lib/src/version_supported.dart
+++ b/packages/nextcloud/lib/src/version_supported.dart
@@ -1,33 +1,56 @@
 part of '../nextcloud.dart';
 
+/// Version of core/Server supported
+const coreSupportedVersion = 26;
+
+/// API version of the news app supported
+const newsSupportedVersion = 'v1-3';
+
+/// API version of the notes app supported
+const notesSupportedVersion = 1;
+
 // ignore: public_member_api_docs
 extension CoreVersionSupported on NextcloudCoreClient {
-  /// Checks if the app on the server is supported by the client
-  Future<bool> isSupported([final NextcloudCoreServerCapabilities_Ocs_Data? capabilities]) async =>
-      (capabilities ?? (await getCapabilities()).ocs.data).version.major == 26;
+  /// Check if the core/Server version is supported by this client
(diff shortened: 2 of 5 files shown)


adil192:
This is perfect

So far the only Dart 3 feature I'm using is not needing `break` statements in switch cases lol

packages/nextcloud/lib/src/version_supported.dart
```diff
@@ -1,33 +1,68 @@
...
+        (capabilities ?? (await getCapabilities()).ocs.data).version.major == coreSupportedVersion.version,
+        coreSupportedVersion,
+      );
 }
 
 // ignore: public_member_api_docs
 extension NewsVersionSupported on NextcloudNewsClient {
   /// Checks if the app on the server is supported by the client
-  Future<bool> isSupported() async {
+  Future<(bool, _SupportedVersion<String>)> isSupported() async {
```
Leptopoda, line 44: Somehow I dislike this interface.

Tuples (or however dart calls them) got introduced so we don't need to create such dummy objects and we still have one here.

Also doesn't the caller already know the app they are testing for?
I'd suggest `(isSupported, version)` or if we really need the app name `(isSupported, version, appName)`

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -60,30 +60,33 @@ class _HomePageState extends State<HomePage> {
             ...appsResult.data!.map((final a) => a.id),
           ]) {
             try {
-              bool? supported;
+              late bool supported;
```
Leptopoda, line 63: dart3 could 
```
token = switch (charCode) {
  slash || star || plus || minus => operator(charCode),
  comma || semicolon => punctuation(charCode),
  >= digit0 && <= digit9 => number(),
  _ => throw invalid()
};
```

Leptopoda (changes requested):
at least the switch could be cleaned up
Also just reviewed online and didn't test locally yet :)

packages/neon/neon/lib/src/pages/home.dart
provokateurin: Ah yes I didn't think about that :)

packages/nextcloud/lib/src/version_supported.dart
provokateurin: Hm yeah you're right, the app doesn't really add any info

packages/nextcloud/lib/src/version_supported.dart
```diff
@@ -1,33 +1,57 @@
...
+  newsSupportedVersion,
+  notesSupportedVersion,
+];
+
 // ignore: public_member_api_docs
 extension CoreVersionSupported on NextcloudCoreClient {
   /// Checks if the app on the server is supported by the client
-  Future<bool> isSupported([final NextcloudCoreServerCapabilities_Ocs_Data? capabilities]) async =>
-      (capabilities ?? (await getCapabilities()).ocs.data).version.major == 26;
+  Future<(bool, int)> isSupported([
```
provokateurin, line 22: The version that is returned is always just the constant and could be queried separately, but I think this interface makes it a little easier to work with.

packages/nextcloud/lib/src/version_supported.dart
```diff
@@ -1,33 +1,57 @@
...
+/// All supported versions
+const supportedVersions = [
+  coreSupportedVersion,
+  newsSupportedVersion,
+  notesSupportedVersion,
+];
+
 // ignore: public_member_api_docs
 extension CoreVersionSupported on NextcloudCoreClient {
   /// Checks if the app on the server is supported by the client
```
Leptopoda, line 21: I know I'm getting nit picky here but the doc comments also need adjustments

packages/nextcloud/lib/src/version_supported.dart
Leptopoda: Yeah but we don't hide it so if somebody wants to look it up separately they can.

packages/nextcloud/lib/src/version_supported.dart
provokateurin: What do you want to have here instead?

packages/nextcloud/lib/src/version_supported.dart
```diff
@@ -1,33 +1,57 @@
...
+const coreSupportedVersion = 26;
+
+/// API version of the news app supported
+const newsSupportedVersion = 'v1-3';
+
+/// API version of the notes app supported
+const notesSupportedVersion = 1;
+
+/// All supported versions
+const supportedVersions = [
```
Leptopoda, line 13: Where do we need this?

I don't think a list is that helpful as here the app name is relevant. `Maybe a Map<appName, versionSupported>` ?

provokateurin: I noticed it too, removed it.

packages/nextcloud/lib/src/version_supported.dart
Leptopoda: I'd mention that it what it returns.
How should I know what the second parameter represents when just using code highlighting in an IDE.
 
 sadly I couldn't find anything in the effective-dart:documentation guide regarding this issue.
 ```dart
 /// returns something cool
(bool issupported, int minVersion) checkSomething() {

}
 ```
 
 Also the IDE does not really highlight the names I gave it in the above example o the best solution might be adding a 'Returns the result of the check and the minimum supported version.'
 
 Or maybe a code example :thinking:

Leptopoda (approved):

Issue #307, opened 2023-05-08
[RFC]: adaptive design ideas


Leptopoda:
While tackling #294 I noticed that our Adaptive design isn't optimal rn.
Mainly that we should use NavigationRail instead of a crippled drawer.

I decided to make use of [flutter_adaptive_scaffold](https://github.com/flutter/packages/tree/main/packages/flutter_adaptive_scaffold) wich is a flutter official package implementing M3 design `AdaptiveLayout`s.

I made a quick POC wich can be found at https://github.com/Leptopoda/nextcloud-neon/tree/cleanup/ResponsiveScaffold (notification counter missing).

Now this showed me that our current design isn't really adaptive and responsive (wich is one of our goals).
The AppBar is fairly useless on non mobile views when using multiple accounts. We show the user avatar two times :woman_shrugging: 

---
Having a look at the M3 design:
https://m3.material.io/foundations/layout/understanding-layout/overview
https://m3.material.io/components/navigation-rail/overview

My idea is to utilize the dual pane design to show a collection of items on one side and an opened item on the other.
Imagine a list of notes/news_articles on the left and the opened article/note on the right.

If we do it right we wouldn't need the BottomNavigationBar and AppBar as view related info from the app bar will be shown on the top of each pane and the bottom bar items in the Navigation rail.

The only problem I have is our usecase of having multiple "apps" with thier own NavigationBar.
While I found one reference in the spec that each pane can have their own BottomBar and  AppBar I doubt this would look good.
Imagine a NavigationRail on the side not related to the view and a half width BottomBar with some items.

Ideally I'd like to replicate the below screenshot where we have a NavigationRail per app that can transform into a BottomNavigationBar on small views. The only problem with this is that we wouldn't have a place for the app/account switcher as the NavigationRail would be app specific.

I'd add those in a HamburgerMenue that opens a drawer on small and medium screens (see below) and sectioned always expanded drawer an large screens (again see below).

<img width="2048" alt="ld4wxkhh-3p_whats new_2x" src="https://user-images.githubusercontent.com/25266387/236845657-35e85e5c-393e-4eca-bd57-ef1644bfd082.png">

I'm no designer but what do you think of this idea?
Can you find a better way that adheres to M3?

---

We'd need to change `AppImplementation` to hold all the extra information and also find a way to talk to the `AdaptiveScaffold` from the apps as they need to manipulate the secondary views.
[...]

PR #329, opened 2023-05-19
RFC: adaptive_layout

Leptopoda:
implements: #307 

This obviously won't be a separate package but it enabled me to iterate faster.
What do you think of it? The `App` class is my equivalent of the `AppImplementation` and you can see with the files app how you don't need to actually make use of the adaptive layout.
In theory an app can also override our default `TabBar` (the app selector) wich might be useful for single app apps (sounds funny).
Yes we are nesting scaffolds :( but that's needed to achieve having both the drawer and fab (I've opened an issue on the flutter side to ask for better documentation wether this is good or bad).

```dart
Scaffold(
  drawer: drawer,
  body: AdaptiveLayout(
    ...: Scaffold(
      floatingActionButton: fab,
    ),
 ),
```

missing:
- better animations
- better fab abstraction
- navigator interception for the second view (in theory we could leave that up to the app as they'll ultimately control everything but we might be able to provide helpful widgets so not everyone needs to reinvent the wheel

Feedback of the current structure is welcome.

---

here a small preview :)
As you can see with the initial wiggle of the window I have some (objectively) cool effects on my windows. This might contribute to wonky animations.

https://github.com/provokateurin/nextcloud-neon/assets/25266387/d6aea36b-a7ca-4070-a9d0-bc39277032ca


provokateurin:
It looks very nice :)
I would move the FAB in the sidebar to the bottom instead of having it at the top, but that is probably just personal preference.
I think for the desktop layout with the tabbar at the top it would be nice to have more visual separation from the rest of the app. In the web UI it's very clear what is the general navigation and what belongs to the app, but here it's harder to tell imo.

provokateurin:
Alright, I haven't had time to review the code yet.

provokateurin:
Do we need to keep this open? If you still want to have this code somewhere then I'd suggest you create a new repo with just the package.

Issue #326, opened 2023-05-18
Local path to `dynamite_runtime` in non-local pubspec


adil192:
When trying to upgrade the `nextcloud` package in Saber (from a525503d2a24ae1840433858b1547c0ca6e23197 to the latest commit), I get this error:
```
$ flutter pub upgrade --major-versions
Error on line 15, column 11: Invalid description in the "nextcloud" pubspec on the "dynamite_runtime" dependency: "../dynamite/dynamite_runtime" is a relative path, but this isn't a local pubspec.
   ╷
15 │     path: ../dynamite/dynamite_runtime
   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
```

provokateurin:
Meh @Leptopoda we forgot to think about that.

provokateurin:
Maybe we should drop the runtime package and just import dynamite in the generated code. I don't think it has much downsides other than adding the dependencies used for the code generator.

Leptopoda:
I don't think this would change anything.

We still import dynamite as a local path so the same error would be shown.

This is only possible rn as we have dynamite as a dev dep but as soon as we bump it to a normal one (what we would need) we get the same error.

Leptopoda:
Actually I did think about that but I must accidentally have added the `publish_to: none` flag to the nextcloud lib.

provokateurin:
Ah yes you are right. I think one possible solution would be to add a `dependency_overrides` for `dynamite_runtime`

provokateurin:
You need to do it like this:
```yaml
dependency_overrides:
  dynamite_runtime:
    git:
      url: https://github.com/provokateurin/nextcloud-neon
      path: packages/dynamite/dynamite_runtime
      ref: COMMIT
```
Replace `COMMIT` with the same commit hash that you use for importing the nextcloud package.

provokateurin:
Ah that sucks. Then the best way right now is probably to add it as a submodule and require the package by path.

provokateurin:
@Leptopoda we could also eliminate the runtime package and generate the content string things for the unit tests separately.

Leptopoda:
I just had a look at how other repos handle this:
https://github.com/google/built_value.dart/blob/master/built_value_generator/pubspec.yaml

It looks like they use a real dependency (in our case we'd use a git dependency) and then override the dep with a local path.

[...]

PR #333, opened 2023-05-21
neon, nextcloud, app: don't use local dependencies

Leptopoda:
fixes: #326 

The only downside of this is that we don't pin dependencies. This means someone could use the nextcloud lib pinned to a specific commit but  get the latest dynamite_runtime.
Worst case would be that consumers would need to add an override that also pins  dynamite_runtime.

Could @adil192 or @muelleel please test and provide feedback :)

### DIFF
--- a/packages/app/pubspec.lock
+++ b/packages/app/pubspec.lock
@@ -146,7 +146,7 @@ packages:
     source: hosted
     version: "0.7.8"
   dynamite_runtime:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       path: "../dynamite/dynamite_runtime"
       relative: true
@@ -177,7 +177,7 @@ packages:
     source: hosted
     version: "6.1.4"
   file_icons:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       path: "../file_icons"
       relative: true
@@ -534,7 +534,7 @@ packages:
     source: hosted
     version: "1.0.0"
   nextcloud:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       path: "../nextcloud"
       relative: true
@@ -790,7 +790,7 @@ packages:
     source: hosted
     version: "0.1.6"
   settings:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       path: "../settings"
       relative: true
@@ -882,7 +882,7 @@ packages:
     source: sdk
     version: "0.0.99"
   sort_box:
-    dependency: transitive
+    dependency: "direct overridden"
     description:
       path: "../sort_box"
       relative: true
--- a/packages/app/pubspec.yaml
+++ b/packages/app/pubspec.yaml
@@ -11,15 +11,25 @@ dependencies:
     sdk: flutter
   flutter_native_splash: ^2.2.19
   neon:
-    path: ../neon/neon
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon
   neon_files:
-    path: ../neon/neon_files
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon_files
   neon_news:
-    path: ../neon/neon_news
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon_news
   neon_notes:
-    path: ../neon/neon_notes
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon_notes
   neon_notifications:
-    path: ../neon/neon_notifications
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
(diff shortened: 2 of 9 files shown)


packages/app/pubspec.lock
```diff
@@ -148,7 +148,7 @@ packages:
   dynamite_runtime:
     dependency: transitive
     description:
-      path: "../dynamite/dynamite_runtime"
+      path: "../../../../../../.pub-cache/git/nextcloud-neon-482bff006d4338964c0704600c50434fe76a19b2/packages/dynamite/dynamite_runtime"
```
Leptopoda, line 151: that's odd as it references my home folder :/

Probably a pub bug.

Leptopoda: Maybe we need to also add an override for dynamite_runtime to app.

This probably references my home dir as the git dependency gets cached there. A perfect example of the above mentioned problem.

Leptopoda: Looks like the CI agrees with me

provokateurin: Yeah the other packages also need the override, otherwise they won't use the local version.

Leptopoda: That's odd. I just added an override and both `flutter analyze` and `flutter pub get` work in the `packages/app` dir and they work fine.

adil192 (commented):
`flutter pub get` works on:
```yaml
dependencies:
  nextcloud:
    git:
      url: https://github.com/Leptopoda/nextcloud-neon
      path: packages/nextcloud
      ref: fix/#326
```
and results in this `pubspec.lock`:
```yaml
  dynamite_runtime:
    dependency: transitive
    description:
      path: "packages/dynamite/dynamite_runtime"
      ref: HEAD
      resolved-ref: "482bff006d4338964c0704600c50434fe76a19b2"
      url: "https://github.com/provokateurin/nextcloud-neon"
    source: git
    version: "1.0.0"
  nextcloud:
    dependency: "direct main"
    description:
      path: "packages/nextcloud"
      ref: "fix/#326"
      resolved-ref: "9197bf208215135605d48499f7ab562b9145aea8"
      url: "https://github.com/Leptopoda/nextcloud-neon"
    source: git
    version: "1.0.0"
```

Leptopoda:
That means it works as expected.

Could you also try to add a dependency override to `dynamite_runtime` pinning it to the same commit as nextcloud?

adil192:
> Could you also try to add a dependency override to `dynamite_runtime` pinning it to the same commit/reference as nextcloud?

```yaml
dependency_overrides:
  dynamite_runtime:
    git:
      url: https://github.com/Leptopoda/nextcloud-neon
      path: packages/dynamite/dynamite_runtime
      ref: fix/#326
```
results in
```yaml
  dynamite_runtime:
    dependency: "direct overridden"
    description:
      path: "packages/dynamite/dynamite_runtime"
      ref: "fix/#326"
      resolved-ref: "9197bf208215135605d48499f7ab562b9145aea8"
      url: "https://github.com/Leptopoda/nextcloud-neon"
    source: git
    version: "1.0.0"
```

Leptopoda:
That's great to see.

Now I only need to find out how to make CI happy :)

Leptopoda:
@provokateurin do you have any idea?

`mono_repo presubmit` does work fine. Even on a fresh clone :(
Can this be some caching issue with the deps in CI?

provokateurin:
Let me give it a try

provokateurin (changes requested):

Leptopoda:
yeah and in theory all other apps (neon, neon_files, ...) also need the override for settings, sortbox ...
It's just not a problem as we don't check in the lock file.

This is getting tedious.

provokateurin:
I think it is still a problem regardless of the lockfile. Otherwise the source is pulled from the wrong place, no?

Leptopoda:
Yep. I only meant that CI didn't catch that as it just didn't care.

packages/app/pubspec.yaml
```diff
@@ -11,15 +11,25 @@ dependencies:
...
   neon_notes:
-    path: ../neon/neon_notes
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon_notes
   neon_notifications:
-    path: ../neon/neon_notifications
+    git:
+      url: https://github.com/provokateurin/nextcloud-neon
+      path: packages/neon/neon_notifications
```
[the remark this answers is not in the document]
Leptopoda: Now I didn't change this one.

What do you think?

Leptopoda:
I'll look into it :)

packages/app/pubspec.yaml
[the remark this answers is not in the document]
provokateurin: Let's keep it this way. I haven't full thought about how other apps will use neon from the outside.

Leptopoda:
implemented through #336

PR #337, opened 2023-05-22
neon: cleanup account handling

Leptopoda:
debugging the old implementation was not fun :)
I'd argue that the new one is also easier to read and less error prone as we filter out equal accounts on the stream level (`activeAccount..distinct()`) so we can't forget to check :)


provokateurin:
Ah yes, that was one of the first pieces of code of this project, so it carrys a lot of bad code. Thanks for the cleanup, I didn't know about the distinct thing!

provokateurin (approved):

Leptopoda:
I tried a bit implementing unit tests for our blocs so we as I noticed it's currently a bit fragile but it's really hard as they depend on each other really good and mocking everything results in very much boilerplate code (even with mocking libs like mocktail or mockito).

Long term we might want to change that or find a good approach for testing.

Leptopoda:
The problem was actually `GlobalOptions` as that was hard to mock with all it's ToggleOptions.

Issue #272, opened 2023-04-22
Use nested router for navigation


provokateurin:
There are some nasty bugs when popping pages on top of the files browser, but they should be fixable by using another router for the files browser. That should also improve performance when going back a directory, because it doesn't need to be loaded again.

Leptopoda:
How did you imagine this being done?

I think we'll need something similar for a nice implementation of #329. My idea was to use [`go_router`](https://pub.dev/packages/go_router) for routing.

The neon would register a router for things like login, settings ... while every app can itself provide a custom router.
This nested router could then be used to navigate within the app (needed for files and the second panel in a multi layout view).

According to my research go_router can handle both use cases with a [ShellRoute](https://pub.dev/documentation/go_router/latest/go_router/ShellRoute-class.html).

1. can we bump this to a global neon issue?
2. if you like the go_router approach I can have a shot at it :)

provokateurin:
Sure thing!

provokateurin:
@Leptopoda This is done, right?

Leptopoda:
No; only the foundation for it.
I'll tackle this with the adaptive stuff coming soon TM

provokateurin:
Right, can't wait for that :)

The files app should be restructured because the file browser is a single widget right now, but it should be pages that open on top of each other. In the current implementation the whole page also reloads if you go back a folder.

PR #338, opened 2023-05-22
Feature/routing

Leptopoda:
first step for: #272 
and in turn for #307

Currently when multiple accounts are logged in and logging out one of them through the avatar in the appBar it will pop you back to the settings page instead of the home screen.

I thought this is acceptable for now as I want to make this avatar an account switcher only and open the settings for login/logout anyways with the adaptive design.

### DIFF
--- a/packages/neon/neon/lib/src/pages/account_settings.dart
+++ b/packages/neon/neon/lib/src/pages/account_settings.dart
@@ -29,8 +29,6 @@ class AccountSettingsPage extends StatelessWidget {
                   AppLocalizations.of(context).accountOptionsRemoveConfirm(account.client.humanReadableID),
                 )) {
                   bloc.removeAccount(account);
-                  // ignore: use_build_context_synchronously
-                  Navigator.of(context).pop();
                 }
               },
               tooltip: AppLocalizations.of(context).accountOptionsRemove,
--- a/packages/neon/neon/lib/src/pages/home.dart
+++ b/packages/neon/neon/lib/src/pages/home.dart
@@ -197,376 +197,378 @@ class _HomePageState extends State<HomePage> {
                       builder: (final context) {
                         if (accountsSnapshot.hasData) {
                           final accounts = accountsSnapshot.data!;
-                          final account = accounts.find(_account.id)!;
-
-                          final isQuickBar = navigationMode == NavigationMode.quickBar;
-                          final drawer = Builder(
-                            builder: (final context) => Drawer(
-                              width: isQuickBar ? kQuickBarWidth : null,
-                              child: Container(
-                                padding: isQuickBar ? const EdgeInsets.all(5) : null,
-                                child: Column(
-                                  children: [
-                                    Expanded(
-                                      child: Scrollbar(
-                                        controller: drawerScrollController,
-                                        interactive: true,
-                                        child: ListView(
+                          final account = accounts.find(_account.id);
+                          if (account != null) {
+                            final isQuickBar = navigationMode == NavigationMode.quickBar;
+                            final drawer = Builder(
+                              builder: (final context) => Drawer(
+                                width: isQuickBar ? kQuickBarWidth : null,
+                                child: Container(
+                                  padding: isQuickBar ? const EdgeInsets.all(5) : null,
+                                  child: Column(
+                                    children: [
+                                      Expanded(
+                                        child: Scrollbar(
                                           controller: drawerScrollController,
-                                          // Needed for the drawer header to also render in the statusbar
-                                          padding: EdgeInsets.zero,
-                                          children: [
-                                            Builder(
-                                              builder: (final context) {
-                                                if (accountsSnapshot.hasData) {
-                                                  if (isQuickBar) {
-                                                    return Column(
-                                                      children: [
-                                                        if (accounts.length != 1) ...[
-                                                          for (final account in accounts) ...[
-                                                            Container(
-                                                              margin: const EdgeInsets.symmetric(
-                                                                vertical: 5,
-                                                              ),
-                                                              child: IconButton(
-                                                                onPressed: () {
-                                                                  _accountsBloc.setActiveAccount(account);
-                                                                },
-                                                                tooltip: account.client.humanReadableID,
-                                                                icon: IntrinsicHeight(
-                                                                  child: NeonAccountAvatar(
-                                                                    account: account,
+                                          interactive: true,
+                                          child: ListView(
+                                            controller: drawerScrollController,
+                                            // Needed for the drawer header to also render in the statusbar
+                                            padding: EdgeInsets.zero,
+                                            children: [
+                                              Builder(
+                                                builder: (final context) {
+                                                  if (accountsSnapshot.hasData) {
+                                                    if (isQuickBar) {
+                                                      return Column(
+                                                        children: [
+                                                          if (accounts.length != 1) ...[
+                                                            for (final account in accounts) ...[
+                                                              Container(
+                                                                margin: const EdgeInsets.symmetric(
+                                                                  vertical: 5,
+                                                                ),
+                                                                child: IconButton(
+                                                                  onPressed: () {
+                                                                    _accountsBloc.setActiveAccount(account);
+                                                                  },
+                                                                  tooltip: account.client.humanReadableID,
+                                                                  icon: IntrinsicHeight(
+                                                                    child: NeonAccountAvatar(
+                                                                      account: account,
+                                                                    ),
                                                                   ),
                                                                 ),
                                                               ),
-                                                            ),
-                                                          ],
-                                                          Container(
-                                                            margin: const EdgeInsets.only(
-                                                              top: 10,
-                                                            ),
-                                                            child: Divider(
-                                                              height: 5,
-                                                              color: Theme.of(context).appBarTheme.foregroundColor,
-                                                            ),
-                                                          ),
-                                                        ],
-                                                      ],
-                                                    );
-                                                  }
-                                                  return DrawerHeader(
-                                                    decoration: BoxDecoration(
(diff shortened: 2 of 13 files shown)


packages/neon/neon/lib/src/blocs/accounts.dart
provokateurin, line 1: I think it would be better to move the comments to the abstract classes. The inheritors will automatically have the documentation too.

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -139,11 +157,35 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
     setActiveAccount(account);
   }
 
+  /// The currently active account.
+  ///
+  /// Equivalent to activeAccount.value but throws a [StateError] when not user is logged in.
+  @visibleForTesting
+  Account get aa {
```
provokateurin, line 164: Could this just be private?

packages/neon/neon/lib/src/blocs/accounts.dart
provokateurin, line 1: Wouldn't it be easier to make the account an optional parameter and the methods just select the active account if no account is passed? One downside I see is that it could lead to mistakes where no account is passed while another account than the active one should be used.

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -114,10 +128,14 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
...
   }
 
+  /// Sets the active account.
+  ///
+  /// If no [account] is specified it will fallback to the first account in [accounts].
+  /// When no account is available [activeAccount] will be set to null triggering a logout.
   @override
-  void setActiveAccount(final Account? account) {
-    final as = accounts.value;
-    activeAccount.add(account ?? as.firstOrNull);
```
provokateurin, line 120: I don't understand why this is here in the first place. If a null account is passed then it should probably just set to be null.

provokateurin (changes requested):

packages/neon/neon/lib/src/blocs/accounts.dart
Leptopoda: It basically is private.

Also as every dart file we have is `part of` neon there isn't anything private in our code anyways.

Also we want to unit test it eventually so why not from the beginning?

provokateurin: But the method could be private to the class, right? I agree for testing it is useful but then it should have a better name.

packages/neon/neon/lib/src/blocs/accounts.dart
Leptopoda: I just added the documentation in here.
Probably should be a separate pr anyways.

I quite like the distinction and I'd like changing the name of the methods to something like `getValueFor` indicating it's for a specific account and having the getter otherwise

packages/neon/neon/lib/src/blocs/accounts.dart
Leptopoda: In dart there isn't any concept of private to the class.
In dart things with an underscore are private to it's package (aka it's file). So visibleForTesting == leadingUnderscore (only difference is that one can be used in tests).

If you really want the 'private' aspect we should rethink having all files `part`ed together.

packages/neon/neon/lib/src/blocs/accounts.dart
Leptopoda: You're right, Just looking at it here on GH we could probably remove it.

Ideally I'd like a private method `_setActiveAccount(final Account? account)` and a public one `setActiveAccount(final Account account)` this way we can still log users out internally of the class by passing null to the private method but external callers can only switch the account and would need to call `removeAccount` to achieve logouts. Otherwise someone could use `setActiveAccount` for logout wich is just wrong.

Leptopoda:
We are getting fairly off topic now :)

I'll pull out the AuthBloc changes and we can discuss this further in a separate PR/Issue or over on matrix.
All comments from the review are related to that and that's nice to here as this means the routing stuff is not to bad.

packages/neon/neon/lib/src/blocs/accounts.dart
provokateurin: Oh you are teaching me new dart things everyday :smile:
I always thought that there are private methods for classes...

provokateurin:
Yes splitting makes sense.

packages/neon/neon/lib/src/blocs/accounts.dart
provokateurin: Yeah I just tried it myself already

packages/neon/neon/lib/src/router.dart
```diff
@@ -1,41 +1,118 @@
 part of '../neon.dart';
```
provokateurin, line 1: As this file will generate some code I'd say let's detach it from the main file.

provokateurin (changes requested):

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -129,13 +127,7 @@ class _HomePageState extends State<HomePage> {
...
   }
 
-  Future _openSettings() async {
-    await Navigator.of(context).push(
-      MaterialPageRoute(
-        builder: (final context) => const SettingsPage(),
-      ),
-    );
-  }
+  void _openSettings() => const SettingsRoute().go(context);
```
provokateurin, line 130: This could be inlined as it's quite simple now

packages/neon/neon/lib/src/router.dart
```diff
@@ -1,41 +1,124 @@
...
+@TypedGoRoute<HomeRoute>(
+  path: '/',
+  name: 'home',
+  routes: [
+    TypedGoRoute<SettingsRoute>(
+      path: 'settings',
+      name: 'Settings',
+      routes: [
+        TypedGoRoute<NextcloudAppSettingsRoute>(
+          path: ':appid',
```
provokateurin, line 63: Account has `account/:accountid` so better to also make this `app/:appid`.

provokateurin (changes requested):

packages/neon/neon/lib/src/router.dart
Leptopoda: Yeah but isn't that a bit redundant?

`/settings/apps/appID`
I just added the accounts as
`/settings/hashID` is not that meaningful.

While currently not our priority the routes will also represent the url if run in the browser.

provokateurin: Hm yeah if you think about it like that then it makes sense.

provokateurin: Yeah also true... just leave it as is :)

provokateurin:
I still wanted to try out the changes myself, but ok :/

Leptopoda:
Oh sorry should I revert?

Github gave me green checkmarks and I thought you reviewd it.

provokateurin:
No worries, I'll still try it out and report if something doesn't work.

provokateurin:
I could turn on the requirement for at least one approval for a PR to be allowed to merge.

Leptopoda:
would make sense but might be a problem when you need to merge something on your own and I can't review it in time.

And you probably don't want to turn this on and of all the time. 
At least now I'm aware of the bad UI.

provokateurin:
I think as the owner of the repo I would still be allowed to merge whenever I want. I'll just keep it as is and we can enable it later on if necessary.

packages/neon/neon/lib/src/pages/settings.dart
```diff
@@ -181,26 +175,15 @@ class _SettingsPageState extends State<SettingsPage> {
...
                         CustomSettingsTile(
                           title: ElevatedButton.icon(
-                            onPressed: () async {
-                              await Navigator.of(context).push(
-                                MaterialPageRoute(
-                                  builder: (final context) => const LoginPage(),
-                                ),
-                              );
+                            onPressed: () {
+                              const LoginRoute().go(context);
```
provokateurin, line 186: This doesn't work. It just goes back to the home page and the login page doesn't show up.

packages/neon/neon/lib/src/router.dart
```diff
@@ -1,41 +1,124 @@
...
+    required this.accountid,
   });
 
-  final AccountsBloc accountsBloc;
+  final String accountid;
+
+  @override
+  Widget build(final BuildContext context, final GoRouterState state) {
+    final bloc = Provider.of<AccountsBloc>(context, listen: false);
+    final account = bloc.accounts.value.find(accountid)!;
```
provokateurin, line 45: When I have multiple accounts and remove the one that is currently active a null exception happens here.

packages/neon/neon/lib/src/pages/account_settings.dart
```diff
@@ -29,8 +29,6 @@ class AccountSettingsPage extends StatelessWidget {
                   AppLocalizations.of(context).accountOptionsRemoveConfirm(account.client.humanReadableID),
                 )) {
                   bloc.removeAccount(account);
-                  // ignore: use_build_context_synchronously
-                  Navigator.of(context).pop();
```
provokateurin, line 33: When you have multiple accounts and remove the one that isn't currently the active one, then the account settings page of the account you just removed stays open.

provokateurin (commented):
I haven't tested everything, but found a few major problems. Please fix them in a new PR

Leptopoda:
2 & 3 are related and will probably be fixed by re adding the pop.
1 might be tricky but should be doable.


I'll fix them asap when I have a bit of time.
Sorry again for merging :(

provokateurin:
No problem :) I won't be around this weekend, so expect the reviews mid next week

PR #339, opened 2023-05-22
Cleanup parts for generated code

provokateurin:
I didn't know this was possible until I saw it in https://github.com/provokateurin/nextcloud-neon/pull/338 :)


Leptopoda (commented):
I personally am not the fan of having everything in one package.
https://github.com/provokateurin/nextcloud-neon/pull/338#discussion_r1200989175 is also related.

It's not the dart way and we completely eliminate the concept of `private` things (although they aren't really private as mentioned in my other comment).

---

I'll only leave a comment here as it's probably up to you to decide what way you want things to go.

Please ping me again if you want to continue this way and I'll thoroughly review.

provokateurin:
I understand your concern. Maybe then we should start splitting things more? That will probably help with the architecture and scaling for the future. @Leptopoda let me know if that is the way you want to go.

PR #342, opened 2023-05-28
Fix/go router

Leptopoda:
followup to #338 

I tested:
- 2 accounts logout the active one
- 2 accounts logout the not active one
- 1 account logout through the settings
- 1 account logout through the appBar account overview thing
- LoginPage pop the active page when adding an account (I think you didn't report that one but I found a small issue)

Problem/inconvenience I found:
when adding an account and entering the server url the vebview opens. Pressing the back button here does abort the AddAccount flow. It'd be better if we handle this similar to the login flow (first account) where the back button clears the server Url and we get to re enter it.

As this existed before I'll open an issue after this one is merged. We should discuss how we handle this as it might make sense to just externalize the vebview into it's own page cleaning up the code a bit.


provokateurin:
Let's discuss the login flow in detail, as I've been changing it a bit for the QR Code thingy, but then it got messy.

Leptopoda:
I've opened #343 for that :)

packages/neon/neon/lib/src/pages/account_settings.dart
```diff
@@ -29,6 +29,8 @@ class AccountSettingsPage extends StatelessWidget {
                   AppLocalizations.of(context).accountOptionsRemoveConfirm(account.client.humanReadableID),
                 )) {
                   bloc.removeAccount(account);
+                  // ignore: use_build_context_synchronously
+                  const HomeRoute().go(context);
```
provokateurin, line 33: Would it be possible to have this logic in the router? When the active account changes then the page switches to the home page but also pops all other routes on top.

packages/neon/neon/lib/src/router.dart
```diff
@@ -60,13 +56,17 @@ class AccountSettingsRoute extends GoRouteData {
...
+          path: 'apps/:appid',
           name: 'NextcloudAppSettings',
         ),
         TypedGoRoute<AccountSettingsRoute>(
           path: 'account/:accountid',
           name: 'AccountSettings',
         ),
+        TypedGoRoute<AddAccountRoute>(
+          path: 'addAccount',
+          name: 'addAccount',
```
provokateurin, line 68:
```suggestion
          path: 'account/add',
          name: 'AddAccount',
```

The account ID can never be `add` so it should be fine and also looks a lot better imo.

packages/neon/neon/lib/src/router.dart
```diff
@@ -98,6 +98,14 @@ class LoginRoute extends GoRouteData {
   Widget build(final BuildContext context, final GoRouterState state) => LoginPage(serverURL: server);
 }
 
+@immutable
+class AddAccountRoute extends GoRouteData {
+  const AddAccountRoute();
+
+  @override
+  Widget build(final BuildContext context, final GoRouterState state) => const LoginPage();
+}
```
provokateurin, line 107: Why is this separate route needed? I see it hides the serverURL paramter, but as far as I can tell it doesn't have any other effect?

Leptopoda: So we can have a backbutton and discard the login.

I guess that the backbutton calls Navigstor.of(context).pop() internally and /login is a top level route so we can't pop from it.

packages/neon/neon/lib/src/router.dart
Leptopoda: I started with it but I had a problem where it would still activate the accountsettings with add as an account id and then failing to find the account.

Having the addAccount route before the accountSettings one in the routes list would probably fix this (didn't check that prossibility).

That's also the reason why I moved `settings/:appID` to `settings/apps/:appID`.
I also prefer the other way so I'll look into it.

packages/neon/neon/lib/src/router.dart
provokateurin: Ah that makes sense. Thanks for the explanation!

packages/neon/neon/lib/src/pages/account_settings.dart
provokateurin: This is how it worked before the whole navigator/router changes (not your go_router changes).

Leptopoda: I think that I tried something similar but will also double check.
I think we could make a followup if it requires to much changes.

provokateurin: Ok. I'd like to have this as it will prevent problems in the future if another place changes the active account but forgets to trigger the home page.

packages/neon/neon/lib/src/router.dart
Leptopoda: Ok as expected this does work.

Should I also revert the `settings/apps/:appID` or not?

packages/neon/neon/lib/src/pages/account_settings.dart
Leptopoda: I'm not sure I fully understand what you want.
- If we have only one logged in account and the user loggs out we want the login page to appear
- if we have multiple logged in accounts and either of them loggs out we want the settings page to be shown (just popping the top route)

or is there any case where we want the user to be navigated to the HomePage?

I think it's fairly trivial to implement in any way so I'll add it to this PR

packages/neon/neon/lib/src/router.dart
provokateurin: Great, but let's stick to the current way

packages/neon/neon/lib/src/pages/account_settings.dart
provokateurin: If there are multiple accounts and the active account gets logged out it should show the HomePage.
If there are multiple accounts and but it's not the active account that is logged out then it should just show the SettingsPage.

I think you can also argue that it should always show the SettingsPage, but this UX feels better to me as the HomePage will always be shown if the active account changes (as it happens for manual account switching).

packages/neon/neon/lib/src/router.dart
Leptopoda: current being the state of this PR or the state of master?
I'm a bit confused.

provokateurin: This PR, sorry for the confusion.

provokateurin (approved):
Nice work!

Leptopoda:
This time merging with a review :)

PR #344, opened 2023-05-31
Refactor user statuses and avatars

provokateurin:
Changes from the feature/spreed branch.


packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+        ],
+      ],
+    );
+  }
+
+  Widget _userStatusIcon({
+    required final bool loading,
+    required final Object? error,
+    required final _UserStatus? userStatus,
+  }) {
```
Leptopoda, line 136: Can we rename this to `_userStatusIconBuilder` and change it's interface to the one  `ResultBuilder.builder` expects? Just taking the userStaus and a context.

all the other stuff can be done in the method body. This way we can replace this with a tearoff and I'm sure this would improve code readability quite a bit (also removes the above duplication that should be gone anyways.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+    final hasEmoji = userStatus?.icon != null;
+    final factor = hasEmoji ? 2 : 3;
+    return SizedBox(
+      height: widget.size,
+      width: widget.size,
```
Leptopoda, line 141: SizedBox.square

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+      child: Align(
+        alignment: Alignment.bottomRight,
+        child: Container(
+          height: widget.size / factor,
+          width: widget.size / factor,
```
Leptopoda, line 146: Maybe also use SizedBox.square here and a sepatate DecoratedBox?

Directly reading that it is squared is easier to understand the layout.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+          decoration: loading || error != null || userStatus == null || hasEmoji
```
Leptopoda, line 147: What is this checking?
I have no idea; we should extarnalize this into a meaningful variable

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+                      color: Colors.red,
+                    )
+                  : hasEmoji
+                      ? Text(
+                          userStatus!.icon!,
+                          style: const TextStyle(
+                            fontSize: 16,
+                          ),
+                        )
+                      : null,
```
Leptopoda, line 171: Also quite hard to read.

Maybe have an external `child` variable before the return statement that gets conditionally assigned.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+                ),
+          child: loading
+              ? CircularProgressIndicator(
+                  strokeWidth: 1.5,
+                  color: Theme.of(context).colorScheme.onPrimary,
+                )
+              : error != null && (error is! NextcloudApiException || error.statusCode != 404)
+                  ? Icon(
+                      Icons.error_outline,
+                      size: widget.size / factor,
```
Leptopoda, line 161: duplicated `widget.size / factor` might make sense externalizing it into a var

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+        return const Color(0xFF49B382);
+      case NextcloudUserStatusType.away:
+        return const Color(0xFFF4A331);
+      case NextcloudUserStatusType.dnd:
+        return const Color(0xFFED484C);
+      default:
+        return Colors.transparent;
+    }
+  }
+}
```
Leptopoda, line 189: Can we make this one of the new cool Dart3 switch things?

We should be able to also inline this into a variable above or into the widget itself (variable preferred)

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+
+class _UserStatus {
+  _UserStatus({
+    required this.icon,
+    required this.status,
+  });
+
+  final String? icon;
+  final NextcloudUserStatusType status;
+}
```
Leptopoda, line 199: Looks quite a bit like a use case of Dart3 records (or however those tuples are called. I should really memorize this)

Leptopoda (changes requested):
Sorry for the strict review :/

I know this has been rebased hundreds of times evolved from old code over the time and is probably also quite old code to begin with.

Nevertheless I treated it as being new code thus the quite harsh review.
Some stuff like the common neon theme might not be 100% on topic here and we can move it to a separate issue but I noticed it so I marked it.

I hope my opinion makes somewhat sense

packages/neon/neon/lib/src/utils/timer.dart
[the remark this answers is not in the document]
provokateurin: TimerBloc sounds good

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+// ^ This is a really strange false positive, it goes of at a very random place without any meaning. Hopefully fixed soon?
+
+part of '../../neon.dart';
+
+const kAvatarSize = 40.0;
+
+class NeonUserAvatar extends StatefulWidget {
+  const NeonUserAvatar({
+    required this.account,
+    this.username,
```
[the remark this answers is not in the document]
provokateurin: No, because we want to display avatars for other users. The account is used for fetching from the server.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
+// ignore_for_file: use_late_for_private_fields_and_variables
+// ^ This is a really strange false positive, it goes of at a very random place without any meaning. Hopefully fixed soon?
+
+part of '../../neon.dart';
+
+const kAvatarSize = 40.0;
```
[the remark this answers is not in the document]
provokateurin: I agree, but what's the best way to have such constants? I would be nice to just rely on some value from M3, e.g. the size of a FAB or similar.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,199 @@
...
+                } else {
+                  return widget.account.client.core.getAvatar(
+                    userId: effectiveUsername,
+                    size: size,
+                  );
+                }
+              },
+            ),
+          ),
+        ),
```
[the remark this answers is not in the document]
provokateurin: wdym? The refactoring was the addition that now the avatar and status can be fetched for any users, not only the current account.

packages/neon/neon/lib/src/widgets/user_avatar.dart
provokateurin: Oh yeah that's neat :)

packages/neon/neon/lib/src/widgets/user_avatar.dart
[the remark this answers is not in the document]
Leptopoda: Totally forgot of the spreed usecase where this came from :woman_facepalming: 

could we make it something like:

```dart
  NeonUserAvatar({
    required this.account,
    final String? username,
    this.showStatus = true,
    this.size = kAvatarSize,
    super.key,
  }) : username = username ?? account.client.username!;

  final Account account;
  final String username;
  final bool showStatus;
  final double size;
```

I don't like havint username and effectiveUsername. Yes we lose the const constructor but I'd argue that we wouldn't even use that as the passed required account isn't ever const either.

packages/neon/neon/lib/src/widgets/user_avatar.dart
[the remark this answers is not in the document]
Leptopoda: A custom theme extension https://api.flutter.dev/flutter/material/ThemeExtension-class.html

They are a custom theme that can be added to any `ThemeData`. Adding it to the top level one in `MaterialApp` will let us access our custom theme wherever we have a context through `Theme.of(context).extensiont<NeonTheme>()` (syntax might be off as I wrote this from the top of my head)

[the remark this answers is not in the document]
Leptopoda: I already acknowledged that this might be better suited in a separate issue/PR

packages/neon/neon/lib/src/widgets/user_avatar.dart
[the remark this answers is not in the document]
Leptopoda: Oh sorry for not being clear. This is two times exactly the same code and even in a stack. 
So you are stacking the same thing on itself  and I can't see a purpose.

I thought this might have been a rebase gone rouge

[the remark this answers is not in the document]
provokateurin: Ah I didn't look at it, but it's probably a bad rebase indeed

Leptopoda:
Looks quite good overall and I like the TimerBloc.

Can we somehow return something similar to the `StreamSubscription` that can be used to cancel a specific callback? Should only need to hold the duration and index of that specific timer. (We should then also use something other than a list to avoid getting holes/emtpy spots in there. Maybe a set?)

I still have a few small cleanup things but can't do a proper review on the phone :)

provokateurin:
Yeah canceling was something I thought about, but wasn't sure how to handle.
I also want to add unit tests for the TimerBloc before merging (I would also create some for the other blocs but we don't have the setup for that right now :/ ).

packages/neon/neon/lib/src/utils/bloc.dart
```diff
@@ -1,7 +1,7 @@
 part of '../../neon.dart';
 
 abstract class Bloc {
-  void dispose();
+  void dispose() {}
```
Leptopoda, line 4: what's the reason?

Ideally I'd like to mark both `Bloc` as an interface with `abstract interface class ...` unless it isn't really an interface :thinking:

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,127 @@
...
+          child: DecoratedBox(
+            decoration: decoration ?? const BoxDecoration(),
+            child: child,
+          ),
+        ),
+      ),
+    );
+  }
+
+  Color _userStatusToColor(final NextcloudUserStatusType userStatusType) => switch (userStatusType) {
```
Leptopoda, line 121: Can we inline this into the line 102?

Or maybe we'll leave it as is and incorporate it into our `NeonTheme` once we have it?

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,45 @@
...
+
+  final Map<int, Timer> _timers = {};
+  final Map<int, List<VoidCallback>> _callbacks = {};
+
+  @override
+  void dispose() {
+    for (final timer in _timers.values) {
+      timer.cancel();
+    }
+    super.dispose();
```
Leptopoda, line 29: Why do we need this? isn't `Bloc` just an interface?

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,127 @@
...
+    if (result.loading) {
+      child = CircularProgressIndicator(
+        strokeWidth: 1.5,
+        color: Theme.of(context).colorScheme.onPrimary,
+      );
+    } else if (result.error != null) {
+      child = Icon(
+        Icons.error_outline,
+        size: scaledSize,
+        color: Colors.red,
```
Leptopoda, line 90: `Theme.of(context).colorScheme.errorContainer` or `Theme.of(context).colorScheme.error`

Probably also for the theme stuff once it comes.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,127 @@
...
+      dimension: widget.size,
+      child: Align(
+        alignment: Alignment.bottomRight,
+        child: SizedBox.square(
+          dimension: scaledSize,
+          child: DecoratedBox(
+            decoration: decoration ?? const BoxDecoration(),
+            child: child,
+          ),
+        ),
```
Leptopoda, line 116: can we make this:

```dart
        child: Container(
          width: scaledSize,
          height: scaledSize,
          decoration: decoration,
          child: child,
        ),
```

We lose the square information but it shows that this is one container that's aligned to the bottom and has a decoration (also the decoration can be null avoiding confusion about that).

I'm not 100% sure if it improves readability or not xD

Leptopoda (approved):
Finally also got to testing this.
Other than the ability to cancel timers the PR is mergable  :+1:

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,127 @@
...
+      ),
+    );
+  }
+
+  Color _userStatusToColor(final NextcloudUserStatusType userStatusType) => switch (userStatusType) {
+        NextcloudUserStatusType.online => const Color(0xFF49B382),
+        NextcloudUserStatusType.away => const Color(0xFFF4A331),
+        NextcloudUserStatusType.dnd => const Color(0xFFED484C),
+        _ => Colors.transparent,
```
Leptopoda, line 125: Just forgot to mention:
we should probably return null here.

packages/neon/neon/lib/src/utils/bloc.dart
provokateurin: I don't think every bloc needs to be disposable. This no longer forces inheritors to implement the function. Adding the interface modifier seems like a good idea though

packages/neon/neon/lib/src/blocs/timer.dart
provokateurin: True, just did it for cleanness (that's not a word lol)

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @override
+  void registerTimer(final Duration duration, final VoidCallback callback) {
```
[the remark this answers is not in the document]
provokateurin: That is correct, but it feels odd to have it here. I'd rather keep it like this and call the callback in the specific cases where needed.

packages/neon/neon/lib/src/widgets/user_avatar.dart
provokateurin: I'm not sure if this will be used somewhere else or not. Maybe for setting the status? Though adding it to the theme doesn't make a lot of sense to me (the color values are pulled from the web UI).

provokateurin: I think I would just leave it here as is and move it somewhere else when needed.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    return SizedBox.square(
+      dimension: widget.size,
+      child: Align(
+        alignment: Alignment.bottomRight,
+        child: Container(
+          width: scaledSize,
+          height: scaledSize,
+          decoration: decoration ?? const BoxDecoration(),
```
Leptopoda, line 113: decoration is optional thus the `?? const BoxDecoration()` can be ommited

packages/neon/neon/lib/src/blocs/user_statuses.dart
```diff
@@ -0,0 +1,87 @@
...
+    this._platform,
+    this._account,
+  ) {
+    unawaited(refresh());
+    TimerBloc().registerTimer(const Duration(minutes: 5), refresh);
+  }
+
+  final NeonPlatform _platform;
+  final Account _account;
+  late Timer _timer;
```
Leptopoda, line 22: Does not look like this gets assigned somewhere.

packages/neon/neon/lib/src/blocs/user_statuses.dart
```diff
@@ -0,0 +1,87 @@
...
+  @override
+  Future load(final String username, {final bool force = false}) async {
+    if (!force && _statuses.containsKey(username)) {
+      return;
+    }
+
+    try {
+      _updateStatus(username, Result.loading());
+
+      late NextcloudUserStatusPublicStatus? data;
```
Leptopoda, line 51: This does not need to be late.
Could even be final.

packages/neon/neon/lib/src/blocs/user_statuses.dart
```diff
@@ -0,0 +1,87 @@
...
+      if (_account.username == username) {
+        final isAway =
+            _platform.canUseWindowManager && (!(await windowManager.isFocused()) || !(await windowManager.isVisible()));
+        data = (await _account.client.userStatus.heartbeat(
+          status: isAway ? NextcloudUserStatusType.away : NextcloudUserStatusType.online,
+        ))
+            .ocs
+            .data
+            .userStatusStatus
+            ?.userStatusPublicStatus;
```
Leptopoda, line 61: I'm not a fan of `(await somefuture).value` I'd rather store the result of await in a separate variable like:

```dart
final results = await _account.client.userStatus.heartbeat(
          status: isAway ? NextcloudUserStatusType.away : NextcloudUserStatusType.online,
        );
data = results.ocs
            .data
            .userStatusStatus
            ?.userStatusPublicStatus;
```

packages/neon/neon/lib/src/blocs/user_statuses.dart
```diff
@@ -0,0 +1,87 @@
...
+      } else {
+        data = (await _account.client.userStatus.getPublicStatus(userId: username)).ocs.data.userStatusPublicStatus;
```
Leptopoda, line 63: same here

Leptopoda (changes requested):
Didn't had a deep look at the bloc yet :sweat_smile:

packages/neon/neon/lib/src/widgets/user_avatar.dart
provokateurin: Ah yes now that it uses a Container

packages/neon/neon/lib/src/blocs/user_statuses.dart
provokateurin: Oops forgot to remove that

provokateurin:
Timers are now cancellable (let me know if you like they way I implemented it) and I added tests for the TimerBloc.

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return TimerBlocTimer(duration, callback);
+  }
+
+  @override
+  void unregisterTimer(final TimerBlocTimer timer) {
+    if (_timers[timer.duration.inSeconds] != null) {
+      _callbacks[timer.duration.inSeconds] =
+          _callbacks[timer.duration.inSeconds]!.where((final cb) => cb != timer.callback).toList();
+    }
```
Leptopoda, line 56: wouldn't `callbacks[timer.duration.inSeconds]?.remove(timer.callback);` also do the trick?

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,72 @@
...
+/// The [TimerBloc] is a singleton.
+/// Sub-second timers are not supported.
+class TimerBloc extends Bloc implements TimerBlocEvents, TimerBlocStates {
+  factory TimerBloc() => _singleton;
+  TimerBloc._();
+
+  static final TimerBloc _singleton = TimerBloc._();
+
+  final Map<int, Timer> _timers = {};
+  final Map<int, List<VoidCallback>> _callbacks = {};
```
Leptopoda, line 27: can we make it a `Set<VoidCallback>`?

I don't think it makes sense to add a callback twice.

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,72 @@
...
+
+  @override
+  void dispose() {
+    for (final timer in _timers.values) {
+      timer.cancel();
```
Leptopoda, line 32: While this will cancel all timers they won't be removed from `_timers` or `_callbacks`.
We might want to clear them aswell (or just reassign the TimerBloc._singleton with a new instance).

Please also add unit tests for that case so this wont happen again.

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,72 @@
...
+}
+
+abstract class TimerBlocStates {}
+
+/// Execute callbacks at defined periodic intervals.
+/// Components can register their callbacks and everything with the same periodicity will be executed at the same time.
+///
+/// The [TimerBloc] is a singleton.
+/// Sub-second timers are not supported.
+class TimerBloc extends Bloc implements TimerBlocEvents, TimerBlocStates {
```
Leptopoda, line 20: We've added unit tests for TimerBloc :tada:
But how do you imagine other classes implementing TimerBloc to mock this?

I'm not sure it is the best way but in the past I've done it like this:

```dart
class T {
factory T() => instance ??= T._();

@VisibleForTesting
factory T.mocked(T mock) => instance ??= mock;

T._();

@VisibleForTesting
static T? instance;
}
```

1. This let's me reset the singleton by setting it's instance to null (maybe related to dispose)
2. I can easily mock it
3. I like calling it instance rather than singleton as this helps people understand it even without knowing the singleton pattern

packages/neon/neon/test/timer_bloc_test.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:neon/neon.dart';
+import 'package:test/test.dart';
+
+void main() {
```
Leptopoda, line 4: we should reset the bloc between tests `tearDown(() => TimerBloc().dispose());`

packages/neon/neon/test/timer_bloc_test.dart
```diff
@@ -0,0 +1,24 @@
...
+
+      expect(stopwatch.elapsedMilliseconds, greaterThan(duration.inMilliseconds));
+      expect(stopwatch.elapsedMilliseconds, lessThan(duration.inMilliseconds * 1.1));
+    });
+
+    test('Unregister timer', () async {
+      const duration = Duration(milliseconds: 100);
+
+      TimerBloc().registerTimer(duration, neverCalled).cancel();
+      await Future.delayed(duration);
```
Leptopoda, line 21: Other than implicitly testing that it doesn't throw something we aren't really testing anything here:
I'd suggest:

```dart
    test('Unregister timer', () async {
      const duration = Duration(milliseconds: 100);
      final callback = neverCalled;

      TimerBloc().registerTimer(duration, callback).cancel();
      await Future.delayed(duration);

      expect(TimerBloc().callbacks[duration.inSeconds], isNot(contains(callback)));
    });
```

packages/neon/neon/test/timer_bloc_test.dart
```diff
@@ -0,0 +1,24 @@
...
+  group('TimerBloc', () {
+    test('Register timer', () async {
+      const duration = Duration(milliseconds: 100);
+
+      final stopwatch = Stopwatch()..start();
+      TimerBloc().registerTimer(duration, stopwatch.stop);
+      await Future.delayed(duration);
+
+      expect(stopwatch.elapsedMilliseconds, greaterThan(duration.inMilliseconds));
+      expect(stopwatch.elapsedMilliseconds, lessThan(duration.inMilliseconds * 1.1));
```
Leptopoda, line 14: We might want to verify that we actually store the data:

```dart
    test('Register timer', () async {
      const duration = Duration(milliseconds: 100);

      final stopwatch = Stopwatch()..start();
      final callback = stopwatch.stop;
      TimerBloc().registerTimer(duration, callback);
      await Future.delayed(duration);

      expect(stopwatch.elapsedMilliseconds, greaterThan(duration.inMilliseconds));
      expect(stopwatch.elapsedMilliseconds, lessThan(duration.inMilliseconds * 1.1));
      expect(TimerBloc().callbacks[duration.inSeconds], contains(callback));
      expect(TimerBloc().timers[duration.inSeconds], isNot(isNull));
    });
```

This requires the underlying `TimerBloc.timers` and `TimerBloc.callbacks` to not have an underscore and be `@VisibleForTesting`

packages/neon/neon/test/timer_bloc_test.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:neon/neon.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('TimerBloc', () {
```
Leptopoda, line 5: As mentioned above we should test that dispose stops all timers and clears it's underlying data.
Being a singleton we should verify this.

packages/neon/neon/lib/src/blocs/timer.dart
```diff
@@ -0,0 +1,72 @@
...
+  @override
+  void unregisterTimer(final TimerBlocTimer timer) {
+    if (_timers[timer.duration.inSeconds] != null) {
+      _callbacks[timer.duration.inSeconds] =
+          _callbacks[timer.duration.inSeconds]!.where((final cb) => cb != timer.callback).toList();
+    }
+  }
+}
+
+class TimerBlocTimer {
```
Leptopoda, line 60: Can we call this `NeonTimer`?

I don't like the current name.

packages/nextcloud/lib/src/nextcloud.openapi.json
```diff
@@ -1894,47 +1894,29 @@
...
-          },
-          "messageId": {
-            "type": "string"
-          },
-          "messageIsPredefined": {
-            "type": "boolean"
-          },
-          "icon": {
-            "type": "string"
+        "allOf": [
```
Leptopoda, line 1897: @Leptopoda I should finalize my really outdated patch that adds inheritance

Leptopoda (changes requested):

packages/neon/neon/lib/src/blocs/timer.dart
provokateurin: Hm you could have a callback with multiple intervals? Although that seems quite useless so we might just not support it

packages/neon/neon/lib/src/blocs/timer.dart
provokateurin: Since it's a singletone disposing doesn't make much sense, right?

packages/neon/neon/lib/src/blocs/timer.dart
provokateurin: Looks good. I wanted to have instances for tests so they don't interfere, but didn't think about this.

packages/nextcloud/lib/src/nextcloud.openapi.json
provokateurin: That would be really neat :) With the automatic OpenAPI specs it will get a lot more important because it appears much more often there.

packages/neon/neon/lib/src/blocs/timer.dart
Leptopoda: A callback with multiple intervals would place the callback in different sets.

I meant `Map<int, Set<VoidCallback>>` still having one set per intervall.

Leptopoda (approved):
Code looks very good now :+1:

PR #346, opened 2023-05-31
neon: Allow completely hiding linear progress indicator

provokateurin:
Changes from the feature/spreed branch.

### DIFF
--- a/packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
+++ b/packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
@@ -3,6 +3,7 @@ part of '../../neon.dart';
 class NeonLinearProgressIndicator extends StatelessWidget {
   const NeonLinearProgressIndicator({
     this.visible = true,
+    this.alwaysThere = true,
     this.margin = const EdgeInsets.symmetric(horizontal: 10),
     this.color,
     this.backgroundColor = Colors.transparent,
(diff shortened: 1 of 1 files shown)


packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
```diff
@@ -3,28 +3,32 @@ part of '../../neon.dart';
 class NeonLinearProgressIndicator extends StatelessWidget {
   const NeonLinearProgressIndicator({
     this.visible = true,
+    this.alwaysThere = true,
```
Leptopoda, line 6: I don't think adding more confusing options to the interface is good.

What exactly is the usecase? That apps can force hide the appBar?
In this case I'd again suggest making a theming option through a ThemeExtension so one could disable this for the entire subtree.

packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
```diff
@@ -3,28 +3,32 @@ part of '../../neon.dart';
...
-          child: visible
-              ? LinearProgressIndicator(
-                  color: color,
-                  backgroundColor: backgroundColor,
-                )
-              : null,
-        ),
-      );
+  Widget build(final BuildContext context) => !alwaysThere && !visible
+      ? Container()
```
Leptopoda, line 21: Instead of using a `Container()` as an empty widget a `SizedBox()` would be preferred as it has a constant constructor that could improve layout performance.

packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
```diff
@@ -3,28 +3,32 @@ part of '../../neon.dart';
...
+      : Container(
+          margin: margin,
+          child: SizedBox(
+            height: 3,
```
Leptopoda, line 25: the height can be specified in the container improving readability.

packages/neon/neon/lib/src/widgets/linear_progress_indicator.dart
Leptopoda, line 1: Can we also start moving this into it's own package?

Leptopoda (changes requested):

PR #347, opened 2023-06-01
neon: Fix image background antialiasing

provokateurin:
Only today I realized that a color filter is needed to fix this. Previously I was trying to do it using a Stack or the background color, but that didn't work out due to the antialiasing that is applied

Before:
image
After:
image

### DIFF
--- a/packages/neon/neon/lib/src/widgets/image_wrapper.dart
+++ b/packages/neon/neon/lib/src/widgets/image_wrapper.dart
@@ -15,14 +15,17 @@ class NeonImageWrapper extends StatelessWidget {
   final BorderRadius? borderRadius;
 
   @override
-  Widget build(final BuildContext context) => SizedBox.fromSize(
-        size: size,
-        child: ClipRRect(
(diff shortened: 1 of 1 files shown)


packages/neon/neon/lib/src/widgets/image_wrapper.dart
```diff
@@ -15,14 +15,17 @@ class NeonImageWrapper extends StatelessWidget {
...
-            child: Center(
-              child: child,
+  Widget build(final BuildContext context) => ClipRRect(
+        borderRadius: borderRadius ?? BorderRadius.zero,
+        child: ColorFiltered(
+          colorFilter: ColorFilter.mode(color, BlendMode.dstATop),
+          child: SizedBox.fromSize(
+            size: size,
+            child: ColoredBox(
+              color: Colors.transparent,
```
Leptopoda, line 25: Huh what happened here? :eyes: 

The ColoredBox could probably be removed

Leptopoda (changes requested):

Leptopoda (approved):

Issue #343, opened 2023-05-29
LoginFlow restructure


Leptopoda:
We currently have our `AccountsBloc` that is added to a provider and thus accessible throughout the app.
My idea is to let it also handle the loginflow by making it instantiate a `LoginBloc` and handle it's lifecycle.
This has the big advantage that we can access the LoginBloc (if instantiated) wherever we have an AccountsBloc.

My intention was to separate the `UrlEntryScreen` and the `WebViewScreen` and having access to the LoginBloc is crucial for this.
We could just pass it as a parameter or instantiate a new Provider making the LoginBloc only accessible for children of the `UrlEntryScreen`.

In #338 I wanted to make routing automatic so once a user is logged in the router would get notified and just redirect us to the homescreen (wich would show the homescreen for the currently active app and account).
Sadly once we wanted to login a second user the redirection logic intercepted our routing to the LoginScreen and again redirected us to the HomeScreen (if we knew that a new login was instantiated we could also decide based on that knowledge).

Lastly we could also move all of
https://github.com/provokateurin/nextcloud-neon/blob/d47461f4456279a231b6cbfde049c2110a288053/packages/neon/neon/lib/src/pages/login.dart#L60-L95

into the AccountsBloc as it is a perfect candidate of business logic that does not need to be in the Widget. As for the error case we could use the streams onError callback to handle the snackbars in the UI.

---

I'm not that sure if it does make sense to move all routing decisions to the go_router redirection feature as I'd need to see how that would turn out but I think that moving LoginBloc into the AccountsBloc would be a good cleanup of some UI code.

Finally we should also have #283 in mind as it's also LoginFlow related.

provokateurin:
Sounds about right, would you mind opening a PR for that?

https://github.com/provokateurin/nextcloud-neon/issues/284 should also be considered because it makes everything a little more complicated. For the login flow we first get the url and need to check the server connection before initializing the flow, but for qr code we get all the information at once and need to check the server connection.
My idea was to have a `ServerConnectionPage` that would show some loading spinners or checkmarks or exclamation marks based on the connection. The four categories for that would be status (is it reachable), server version (compatibility), maintenance mode (not really relevant for qr login, but login flow won't work) and the credentials (basically what https://github.com/provokateurin/nextcloud-neon/blob/d47461f4456279a231b6cbfde049c2110a288053/packages/neon/neon/lib/src/models/account.dart#L10 is doing, but showing an error when a 401 is returned).

[...]

PR #348, opened 2023-06-02
RFC: Login docs

provokateurin:
https://github.com/provokateurin/nextcloud-neon/issues/343


Leptopoda:
Looks good.
I'd like to make the validations (validate server details) and (validate login credentials) a condition. The validation can fail and what happens.
Or do you think this is irrelevant for this chart?

Does it make sense to name the end state? Like logged in?

---

a more general question:
I think once our doc comments are mature enough we probably want to host the generated docs either here on GH-Pages or by pub.dev once published. In this case we'd probably want to also publish this with it. I've never used it but dart doc supports topics. Arbitrary markdown that will be added to the generated docs. I only saw that in the [go_router docs](https://pub.dev/documentation/go_router/latest/) and think this could be good to use.

In that case we should start with a structure that'll support this flow (separate md per documentation part). 

---

How didn't I know plantuml before. I've only used umlet and it's just awefull. Would've made a lot easier.

docs/login.puml
```diff
@@ -0,0 +1,35 @@
...
+    endif
+endif
+
+if "Selected login method?" then
+-[#blue]->[Login flow] "Validate server details"
+-[#blue]-> "Open login flow page"
+-[#blue]-> "Initiate login flow"
+-[#blue]-> "Open web browser"
+-[#blue]-> "Wait for login flow result"
+-[#blue]-> "Validate login details"
```
provokateurin, line 19: This one is obsolete, as we can be sure that the credentials returned by the login flow are correct (nevertheless we need to fetch the username, so we might as well leave it in)

provokateurin: @Leptopoda Should we leave it or not? The logic would be exactly the same and we should also display some UI

provokateurin:
I tried adding the failure cases but the diagram looks horrible :/

provokateurin:
And the diagram gets way too complicated because it's not possible to properly reuse these nodes.

Leptopoda (approved):

PR #359, opened 2023-06-04
neon: Fix missing active account

provokateurin:


packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -37,19 +37,28 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
     });
 
     final as = accounts.value;
+
     if (_globalOptions.rememberLastUsedAccount.value && _storage.containsKey(_keyLastUsedAccount)) {
       final lastUsedAccountID = _storage.getString(_keyLastUsedAccount);
-      activeAccount.add(lastUsedAccountID != null ? as.find(lastUsedAccountID) : null);
```
provokateurin, line 42: For some reason this code failed because the key was there, but the value was null.

provokateurin: I assume I was wiped here when I clicked on the notification: https://github.com/provokateurin/nextcloud-neon/blob/35eb2d65799f3e6a5f53b62da8cac868d3e52f86/packages/neon/neon/lib/src/app.dart#L143
But I don't understand how the account could be null.

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -37,19 +37,28 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
...
-      );
+      if (lastUsedAccountID != null) {
+        final aa = as.find(lastUsedAccountID);
+        if (aa != null) {
+          activeAccount.add(aa);
+        }
+      }
     }
+
+    _globalOptions.initialAccount.stream.listen((final lastAccount) {
```
Leptopoda, line 51: Why do we need to listen to this?
we only rely on the executed logic on startup but are still listening to it the for entire life cycle of the app.

Leptopoda:
Could it be that 
https://github.com/provokateurin/nextcloud-neon/blob/35eb2d65799f3e6a5f53b62da8cac868d3e52f86/packages/neon/neon/lib/src/blocs/accounts.dart#L28

needs to be called?
Currently we load the accounts and then register the listener thus not executing it.

packages/neon/neon/lib/src/blocs/accounts.dart
provokateurin: I forgot to revert it to the `.stream.first.then`

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -37,19 +37,28 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
...
-          final account = lastAccount != null ? as.find(lastAccount) : null;
-          if (account != null) {
-            activeAccount.add(account);
-          }
-        }),
-      );
+      if (lastUsedAccountID != null) {
+        final aa = as.find(lastUsedAccountID);
+        if (aa != null) {
+          activeAccount.add(aa);
```
Leptopoda, line 46: Can we use setActiveAccount?

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -37,19 +37,28 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
...
+        }
+      }
     }
+
+    _globalOptions.initialAccount.stream.listen((final lastAccount) {
+      final account = lastAccount != null ? as.find(lastAccount) : null;
+      // As long as no active account has been set it should set the active account
+      if (activeAccount.valueOrNull == null) {
+        if (account != null) {
+          activeAccount.add(account);
```
Leptopoda, line 56: Can we use setActiveAccount?

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -37,19 +37,28 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
...
+        } else if (as.isNotEmpty) {
+          activeAccount.add(as.first);
```
Leptopoda, line 58: Can we use setActiveAccount?

provokateurin:
>  Currently we load the accounts and then register the listener thus not executing it.

I don't understand what you mean. In the constructor we listen to account changes and update the global options accordingly, so I don't see a problem there?

Leptopoda:
My bad. I thought that we are adding event's to the stream and then add a listener thus the listener would not be called as the events where already added (before we had a listener).

Behaviorsubject being a broadcast stream obviously doesn't like that and the listener will still be called.

Leptopoda (approved):

PR #360, opened 2023-06-04
Fix account selector avatar

provokateurin:


Leptopoda (changes requested):
Can't we just use a LayoutBuilder with a size constraint?

This could look something like:
[0001-patch.txt](https://github.com/provokateurin/nextcloud-neon/files/11654054/0001-patch.txt)

Yep it's based on the cleaner code as it was easier to work with it.
This should enable us to eliminate the use of `IntrinsicWidth` or `IntrinsicHeight`.

- This patch does not allow forcing the size to be too big but if we want to enable such (potentially stupid) stuff one could `size = widget.size ?? constraints.constrain(Size.square(kAvatarSize)).shortestSide;`.

Also treat that patch as an experiment if that would even work that I came up in a few minutes (it might be broken in other ways :sweat_smile:)

provokateurin:
I'll wait with this for https://github.com/provokateurin/nextcloud-neon/pull/344 getting merged. Rebasing would be a PITA

Leptopoda:
Now it should be fairly easy to use the patch.

packages/neon/neon/lib/src/widgets/user_avatar.dart
```diff
@@ -34,47 +35,50 @@ class _UserAvatarState extends State<NeonUserAvatar> {
...
-          ),
-        ],
-      ],
-    );
-  }
+  Widget build(final BuildContext context) => LayoutBuilder(
+        builder: (final context, final constraints) {
+          final isDark = Theme.of(context).brightness == Brightness.dark;
+          size = constraints.constrain(Size.square(widget.size)).shortestSide;
+          final pixelSize = (size * MediaQuery.of(context).devicePixelRatio).toInt();
```
Leptopoda, line 42: Why do we need this? can't we use size directly?

Leptopoda (approved):
Didn't test it but looks ok if the above is intentional.

packages/neon/neon/lib/src/widgets/user_avatar.dart
provokateurin: The flutter size is not the same size as physical pixels. To get an image with the correct resolution for the display we need to convert the flutter size into the physical pixel size.

provokateurin: >   /// The number of device pixels for each logical pixel. This number might not
  /// be a power of two. Indeed, it might not even be an integer. For example,
  /// the Nexus 6 has a device pixel ratio of 3.5.

Leptopoda: Ah yeah makes sense

PR #363, opened 2023-06-07
Feature/adaptive layout

Leptopoda:
We need to make the apps independent form other parameters so we can easily put it into a go_route.

Now every app will have it's own provider that neon will register for it.

### DIFF
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -144,4 +144,8 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
 
     return _blocs[appImplementation.id] = appImplementation.buildBloc(_account.client) as T;
   }
+
+  List<Provider>? getAppProviders() => appImplementations.valueOrNull?.data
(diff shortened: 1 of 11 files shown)


provokateurin:
Very nice changes! I wanted to do this injection for a while.
Do you think it makes sense to also inject the options for each of the apps? Right now they are also passed to the blocs and just available via `bloc.options`, so it doesn't add much, but it might be cleaner.

provokateurin (approved):

Leptopoda:
after the adaptive changes I'd like to make `AppImplementation` an interface so yes this makes sense long term.

provokateurin:
The notifications app is broken because the provider is not injected (it doesn't appear in the normal appImplementations list) but it also fails to find it due to the missing provider for the new route when the page is opened. The MultiProvider should probably wrap the whole MaterialApp to prevent this.

Leptopoda:
Ah yes. I didn't think about that.

I'll need to find a solution as the next step would be to use go router for routing and I think my assumptions there will also break the notifications app.

Back to the drawing board (I might get to it tomorrow evening.

While wrapping the entire MaterialApp will probably fix this I don't think it's the best solution going forward.

provokateurin:
Yeah something temporary would be nice to have it working again :)

PR #371, opened 2023-06-08
neon: fix blocProvider of notifications app

Leptopoda:
does this work?
I couldn't test it :(

It might be broken when opening the app from a notification.
If not we really need to wrap the entire app in the providers.

### DIFF
--- a/packages/neon/neon/lib/neon.dart
+++ b/packages/neon/neon/lib/neon.dart
@@ -193,12 +193,7 @@ Future runNeon({
           create: (final _) => branding,
         ),
       ],
-      child: NeonApp(
-        accountsBloc: accountsBloc,
-        sharedPreferences: sharedPreferences,
-        platform: platform,
-        globalOptions: globalOptions,
-      ),
+      child: const NeonApp(),
     ),
   );
 }
--- a/packages/neon/neon/lib/src/app.dart
+++ b/packages/neon/neon/lib/src/app.dart
@@ -2,17 +2,8 @@ part of '../../neon.dart';
 
 class NeonApp extends StatefulWidget {
   const NeonApp({
-    required this.accountsBloc,
-    required this.sharedPreferences,
-    required this.platform,
-    required this.globalOptions,
     super.key,
   });
-
-  final AccountsBloc accountsBloc;
-  final SharedPreferences sharedPreferences;
-  final NeonPlatform platform;
-  final GlobalOptions globalOptions;
 
   @override
   State<NeonApp> createState() => _NeonAppState();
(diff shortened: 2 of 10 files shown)


Leptopoda:
Maybe once I did the go_router changes we can let each app inject the provider in it's own route.

This way we are always sure that the subtree containing the app has access to the relevant provider.

provokateurin:
It would also be nicer for separate. Apps should be able to access the bloc of another app (at least directly without any interface in between).

Leptopoda:
In that case it would probably be best to register them before the material app as suggested.

provokateurin:
image

I think this is because of the interface?

provokateurin:
This patch makes it work:
[patch.txt](https://github.com/provokateurin/nextcloud-neon/files/11710695/patch.txt)

It was also trying to put a provider inside a provider instead of the bloc.

provokateurin:
I haven't tested opening the app from a notification, but it should work as well, because it takes the same code path.

Leptopoda:
The issue actually is that the notifications AppInterface does not specify the generics

https://github.com/provokateurin/nextcloud-neon/blob/6b60409912de0599237aa0caf2042732b91fdbec/packages/neon/neon_notifications/lib/neon_notifications.dart#L17

But we rely on it for the provider
https://github.com/provokateurin/nextcloud-neon/blob/6b60409912de0599237aa0caf2042732b91fdbec/packages/neon/neon/lib/src/utils/app_implementation.dart#L27-L29

---

I also noticed another few issues like not caching the bloc as we did before
https://github.com/provokateurin/nextcloud-neon/blob/6b60409912de0599237aa0caf2042732b91fdbec/packages/neon/neon/lib/src/blocs/apps.dart#L140-L151

leading to potentially having multiple blocs for the same app (or even memory leaks as we don't properly dispose them rn)

Also getAppsBloc is also still there again leading to multiple having instances of the same bloc.
I've fixed some parts of this already and hope to update the PR today.

---

I've also switched to injecting the providers before the material app so we can access blocs of other apps throughout the app (I initially thought this was stupid but limitations in the cookbook api actually require us uploading some files so we definitely need the filesBloc there)

provokateurin:
LGTM, but I want to test it first. You could squash https://github.com/provokateurin/nextcloud-neon/pull/371/commits/391ae1c96bb30eba050ef070638ad32ab0b474c3 and https://github.com/provokateurin/nextcloud-neon/pull/371/commits/99309a9279bca2d98f741dc62081499bee247f23.

provokateurin (approved):

PR #374, opened 2023-06-13
Refactor/iterable

Leptopoda:
as we don't need the interface being a list we should relax the type a bit.


packages/neon/neon/lib/neon.dart
provokateurin, line 1: Did you intend to commit the changes in this and the other files with the binding? They don't look related to your iterable cleanup.

Leptopoda:
Yeah I can make a separate PR.

I started with the iterable and thought why not clean up the `runNeon` interface even more.
I could also just rename the PR to cleanup `runNeon` interface.

provokateurin:
You can leave it, I had a weird diff view, but the changes make sense.

PR #376, opened 2023-06-13
neon_news, neon_notes: use records in dialogs

Leptopoda:
Improves typesafety and readability

### DIFF
--- a/packages/neon/neon_notes/lib/widgets/notes_floating_action_button.dart
+++ b/packages/neon/neon_notes/lib/widgets/notes_floating_action_button.dart
@@ -13,7 +13,7 @@ class NotesFloatingActionButton extends StatelessWidget {
   @override
   Widget build(final BuildContext context) => FloatingActionButton(
         onPressed: () async {
-          final result = await showDialog<List>(
+          final result = await showDialog<(String, String?)>(
             context: context,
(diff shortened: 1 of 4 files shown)


packages/neon/neon_notes/lib/widgets/notes_floating_action_button.dart
```diff
@@ -13,17 +13,18 @@ class NotesFloatingActionButton extends StatelessWidget {
   @override
   Widget build(final BuildContext context) => FloatingActionButton(
         onPressed: () async {
-          final result = await showDialog<List>(
+          final result = await showDialog<(String, String?)>(
```
provokateurin, line 16: Is it possible to do something like `final (title, category)? = await showDialog<(String, String?)>(`? I really don't know, but it might make it a little cleaner

Leptopoda: I tried to but because of the nullability it's not working.
I mean how would the if check look like?

The only thing we could do it use `result.$1` and `result.$2` instead of assigning them into different variables (We only use them once)

provokateurin: I think assigning to variables makes the code easier to understand, so better leave it as is

PR #377, opened 2023-06-13
Refactor/blocs

Leptopoda:
Further testing is needed for this as this is quite big.


Leptopoda:
And it's completely broken :)
Didn't think of the navigator killing our providers. I'll leave it as is and re think of this once go_router is handling our apps.

PR #381, opened 2023-06-13
Move logo to branding

provokateurin:
From https://github.com/provokateurin/nextcloud-neon/pull/372


Leptopoda (approved):
We should refactor the code to make it more readable but it's good for now

PR #386, opened 2023-06-13
docs: Add TOC

provokateurin:


Leptopoda (approved):
We could also rename `docs.md` to `README.md` displaying it automatacially

provokateurin:
Ah good idea :D Sometimes the solution can be so simple

PR #387, opened 2023-06-14
Feature/adaptive layout

Leptopoda:
I haven't profiled it but it should improve the performance (on startup and when switching apps or accounts) quite a bit.

It removes the quickbar drawer option and I don't plan to reimplement it after the adaptive layout changes thus deprecating the entire drawer settings (We can still discuss this).

This change also removes the loading bar in the drawer but I don't think we need it and should actually start to remove a few of them.


Leptopoda:
Wtf is the ci trying to do :thinking:

packages/neon/neon/lib/src/utils/global_options.dart
```diff
@@ -232,5 +233,6 @@ class GlobalOptions {
 enum NavigationMode {
   drawer,
   drawerAlwaysVisible,
+  @Deprecated("The new design won't use this anymore")
```
provokateurin, line 236: You can just remove it completely. That's the benefit of not having and promised stability ;)

provokateurin:
I think you can easily squash https://github.com/provokateurin/nextcloud-neon/pull/387/commits/313194ebf621fb4519cd8a619fabc4f9c5fd90e1 and https://github.com/provokateurin/nextcloud-neon/pull/387/commits/eb19942092fcd91b1597b77a07bd33a2b14fa78b

packages/neon/neon/lib/src/widgets/drawer_destination.dart
```diff
@@ -0,0 +1,137 @@
...
+
+    return NavigationDrawerDestination(
+      label: labelWidget,
+      icon: neonDestination.icon(),
+      selectedIcon: neonDestination.selectedIcon,
+    );
+  }
+}
+
+extension TabnExtension on Tab {
```
provokateurin, line 132: The extension name looks like a typo?

packages/neon/neon/lib/src/widgets/drawer.dart
```diff
@@ -0,0 +1,208 @@
...
+    if (index < _apps.length) {
+      unawaited(_appsBloc.setActiveApp(_apps[index].id));
+      //context.goNamed(apps[index].routeName);
+
+      setState(() {
+        _activeApp = index;
+      });
+    } else {
+      const SettingsRoute().go(context);
+    }
```
provokateurin, line 88: This logic isn't that easy to understand. Could you add a comment to explain why this check exists? It took me few moments to understand what is going on.

provokateurin (changes requested):
Changes look very good. Will need to test it myself though before giving it the green light

packages/neon/neon/lib/src/utils/global_options.dart
Leptopoda: Not yet.
If someone (like my phone) has the navigation set to it it would fail when loading from storage.
I'll remove the entire setting in a followup.

packages/neon/neon/lib/src/widgets/drawer_destination.dart
Leptopoda: We probably don't even need the Tap. it's a leftover from testing on my other branch but I quite like having it

provokateurin:
Should I review again? You can just click the button again to let me know :)

Leptopoda:
Yeah I just pushed it and left home. Totally forgot that :)

packages/neon/neon/lib/src/widgets/drawer.dart
```diff
@@ -0,0 +1,210 @@
...
+  @override
+  void dispose() {
+    _tabController.dispose();
+    super.dispose();
+  }
+
+  void onAppChange(final int index) {
+    setState(() {
+      _activeApp = index;
+    });
```
provokateurin, line 81: I already expected this to break when I saw this change. When you go to the settings item it correctly opens the settings page, but afterwards the settings item is selected. I guess you can just move it to the end of the method to fix this.

provokateurin (changes requested):

provokateurin:
Looks very nice and works well! Can't wait for the new layout :)

packages/neon/neon/lib/src/widgets/drawer.dart
Leptopoda: I only tested it with the normal drawer and totally forgot the persistent one.

Leptopoda:
> Looks very nice and works well! Can't wait for the new layout :)

It'll probably be delayed as I need to do further restructures/cleanups in the home.dart to keep the transition small.

provokateurin (approved):

PR #388, opened 2023-06-14
Deps/update

Leptopoda:


packages/app/linux/flutter/generated_plugins.cmake
```diff
@@ -3,6 +3,7 @@
 #
 
 list(APPEND FLUTTER_PLUGIN_LIST
+  file_selector_linux
```
provokateurin, line 6: This is really confusing for me, because I was these lines getting added when I built the app using the docker container, but when I build it locally without the container the lines are removed again. I must have some weird caching going on or something like that.

packages/app/pubspec.yaml
```diff
@@ -38,7 +38,7 @@ dev_dependencies:
     sdk: flutter
   integration_test:
     sdk: flutter
-  material_design_icons_flutter: ^6.0.7096
+  material_design_icons_flutter: ^7.0.7296
```
provokateurin, line 41: Nice, this also resolves my tree shaking problem for web

provokateurin:
Should we add dependabot or renovate?

Leptopoda:
I only know the former; Whats the benefit of renovate?
I don't know how it would do with melos or our strict flutter version pinning but it generally sounds good.

I only made the PR as I was annoyed by flutter always generating the plugins thing and I had to watch for it not being accidentally staged/commited.

packages/app/linux/flutter/generated_plugins.cmake
Leptopoda: That's odd. Did you resolve this?

provokateurin:
Renovate is not GH/MS and it is a lot more configurable iirc, but both are good options I guess

PR #389, opened 2023-06-15
Refactor/option

Leptopoda:
The option builder was adding two StreamBuilders to the widget tree where one would be enough.

This also makes Option.defaultvalue an actual value instead of a Stream.


Leptopoda:
So it's safe to merge or should we keep it if we need it in the future?
I don't think the initial value needs to be adaptive (I mean that's the reason of the initial value isn't it?)

PR #390, opened 2023-06-15
RFC: Feature/error bloc

Leptopoda:
I need some feedback for the second commit:

### Problem:
I want to get rid of
https://github.com/provokateurin/nextcloud-neon/blob/bba5c2b5757c2897f9810aded8cfc41738589ce4/packages/neon/neon/lib/src/pages/home.dart#L45-L84

As it's not really UI related and should be part of the blocs. On the other hand we are adding new listeners whenever the capabilities change and don't dispose the old ones wich could hunt us in the future.
Finally we search for the first unsupported app and then immediately show an error wich could be bad Ux when multiple apps are outdated a user might inform their admin about the needed changes and only afterwards find out that another app also needs updating (yeah bad admin if they don't just update everything when being in there but it's possible).

### Idea:
As you can see I moved the logic into the appsBloc and my Idea is to just communicate back to the UI if we find a mismatch. 
I thought of implementing an ErrorBloc where global errors or app specific errors can be added. This would allow us to have a central place to display and handle errors (no more 5 error icons similar to our multiple error bars).

### Problems with the Proposal:
we'd somehow need access to the l10n form within our ErrorBloc (see the commit) and I don't like this :(
We'd need to keep the l10n reference in the bloc updated when the language changes...

What do you think?

---

The first commit can be disregarded just some experiments for now.

### DIFF
--- a/packages/neon/neon/lib/neon.dart
+++ b/packages/neon/neon/lib/neon.dart
@@ -24,6 +24,7 @@ import 'package:neon/l10n/localizations.dart';
 import 'package:neon/src/blocs/blocs.dart';
 import 'package:neon/src/models/account.dart';
 import 'package:neon/src/models/push_notification.dart';
+import 'package:neon/src/pages/no_apps.dart';
 import 'package:neon/src/router.dart';
 import 'package:neon/src/widgets/drawer.dart';
 import 'package:neon/src/widgets/drawer_destination.dart';
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -50,6 +50,8 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
             }
           }),
         );
+
+        unawaited(_checkCompatibility());
       }
     });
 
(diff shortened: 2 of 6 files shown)


provokateurin:
I once hat such a global ErrorBloc, but it was not really nice. Errors should be shown where they happen. But I agree that 5 error icons or dialogs or snackbars is a bit much. It does make sense to have some global error handling though, so idk what the best way is

Leptopoda:
regarding the app version we can just add a separate stream to the appsBloc that holds the info and the UI can query it in the same place it did before. Just that the fetching is happening in the bloc.

---

We could also adapt the `FirstLaunchBloc` to be a `StartupBloc` and handle both the initial loading and errors. Future errors would be handled where they happen but all the startup checks can happen in the beginning.

---

My idea is to get rid of the `AppsBloc` in the `home.dart`. With the apps transitioning to go_router the activeAppID would be handled by the routes (in `/apps/:appid`).

With the appBar becoming its own widget (similar to the drawer) the HomePage widget would only be some scaffolding for our adaptive layouts.

Then the only state in the home page would be the appImplementations (mostly needed for loading and error handling) so also getting rid of this makes sense.

provokateurin:
> regarding the app version we can just add a separate stream to the appsBloc that holds the info and the UI can query it in the same place it did before. Just that the fetching is happening in the bloc.

Yep sounds like a good improvement

> We could also adapt the FirstLaunchBloc to be a StartupBloc and handle both the initial loading and errors. Future errors would be handled where they happen but all the startup checks can happen in the beginning.

We could also just keep the FirstLaunchBloc and just add the StartupBloc? OTOH it's they both do similar things and can just be one bloc :woman_shrugging:

Issue #391, opened 2023-06-16
Support for Nextcloud Server v27


AdmerStroh:
Hello,
I want to kindly ask, if it would be possible to support Nextcloud 27. 
https://github.com/adil192/saber/issues/717

Keep all the good work up and thank you in advance :)

ceskyDJ:
Are there any Nextcloud-specific problems with supporting multiple versions? I think it should be a deal breaker for apps, when their authors see this problem. It'd cause problems for many users, because they often can't upgrade Nextcloud to the new version (for commercial instances) or can't do it for some reason (e.g. incompatibility of important apps). So, I think, this should be resolved in some way. Maybe you can just switch communication objects (I don't know your code, but if you designed using DI, it shouldn't be a big problem) for now and come up with a better solution in the future. What do you think?

provokateurin:
The problem with supporting multiple versions is because of breaking API changes. We have some plans on how to deal with this but there will also be some changes in the future on the Nextcloud side that will make APIs much more stable

ceskyDJ:
I see, thanks for your explanation. So, what about my proposal? Is it hard to implement due to the design of your library? Do you know something more about the "stabilizing" of the Nextcloud API (especially when it is planned to be released)?

provokateurin:
I also work at Nextcloud and will be responsible for creating a versioning/compatibility strategy. I haven't written down anything concrete yet and it will still be some time until I pickup that task, but we aim to provide API stability for about 2-3 years before introducing breaking changes.
If you have anything special in mind please let me know, I will consider it :)

ceskyDJ:
Thanks for providing more information. I did a quick code lookup and finally understood the problem (at least I believe it). You depend on other libraries with break changes, so you can't do anything with this problem. I thought you set up your own wrapper of the Nextcloud API (so WebDAV, CalDAV, CardDAV, etc.), so you would be able to have more wrappers' implementations and just switch between them by the Nextcloud server version.

provokateurin:
No, we don't depend on other libraries for the Nextcloud API. All the code for that is in this repository. As I said, we have some plans to support multiple versions in the future (probably starting with 27 or 28)

PR #396, opened 2023-06-17
Support Nextcloud 27

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/391

The news app sadly hasn't been updated for/is broken on 27. It only affects the web UI which we don't care about (See https://github.com/nextcloud/news/issues/2242). @Leptopoda should we merge regardless or wait for the new app to get fixed? The linked issue sounded a bit like resolving the problem could take some time...

### DIFF
--- a/packages/nextcloud/test/webdav.dart
+++ b/packages/nextcloud/test/webdav.dart
@@ -47,7 +47,7 @@ Future run(final DockerImage image) async {
         ),
       ))
           .responses;
-      expect(responses, hasLength(9));
+      expect(responses, hasLength(10));
       final props =
           responses.singleWhere((final response) => response.href!.endsWith('/Nextcloud.png')).propstats.first.prop;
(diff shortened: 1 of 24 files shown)


packages/nextcloud/test/webdav.dart
Leptopoda, line 1: Does the default NC now include more files or why is this change required

Leptopoda (approved):
LGTM;

I've been using neon with 27 for a few days now and it works :)

Leptopoda:
If the news app works for us we should go forward with the update.

It's at the users discretion to update to 27 with incomplete news web support (I for example don't use news anymore and thus updated to 27 fairly quickly).

packages/nextcloud/test/webdav.dart
provokateurin: Yep

PR #399, opened 2023-06-18
Implement unified search

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/95

Not useful right now because we can't handle the search results (clicking on the result and executing the link). Needs more go router changes. The UI is also just there to get it working, it's not polished or anything.


Leptopoda (changes requested):
Didn't test it but some initial code feedback

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -235,6 +235,18 @@ class _HomePageState extends State<HomePage> {
                         ],
                       ),
                       actions: [
+                        if (appImplementations.data != null) ...[
```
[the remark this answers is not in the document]
provokateurin: Ah I forgot about that (and will probably in future PRs :sweat_smile:)

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,108 @@
...
+    this._requestManager,
+    this._client,
+    this._appImplementations,
+  );
+
+  final RequestManager _requestManager;
+  final NextcloudClient _client;
+  final Iterable<AppImplementation> _appImplementations;
+  late final _implementedProviders =
+      _appImplementations.map((final app) => app.unifiedSearchProviders).reduce((final a, final b) => [...a, ...b]);
```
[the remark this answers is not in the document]
provokateurin: I didn't know about expand.
I think having the same provider twice should actually be an error.

packages/neon/neon/lib/src/pages/search.dart
```diff
@@ -0,0 +1,88 @@
...
+                        style: Theme.of(context).textTheme.headlineSmall,
+                      ),
+                      NeonException(
+                        result.error,
+                        onRetry: bloc.refresh,
+                      ),
+                      NeonLinearProgressIndicator(
+                        visible: result.loading,
+                      ),
+                      for (final entry in result.data?.entries ?? <NextcloudCoreUnifiedSearchResultEntry>[]) ...[
```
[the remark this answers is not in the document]
provokateurin: The check above was actually wrong. The card should be displayed even if there is no data yet (for loading and error states).

packages/neon/neon/lib/src/pages/search.dart
```diff
@@ -0,0 +1,88 @@
...
+          title: Text(AppLocalizations.of(context).search),
+        ),
+        body: ResultBuilder<Iterable<NextcloudCoreUnifiedSearchProvider>>(
+          stream: bloc.providers,
+          builder: (final context, final providers) => NeonListView(
+            items: providers.data,
+            isLoading: providers.loading,
+            error: providers.error,
+            onRefresh: bloc.refresh,
+            builder: _buildProvider,
```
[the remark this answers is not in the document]
provokateurin: Yep, that would be really good. In this PR I noticed again how not great the NeonListView is right now (not even performance wise but just how you need to use it).

packages/neon/neon/lib/src/pages/search.dart
```diff
@@ -0,0 +1,88 @@
...
+                        provider.name,
+                        style: Theme.of(context).textTheme.headlineSmall,
+                      ),
+                      NeonException(
+                        result.error,
+                        onRetry: bloc.refresh,
+                      ),
+                      NeonLinearProgressIndicator(
+                        visible: result.loading,
+                      ),
```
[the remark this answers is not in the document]
provokateurin: Yes, but only for the providers. This is for the search results

packages/neon/neon/lib/src/pages/search.dart
```diff
@@ -0,0 +1,88 @@
...
+
+    bloc = Provider.of<AccountsBloc>(context, listen: false).activeAppsBloc.unifiedSearchBloc;
+  }
+
+  @override
+  Widget build(final BuildContext context) => Scaffold(
+        appBar: AppBar(
+          title: Text(AppLocalizations.of(context).search),
+        ),
+        body: ResultBuilder<Iterable<NextcloudCoreUnifiedSearchProvider>>(
```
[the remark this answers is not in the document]
provokateurin: Won't work because of the mapping

packages/neon/neon/lib/src/pages/search.dart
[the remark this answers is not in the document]
provokateurin: I wanted to add it now, but it doesn't make much sense because of the sorting that is going on. There is only on occurrences where we directly pass the result to the list view.

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,109 @@
...
+    this._requestManager,
+    this._client,
+    final Iterable<AppImplementation>? appImplementations,
+  ) : _appImplementations = appImplementations ?? [];
+
+  final RequestManager _requestManager;
+  final NextcloudClient _client;
+  final Iterable<AppImplementation> _appImplementations;
+  late final _implementedProviders =
+      _appImplementations.expand((final app) => app.unifiedSearchProviders?.toList() ?? []).toSet();
```
Leptopoda, line 24:
```dart
      _appImplementations.expand((final app) => app.unifiedSearchProviders ?? {}).toSet();
```
should also work

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,109 @@
...
+
+  BehaviorSubject<Map<NextcloudCoreUnifiedSearchProvider, Result<NextcloudCoreUnifiedSearchResult>>> get results;
+}
+
+class UnifiedSearchBloc extends InteractiveBloc implements UnifiedSearchBlocEvents, UnifiedSearchBlocStates {
+  UnifiedSearchBloc(
+    this._requestManager,
+    this._client,
+    final Iterable<AppImplementation>? appImplementations,
+  ) : _appImplementations = appImplementations ?? [];
```
Leptopoda, line 18: I don't see a reason why _appImplementations can't be null?

packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -192,4 +192,10 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
 
   List<Provider> get appBlocProviders =>
       _allAppImplementations.map((final appImplementation) => appImplementation.blocProvider).toList();
+
+  UnifiedSearchBloc get unifiedSearchBloc => UnifiedSearchBloc(
+        _requestManager,
+        _account.client,
+        appImplementations.value.data,
+      );
```
Leptopoda, line 200: I don't think SearchBloc where _appImplementations is empty or null make sense so we should probably jsut make the getter nullable.

Also `appImplementations.value` could be risky. I think we would need to use `appImplementations.valueOrNull` but I didn't 100% check that

Leptopoda (changes requested):

packages/neon/neon/lib/src/blocs/unified_search.dart
provokateurin: No, it expects a list to be returned

packages/neon/neon/lib/src/blocs/apps.dart
provokateurin: You are right, but the button only appears if `appImplementations` has a value. Still not that great...

packages/neon/neon/lib/src/blocs/unified_search.dart
Leptopoda: Just checked and it does work.
Obviously the part after `??` must be changed to match the type of unifiedSearchProviders (a Set)

We would save a couple iterations this way

provokateurin: Huh maybe I did something wrong. Will fix it in the next iteration

packages/neon/neon/lib/src/blocs/apps.dart
Leptopoda: I just looked at it and we already check so we could indeed make it nullable or add an assertion or something.
```dart
if (appImplementations.hasData) ...[
]
```

Leptopoda:
@provokateurin ready for review? or are you still working on it?

provokateurin:
No, I'm just rebasing it from time to time so that it doesn't get too out of sync (same for the other PRs)

provokateurin:
Depends on https://github.com/nextcloud/neon/pull/589

packages/neon/neon/test/account_test.dart
```diff
@@ -22,4 +22,42 @@ void main() {
...
   });
+
+  group('URI', () {
+    const testURL = 'apps/test?123=456#789';
+
+    for (final (serverURL, path) in {
+      'http://localhost': '',
+      'http://localhost:443': '',
+      'http://localhost:443/nextcloud': '/nextcloud',
+    }.entries.map((final entry) => (entry.key, entry.value))) {
```
Leptopoda, line 33: Can you just use a list with records?

```dart
for (final (serverURL, path) in [
      ('http://localhost', ''),
      ('http://localhost:443', ''),
      ('http://localhost:443/nextcloud', '/nextcloud'),]) {
```

should also work and is way more readable

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -0,0 +1,142 @@
...
+        builder: (final context, final snapshot) => AnimatedSize(
+          duration: const Duration(milliseconds: 100),
+          child: _buildProvider(
+            context,
+            accountsBloc,
+            bloc,
+            snapshot.key,
+            snapshot.value,
+          ),
+        ),
```
Leptopoda, line 46: This is to animate in the elements?

I'd say AnimatedListView please but let's keep it as a followup for #585

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  Widget? _buildFallbackIcon(
+    final BuildContext context,
+    final Account account,
+    final CoreUnifiedSearchResultEntry entry,
+  ) {
+    if (entry.icon.startsWith('/')) {
+      return NeonCachedImage.url(
+        size: const Size.square(kAvatarSize * 2 / 3),
```
Leptopoda, line 126: externalize this into a variable as it's used by both icons

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -0,0 +1,142 @@
...
+    );
+  }
+
+  Widget _buildIcon(final BuildContext context, final Account account, final CoreUnifiedSearchResultEntry entry) {
+    if (entry.thumbnailUrl.isNotEmpty) {
+      return NeonCachedImage.url(
+        size: const Size.square(kAvatarSize),
+        url: entry.thumbnailUrl,
+        account: account,
+        // The thumbnail URL might be set but a 404 is returned because there is not preview available
```
Leptopoda, line 111:
```suggestion
        // The thumbnail URL might be set but a 404 is returned because there is no preview available
```

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -77,6 +77,18 @@ class Account implements Credentials {
...
+  Uri completeUri(final Uri uri) {
+    final result = Uri.parse(serverURL).replace(
+      queryParameters: uri.queryParameters,
+      fragment: uri.fragment,
+    );
+    return result.replace(
+      path: uri.hasAbsolutePath ? uri.path : '${result.path}/${uri.path}',
+    );
+  }
+
```
Leptopoda, line 90: I think UriBuilder from https://pub.dev/packages/uri could clean this up a bit.

Please also add doc comments as after 5 minutes I still don't understand what this is doing

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,161 @@
...
+
+    try {
+      results.add(results.value.asLoading());
+      final providers = (await _account.client.core.unifiedSearch.getProviders()).ocs.data;
+      results.add(
+        Result.success({
+          for (final provider in providers) ...{
+            provider: Result.loading(),
+          },
+        }),
```
Leptopoda, line 95: Please use a generator function here

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,161 @@
...
+      ...Map.fromEntries(
+        results.entries.where((final entry) => _providerMatchesApp(entry.key, activeApp)),
+      ),
+      ...Map.fromEntries(
+        results.entries
+            .whereNot((final entry) => _providerMatchesApp(entry.key, activeApp))
+            .where((final entry) => _hasEntries(entry.value))
+            .sorted((final a, final b) => _sortEntriesCount(a.value, b.value)),
+      ),
+    };
```
Leptopoda, line 150: A generator function could also make sense here.

packages/neon/neon/lib/src/widgets/cached_image.dart
```diff
@@ -160,8 +169,21 @@ class _NeonCachedImageState extends State<NeonCachedImage> {
...
+  Widget _buildError(final dynamic error) =>
+      widget.errorBuilder?.call(context, error) ??
+      NeonException(
+        error,
+        onRetry: () {
+          setState(() {});
+        },
+        onlyIcon: true,
+        iconSize: widget.size?.shortestSide,
+        color: widget.iconColor ?? Theme.of(context).colorScheme.error,
```
Leptopoda, line 187: I think the error should always be displayed in the error color. Otherwise it might not get noticed.

packages/neon/neon/lib/src/widgets/cached_image.dart
```diff
@@ -61,8 +66,20 @@ class NeonCachedImage extends StatefulWidget {
   final Color? svgColor;
   final Color? iconColor;
 
-  static Future<Uint8List> _getImageFromUrl(final String url) async {
-    final file = await _cacheManager.getSingleFile(url);
+  final Widget? Function(BuildContext, dynamic)? errorBuilder;
```
Leptopoda, line 69: Separate typedef; perhaps `ErrorWidgetBuilder`.

This could also be in the NeonException as other widgets might also need to use it.

Leptopoda (changes requested):
Some initial requests.

packages/neon/neon/test/account_test.dart
provokateurin: Ah true :woman_facepalming:

packages/neon/neon/lib/src/models/account.dart
provokateurin: I was using that, but then found it unnecessary as it was possible with just Uri alone. Will add a doc comment.

packages/neon/neon/lib/src/widgets/unified_search_results.dart
Leptopoda: I think we can leave it as is.

The NeonListView change will also require some other smaller resteuctures here that don't make sense to have now.

I don't think all NeonListViews need to be animated but in the PR you can see a separate `NeonListView.custom` constructor. 

Passing an AnimatedSliverList (or however it's called) into that should do the trick.

packages/neon/neon/lib/src/blocs/unified_search.dart
Leptopoda: Make a `sync*` annotated function that returns `Iterable<MapEntry> and `yield` the values you want to add.

Then you can call the mentioned Map.fromEntries.

This should require creating less maps and "changing" the pointer of the enties.

Maybe https://www.geeksforgeeks.org/generators-in-dart/ is helpful.

provokateurin: Ah, I only knew async* generators, this makes sense :)

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  void _updateResults(
+    final CoreUnifiedSearchProvider provider,
+    final Result<CoreUnifiedSearchResult> result,
+  ) =>
+      results.add(
+        Result.success(
+          _sortResults({
+            ...results.value.data ?? {},
```
Leptopoda, line 133:
```suggestion
            ...?results.value.data,
```

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,167 @@
...
+            .sorted((final a, final b) => _sortEntriesCount(a.value, b.value)),
+      ),
+      ...Map.fromEntries(
+        results.entries
+            .whereNot((final entry) => _providerMatchesApp(entry.key, activeApp))
+            .where((final entry) => _hasEntries(entry.value))
+            .sorted((final a, final b) => _sortEntriesCount(a.value, b.value)),
+      ),
+    };
+  }
```
Leptopoda, line 157:
```suggestion
Iterable<MapEntry<CoreUnifiedSearchProvider, Result<CoreUnifiedSearchResult>>> _sortResults(
  final Map<CoreUnifiedSearchProvider, Result<CoreUnifiedSearchResult>> results,
) sync* {
  final activeApp = _appsBloc.activeApp.value;

  yield* results.entries
      .where((final entry) => _providerMatchesApp(entry.key, activeApp))
      .sorted((final a, final b) => _sortEntriesCount(a.value, b.value));
  yield* results.entries
      .whereNot((final entry) => _providerMatchesApp(entry.key, activeApp))
      .where((final entry) => _hasEntries(entry.value))
      .sorted((final a, final b) => _sortEntriesCount(a.value, b.value));
}
```

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  Widget _buildThumbnail(final BuildContext context, final Account account, final CoreUnifiedSearchResultEntry entry) {
+    if (entry.thumbnailUrl.isNotEmpty) {
+      return NeonCachedImage.url(
+        size: const Size.square(kAvatarSize),
+        url: entry.thumbnailUrl,
+        account: account,
+        // The thumbnail URL might be set but a 404 is returned because there is no preview available
+        errorBuilder: (final context, final error) => _buildFallbackIcon(context, account, entry),
```
Leptopoda, line 112:
```suggestion
        errorBuilder: (final context, final _) => _buildFallbackIcon(context, account, entry),
```

Leptopoda (changes requested):
Some final touchups

packages/neon/neon/lib/src/blocs/unified_search.dart
```diff
@@ -0,0 +1,160 @@
...
+    _appsBloc.activeApp.listen((final _) {
+      if (enabled.value) {
+        disable();
+      }
+    });
+  }
+
+  final AppsBloc _appsBloc;
+  final Account _account;
+  String _term = '';
```
Leptopoda, line 42: I think nullable makes sense here

Leptopoda (approved):
other than that :+1:

packages/neon/neon/lib/src/blocs/unified_search.dart
provokateurin: I used empty string because that is also how the search is reset. If there was null but also empty string then there would be two "empty" possibilities.

PR #402, opened 2023-06-19
neon: refactor page loading

Leptopoda:


packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -106,39 +106,26 @@ class _HomePageState extends State<HomePage> {
...
-              body = MultiProvider(
+              Widget body = MultiProvider(
                 providers: _appsBloc.appBlocProviders,
                 child: Scaffold(
                   key: _scaffoldKey,
                   resizeToAvoidBottomInset: false,
                   drawer: !drawerAlwaysVisible ? drawer : null,
                   appBar: appBar,
-                  body: body,
+                  body: appView ?? noAppsView,
```
provokateurin, line 128: I would rather use the null fall through operator to construct the no apps view here. Otherwise you are checking two conditions and they are not even the same (which could lead to problems for edge-cases).

provokateurin (changes requested):

packages/neon/neon/lib/src/pages/home.dart
Leptopoda: Do you mean like:
```dart
Scaffold(
  body: appView ?? Center(child: Text('no apps')),
);
```
?

This does not work because when the appimplementations haven't finished loading yet they are null wich would result in the 'no apps found error' being shown while loading the app.


I think I`ll revert it to the builder and return a empty SizedBox when loading. At least removing all the collum stuff:

Something like:
```dart
final body = Builder(builder: (context){
  if(appimplementations == null){
    return const SizedBox();
  }
  
  if(appimplementations.isEmpty){
    return // error page;
  }
  
  return appimplementations.currentPage;
});
```

provokateurin: Yes, but that would only be shown on the very first load, as the cached data would be shown otherwise.

Leptopoda: Yeah but I'd rather show an empty container (or SizedBox) than the error message.

Also depending on timing the error message would only be shown for a very brief time wich wouldn't be good for a11y

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -107,28 +107,22 @@ class _HomePageState extends State<HomePage> {
...
-                          Expanded(
-                            child: appImplementations.data!.find(activeAppIDSnapshot.data!)!.page,
-                          ),
-                        ],
-                      ],
-                    ],
-                  ],
-                ),
+                builder: (final context) {
+                  if (!appImplementations.hasData || !activeAppIDSnapshot.hasData) {
```
Leptopoda, line 111: we shouldn't check for `!activeAppIDSnapshot.hasData` here.

When we don't have any supported app (appImplementations.data!.isEmpty) we also don't have an active app.
This check prevents us from showing the noapps view.

provokateurin:
I only see `Error: Could not download app notes`. This sometimes happens because the app store does weird things. Just need to restart the job in that case.

provokateurin (approved):

PR #403, opened 2023-06-19
neon: dispose appVersionCheck subscribtions

Leptopoda:
I'm not yet 100% convinced if this is the right way :thinking: 

What do you think?

### DIFF
--- a/packages/neon/neon/lib/src/pages/home.dart
+++ b/packages/neon/neon/lib/src/pages/home.dart
@@ -18,6 +18,7 @@ class _HomePageState extends State<HomePage> {
   late GlobalOptions _globalOptions;
   late AccountsBloc _accountsBloc;
   late AppsBloc _appsBloc;
+  late StreamSubscription _versionCheckSubscription;
 
   @override
   void initState() {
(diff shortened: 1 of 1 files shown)


packages/neon/neon/lib/src/utils/global_popups.dart
```diff
@@ -7,37 +7,65 @@ class GlobalPopups {
...
       return;
     }
+    _registered = true;
 
-    final globalOptions = Provider.of<GlobalOptions>(context, listen: false);
     final firstLaunchBloc = Provider.of<FirstLaunchBloc>(context, listen: false);
     final nextPushBloc = Provider.of<NextPushBloc>(context, listen: false);
+    final accountsBloc = Provider.of<AccountsBloc>(context, listen: false);
+    final account = accountsBloc.activeAccount.value!;
+    final appsBloc = accountsBloc.activeAppsBloc;
```
provokateurin, line 25: This won't work with multiple accounts because the second register call will not reach this point.

Leptopoda: Ah yes;
I thought about that but probably forgot it again :)

Would it make sense to have a second method that registers those or just a separate singleton?

provokateurin: IMO the purpose of global is to be above any context like an account or app. This kind of defeats the purpose. Maybe a different place is better? You could make another similar singletone `AccountPopups` that gets registered in the home page as well.

Leptopoda:
Ok I scrapped all the cahnges.

We need to make sure to propperly dispose them as go_router might/will push a new HomePage for every app that we navigate to.
We still need to figure out how to handle that the user already clicked the ok button so we don't spam them in deprecation messages on every navigation request.

provokateurin (approved):

PR #405, opened 2023-06-20
ci: Add dependabot

provokateurin:
We could also add docker to allow automatic updates of the Dockerfile used for development and tests, but since we need to adjust the submodules and code for that every time, it doesn't make a lot of sense to have it. If it could open an issue instead of a PR for that it would be a different thing.


Leptopoda (approved):
LGTM

PR #395, opened 2023-06-17
Feature/resultsubject

Leptopoda:

PR #406, opened 2023-06-20
Use Result helper methods

provokateurin:
Cleanup after https://github.com/provokateurin/nextcloud-neon/pull/395

### DIFF
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -34,17 +34,17 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     });
 
     appImplementations.listen((final result) {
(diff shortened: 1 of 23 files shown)


Leptopoda (approved):
Nice :+1: 

maybe we should also use `requireData` instead of `data!`?

provokateurin:
Ah I forgot to look for those. Let me add them

Leptopoda (commented):
we still have some checks (probably only searched for ` !=`)

https://github.com/provokateurin/nextcloud-neon/blob/ed0fe992a00df2238b37be63f136117bd1c7137f/packages/neon/neon_news/lib/widgets/articles_view.dart#L39

provokateurin:
Ah ok, in this case the entire check can be removed. Will search for more occurrences later

Leptopoda (approved):

PR #411, opened 2023-06-22
Refactor/active app

Leptopoda:


packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -13,7 +13,7 @@ abstract class AppsBlocStates {
 
   BehaviorSubject<Result<NotificationsAppInterface?>> get notificationsAppImplementation;
 
-  BehaviorSubject<String?> get activeAppID;
+  BehaviorSubject<AppImplementation> get activeApp;
```
provokateurin, line 16: Should have been done this way from the start :disappointed:

packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -3,7 +3,7 @@ part of 'blocs.dart';
 typedef NextcloudApp = NextcloudCoreNavigationApps_Ocs_Data;
 
 abstract class AppsBlocEvents {
-  void setActiveApp(final String? appID);
+  void setActiveApp(final String appID);
```
provokateurin, line 6: Why is this required?

Leptopoda: 1. I don't see why we should allow it. I mean how would we set the active app to nothing? If we ever need this we could add something like `void clearActiveApp();` .

2. in the case null was passed we weren't doing anything and just returning (plus maybe wasting some time in the await depending on the state of the app). So it didn't really do anything.

provokateurin: I think if you had no compatible apps this method was triggered with a null value, but I guess that was a bug from the start. Please make sure this works when no apps are supported.

Leptopoda: Oh this is indeed something I forgot;
I just forced the nullcheck as I thought initialApp will be set to a value before calling this but having a closer look it can indeed be null.

I'll just check it before calling setActiveApp like:

```dart
if (!activeApp.hasValue && initialApp != null) {
  await setActiveApp(initialApp!);
}
```

provokateurin (approved):

Leptopoda:
Did the CI get flaky again? :(

provokateurin:
The tests are fine. I've been seeing increased network failures during requests, but I don't know why. I think we could enabled retries on failed tests, that should make it more stable.

Issue #323, opened 2023-05-17
Stream based decoding and deserialization (WebDAV and NC)


Leptopoda:
we currently do:
```dart
var chunks = await stream.toList();
var data = Uint8List.fromList(chunks.reduce((final value, final element) => [...value, ...element]));
data = utf8.decode(data);

// work with data
```

while we could decode the stream directly with `transform(utf8.decoder).join()` as suggested by the docs of HttpClientResponse.
I could imagine this being more perfomant and easier to work with than our current approach of iteratively combining the chunks into a growing list.
In theory we could even chain a `transform(json.decoder)` although I don't know how this would handle a single String being returned (as we need strings as strings).

---

We might also be able to use a similar approach for webdav and deserialize the response on the fly again making it more performant.
https://github.com/renggli/dart-xml/issues/84 might be helpful.

PR #413, opened 2023-06-22
Feature/streams

Leptopoda:
Should reduce memory operations quite a bit and thus improve responsivenes/performance.

- We lose the `RawResponse` class as the `HttpClientResponse` takes it's place wich does not have a nice `toString` method.

- We gain some better modularity by having shared stuff in the dynamite_helper class. This also enables us to unit test them easily but I don't see any value in this.

the nextcloud package will export the dynamite_runtime http stuff so other libraries can still make use of it (like the webdav client). While being at it I also made all fields of `DynamiteClient` final instead of some being late final.

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
@@ -14,6 +14,17 @@ extension DynamiteHttpClientResponseBody on HttpClientResponse {
   }
 
   Future<String> get body => transform(utf8.decoder).join();
+
+  Future<dynamic> get jsonBody => transform(utf8.decoder).transform(json.decoder).first;
+
+  Map<String, String> get responseHeaders {
+    final responseHeaders = <String, String>{};
+    headers.forEach((final name, final values) {
+      responseHeaders[name] = values.last;
+    });
+
+    return responseHeaders;
+  }
 }
 
 class DynamiteResponse<T, U> {
@@ -30,8 +41,8 @@ class DynamiteResponse<T, U> {
   String toString() => 'DynamiteResponse(data: $data, headers: $headers)';
 }
 
-class RawResponse {
-  RawResponse(
+class DynamiteApiException implements Exception {
+  DynamiteApiException(
     this.statusCode,
     this.headers,
     this.body,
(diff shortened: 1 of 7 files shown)


packages/dynamite/dynamite_runtime/lib/src/http_client.dart
```diff
@@ -0,0 +1,129 @@
...
+    this.statusCode,
+    this.headers,
+    this.body,
+  );
+
+  final int statusCode;
+
+  final Map<String, String> headers;
+
+  final String body;
```
provokateurin, line 55: Shouldn't this be Uint8List? Just in case we get some binary data but are unable to handle it for whatever reason and then throw the exception. On the other hand this would make it quite annoying to read the error message

provokateurin:
LGTM, just one minor question

provokateurin:
And this would close https://github.com/provokateurin/nextcloud-neon/issues/323, right?

Leptopoda:
> And this would close https://github.com/provokateurin/nextcloud-neon/issues/323, right?

Thanks for the reminder.
We could also use streams for webdav but while xml supports it xml_serializable doesnt.

We can close it with this as the maintainer said that they don't plan to implement it in the foreseeable future. Once they support it we can open it again.

provokateurin:
I think decoding json and xml using streams is not really possible/has no advantages because we only get one response, right? At least that's what I found out for json, not sure about xml

Leptopoda:
The main benefit is when we get a list.
built_value let's us decode json lists with streams. I'd asume that webdav also returns lists somewhere.

But I agree that in most cases this won't benefit us. This can also be seen as the `jsonBody` getter just returns the first element found.

packages/dynamite/dynamite_runtime/lib/src/http_client.dart
Leptopoda: I could switch it back to the binary stuff and only utf.decode it when needed (performance isn't really a problem here) but the old way we just did the same.
Decode it in the toString method. Now if we have binary data the old way would probably have thrown an exception when calling toString. The new way at least throws this in the request itself where we catch various errors.

I think the best way would be to pass the binary data into this but still wrap the toString in a try catch and just printing `binary` in that case.

provokateurin: I like that idea

Leptopoda:
I ended up not doing it in the toString so we don't need to so we don't need to store potentially large binary data in the error object just to discard it again when printing.

Hope this makes sense.

provokateurin (approved):

Issue #177, opened 2023-01-11
Investigate melos


provokateurin:
https://github.com/invertase/melos could be better than mono_repo

Leptopoda:
It even natively supports fvm:
https://melos.invertase.dev/~melos-latest/configuration/overview#sdkpath

provokateurin:
I think using melos and mono_repo together is the way to go. mono_repo really makes the CI parts easier so I want to keep it for that.

Leptopoda:
After some investigation I can see how mono_repo generates clean separated CI configs but I think that the potential of melos is also big.

We can specify commands for everything we need (linting, formating and testing) and then run `melos run lint`, `melos run format-check` and `melos run analyze` in CI.

a really advanced use case is what the firebase team does https://github.com/firebase/flutterfire/blob/master/melos.yaml

and we also get fvm support

provokateurin:
Hm yeah that's quite cool. We can drop mono_repo if we have a good replacement for the CI parts

PR #414, opened 2023-06-23
Feature/melos

Leptopoda:
We'll need the `melos run generate_neon` to efficiently generate all the go_router code for all apps.
While being at it I also migrated the formatting and l10n generation to melos. I think we can now close: #177 with this.

I removed our custom scripts as I don't think they are needed anymore. I also think that the mono_repo leftover can safely be removed now?


provokateurin:
Ah I oversaw the mono_repo file, it should have been removed with my recent PR

melos.yaml
```diff
@@ -22,3 +22,6 @@ scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test"
+  generate_neon: melos exec --scope="neon*" -- dart run build_runner build --delete-conflicting-outputs
+  format: melos exec dart format --fix --line-length 120 .
+  gen-l10n: melos exec --flutter --dir-exists="lib/l10n" flutter gen-l10n && melos run format
```
provokateurin, line 27:
```suggestion
  generate:build_runner: melos exec --scope="neon*" -- dart run build_runner build --delete-conflicting-outputs
  format: melos exec dart format --fix --line-length 120 .
  generate:l10n: melos exec --flutter --dir-exists="lib/l10n" flutter gen-l10n && melos run format
```

What do you think about this?
I think it is better because the names use the same format.
I also changed it from `neon` to `build_runner` because that's more accurate and avoids confusion with our `generate-neon.sh` script.

melos.yaml
```diff
@@ -22,3 +22,6 @@ scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test"
+  generate_neon: melos exec --scope="neon*" -- dart run build_runner build --delete-conflicting-outputs
+  format: melos exec dart format --fix --line-length 120 .
```
provokateurin, line 26:
```suggestion
  format: dart format --fix --line-length 120 .
```

You don't need to run it for every package individually. See the `format-check` script.

tool/generate-neon.sh
```diff
@@ -33,8 +33,8 @@ copy_nextcloud_app_svg notifications external/nextcloud-notifications
...
   wget https://raw.githubusercontent.com/nextcloud/promo/master/nextcloud-logo-inverted.svg -O assets/logo_nextcloud.svg
 
-  fvm dart run build_runner build --delete-conflicting-outputs
-  fvm flutter gen-l10n
+  melos run generate_neon
+  melos run gen-l10n
 )
 
-./tool/format.sh
+ melos run format
```
provokateurin, line 40:
```suggestion
melos run format
```

melos.yaml
```diff
@@ -22,3 +22,6 @@ scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test"
+  generate_neon: melos exec --scope="neon*" -- dart run build_runner build --delete-conflicting-outputs
```
provokateurin, line 25: I noticed this also runs on packages that don't use build_runner. How about adding an empty `build.yaml` everywhere where we use build_runner and then filter for that file?

provokateurin (changes requested):

melos.yaml
Leptopoda: I limited the scope to only run build_runner in `packages/neon` to avoid running dynamite/generating nextcloud wich can be a lot sometimes (and suspected to grow once we add more specs and fix oom).

I also named the l10n command to be the same as the flutter built in one but that doesn't need to be the case

provokateurin: Right, but then we still have confusion with the generate-neon.sh script?

melos.yaml
Leptopoda: Yeah that's bad.

I just cherry picked this from my go_router branch and there every app will be expected to use build_runner for the router. In that case it wouldn't be needed.

Should I change this anyways?

melos.yaml
Leptopoda: what about `generate:neon:buildrunner` ?

melos.yaml
provokateurin: Yeah, at least for this PR please. You can remove it later on. But it might happen that some app won't use go_router, right? Like a really simple single page app (i don't know any, but it could happen).

melos.yaml
provokateurin: Perfect

packages/neon/neon/lib/src/router.g.dart
provokateurin, line 1: I think you missed running the formatter.

Leptopoda: It's all garbage :)

I also ammended it to the wrong commit xD

Leptopoda: Should the generate command also run the formatter automatically?

provokateurin: Would be nice, yes

melos.yaml
```diff
@@ -22,3 +22,6 @@ scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
+  generate:neon:build_runner: melos exec --scope="neon*" --file-exists="build.yaml" -- dart run build_runner build --delete-conflicting-outputs && melos run format
+  format: dart format --fix --line-length 120 .
+  generate:l10n: melos exec --flutter --dir-exists="lib/l10n" flutter gen-l10n && melos run format
```
provokateurin, line 27:
```suggestion
  generate:neon:l10n: melos exec --flutter --dir-exists="lib/l10n" flutter gen-l10n && melos run format
```

Could also do this since it's the only use case, but technically not limited so can leave it as well

tool/generate-neon.sh
```diff
@@ -33,8 +33,8 @@ copy_nextcloud_app_svg notifications external/nextcloud-notifications
   # Nextcloud logo
   wget https://raw.githubusercontent.com/nextcloud/promo/master/nextcloud-logo-inverted.svg -O assets/logo_nextcloud.svg
 
-  fvm dart run build_runner build --delete-conflicting-outputs
-  fvm flutter gen-l10n
+  melos run generate_neon
+  melos run gen-l10n
```
provokateurin, line 37: Wrong names

tool/generate-app.sh
```diff
@@ -37,7 +37,7 @@ function export_mipmap_icon_all() {
   wait
 
   fvm dart run flutter_native_splash:create
-  fvm flutter gen-l10n
+  melos run gen-l10n
```
provokateurin, line 40: Wrong name

melos.yaml
```diff
@@ -22,3 +22,6 @@ scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
+  format: dart format --fix --line-length 120 .
```
provokateurin, line 25: Maybe move this one before or after the format-check script

tool/generate-app.sh
```diff
@@ -37,7 +37,7 @@ function export_mipmap_icon_all() {
   wait
 
   fvm dart run flutter_native_splash:create
-  fvm flutter gen-l10n
+  melos run generate:neon:l10n
```
provokateurin, line 40: I think this doesn't even belong here in the first place. Generating all the app assets etc. is separate from generating l10n.

provokateurin (changes requested):

provokateurin (approved):

PR #416, opened 2023-06-25
Make tests go brrr

provokateurin:
This cuts down the time to run all nextcloud tests on my machine by about 10x.

The thing that improves the speed the most is using php instead of apache to run the server because it apparently takes ages to startup. It usually spent almost all the time of a test run just waiting for the webserver. About 5x improvement.

Using concurrency cut down the speed by about 2x again, though in the end its slower because there are not enough test suites available to run in parallel. I wonder if they can be split up more to make it even faster.


melos.yaml
```diff
@@ -21,4 +21,4 @@ ide:
 scripts:
   format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
-  test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test"
+  test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
```
Leptopoda, line 24: duplicate concurrency argument

provokateurin: But the melos concurrency is for all package and the flutter test concurrency is inside each package. The over all concurrency is limited to 1 to avoid producing garbage output

Leptopoda (approved):
That's awesome :tada:
3x faster on my side. That's very helpful.

Couldn't we cache the built docker image? What was the issue with the layer cache you tested?
I think we should not rebuilt it on every run:
1. The app store can do it's things and fail the run
2. It takes time
3. we should keep the server we are testing against stable (We are already pinning the server and apps version).

If it's not suitable to cache the docker layers we could just cache the entire image by pushing it to ghcr.

provokateurin:
Yeah I tried the docker layer caching, but the action wasn't working. Maybe there is a better one that we can use. Docker layer caching for the publish workflow would also be very helpful, but that workflow needs some general cleanup.

provokateurin:
> We are already pinning the server and apps version

This isn't true for apps because there is no way to tell the occ install command which version we want. I already opened an issue in the server repo for that, but I don't think it will be fixed soon. We could try downloading the tarballs ourselves and installing them manually to pin the app versions correctly.

PR #418, opened 2023-06-26
Refactor/dynamite runtime

provokateurin:


packages/nextcloud/lib/src/nextcloud.openapi.dart
```diff
@@ -10,68 +10,11 @@ import 'package:built_value/serializer.dart';
...
-  NextcloudResponse(
-    super.data,
-    super.headers,
-  );
-
-  @override
-  String toString() => 'NextcloudResponse(data: $data, headers: $headers)';
-}
-
-class NextcloudApiException extends DynamiteApiException {
```
Leptopoda, line 29: I'd still like having a separate exception.

An app could also include a nextcloud unrelated openapi schema and this makes it impossible to distinguish them.

packages/nextcloud/lib/src/nextcloud.openapi.dart
```diff
@@ -10,68 +10,11 @@ import 'package:built_value/serializer.dart';
...
 part 'nextcloud.openapi.g.dart';
 
-class NextcloudResponse<T, U> extends DynamiteResponse<T, U> {
-  NextcloudResponse(
-    super.data,
-    super.headers,
-  );
-
-  @override
-  String toString() => 'NextcloudResponse(data: $data, headers: $headers)';
```
Leptopoda, line 26: same as below.
The toString cuttently contains the prefix wich could make debugging easier.

Leptopoda (commented):
Having the Auth in dynamite_runtime makes sense ut we should still have custom types to better distinguish/debug the case where we have multiple separate generated apis.

packages/nextcloud/lib/src/nextcloud.openapi.dart
provokateurin: Why would it? We have stacktraces, right?

Leptopoda: I was acrually thinking of trying to handle the error.
Imagine hypothetical case where a dev might want to generate two separate specs (dynatime should be general purpose) and the dev might want to handle the errors differently (in a try catch or the UI).

For this they'd need to know what they are dealing with.

This is just dynamite stuff and we'd throw the spec specific one (NewsError, NotesError) but neon could just handle the parent (DynamiteError).

provokateurin: Makes sense, although I don't know if anyone will ever have this use-case. As you said for neon we should just handle the generic DynamiteException.

Leptopoda (approved):

PR #423, opened 2023-06-26
Feature/nested router

Leptopoda:
I think it's ready for a first review.

TODO:
- [x] bad web support: Currently the AppsBloc.activeApp and the go_router don't work together. this means a user on the web could enter a supported url but the appsbloc would not update it's state. The adaptive layout branch might change this but it needs a fix for now (even though we don't target web yet)
- [x] The router get's build once meaning that all supported apps by neon are added to the routes list. On mobile this is not a problem but on web a user could enter the url of an app that isn't supported by the server or even disabled for that user in a multi account setup.
- [x] The ShellRoute handling the HomePage is not stateful (StatefulShellRoute only became a thing recently after I had started the go_router work).  This means that a user navigating back from the settings page will get an empty (loading like) page.
- [x] AppImplementation does throw an error when no route is specified. We should make it required.
- [ ] All apps got a new static appID field that will be used in the routes. We should replace all occurrences of the old hard coded values (probably a followup PR and especially useful in the AppsBloc).
- [x] The doc comment of AppsBloc._initialAppID could be improved also including the comment in it
- [ ] make use of go_router in all apps (mainly the files app). Probably a followup PR for each app
- [ ] opening an app from the system tray is broken

### DIFF
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -50,27 +50,22 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     });
 
     appImplementations.listen((final result) {
-      if (result.hasData) {
-        final options = _accountsBloc.getOptionsFor(_account);
-        unawaited(
-          options.initialApp.stream.first.then((var initialApp) async {
-            if (initialApp == null) {
-              if (result.requireData.tryFind('files') != null) {
-                initialApp = 'files';
-              } else if (result.requireData.isNotEmpty) {
-                // This should never happen, because the files app is always installed and can not be removed, but just in
-                // case this changes at a later point.
-                initialApp = result.requireData.first.id;
-              }
-            }
-            if (!activeApp.hasValue && initialApp != null) {
-              await setActiveApp(initialApp);
-            }
-          }),
-        );
-
-        unawaited(_checkCompatibility());
+      if (!result.hasData) {
+        return;
       }
+
+      final options = _accountsBloc.getOptionsFor(_account);
+      unawaited(
+        options.initialApp.stream.first.then((var initialApp) async {
+          initialApp ??= _getInitialAppFallback();
+
+          if (!activeApp.hasValue && initialApp != null) {
+            await setActiveApp(initialApp);
+          }
+        }),
+      );
+
+      unawaited(_checkCompatibility());
     });
 
     _capabilitiesBloc.capabilities.listen((final result) {
@@ -84,6 +79,25 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     });
 
     unawaited(refresh());
+  }
+
+  /// Determines the appid of initial app.
+  ///
+  /// It requires [appImplementations] to have both a value and data.
+  ///
+  /// The files app is always installed and can not be removed so it will be used, but in the
+  /// case this changes at a later point the first supported app will be returned.
+  ///
+  /// Returns null when no app is supported by the server.
+  String? _getInitialAppFallback() {
+    final supportedApps = appImplementations.value.requireData;
+    if (supportedApps.tryFind('files') != null) {
+      return 'files';
+    } else if (supportedApps.isNotEmpty) {
+      return supportedApps.first.id;
+    }
+
+    return null;
   }
 
   Future<void> _checkCompatibility() async {
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -9,6 +9,7 @@ import 'package:neon/src/blocs/capabilities.dart';
 import 'package:neon/src/models/account.dart';
 import 'package:neon/src/models/app_implementation.dart';
 import 'package:neon/src/models/notifications_interface.dart';
+import 'package:neon/src/router.dart';
 import 'package:neon/src/settings/models/nextcloud_app_options.dart';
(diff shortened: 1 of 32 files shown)


packages/neon/neon/lib/src/utils/app_implementation.dart
```diff
@@ -53,6 +53,25 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppSpecificO
     );
   }
 
+  /// Route for the app.
+  ///
+  /// All pages of the app must be specified as subroutes.
+  /// If this is not [GoRoute] an inital route name must be specified by overriding [initialRouteName].
+  RouteBase get route => throw UnimplementedError();
```
provokateurin, line 60: How about making this abstract? Then every app needs to explicitly set this. I think that leads to less confusion

packages/neon/neon/lib/src/utils/app_route.dart
```diff
@@ -0,0 +1,16 @@
...
+abstract class NeonAppRoute extends GoRouteData {
+  const NeonAppRoute();
+
+  @override
+  Page<void> buildPage(final BuildContext context, final GoRouterState state) => NoTransitionPage(
+        child: build(context, state),
+      );
+}
+
+const appRoutSuffix = '/apps/';
```
provokateurin, line 16:
```suggestion
const appsRouteSuffix = '/apps/';
```

packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -70,6 +65,23 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     unawaited(refresh());
   }
 
+  /// Determines the id of initial app.
+  ///
+  /// It requires [appImplementations] to have both a value and data.
+  /// Returns null when the server does not any app.
+  String? get _initialAppID {
```
provokateurin, line 72:
```suggestion
  String? _getInitialAppFallback() {
```

I don't like that it has so much logic just for a getter function.
You could also add the initialApp as a paramter that is returned instead of null, that would make it a bit cleaner even.

provokateurin (changes requested):
Already looks pretty good.
I think for this PR you can leave out the web stuff, we don't really care about it right now. Follow-up PR is of course welcome.

packages/neon/neon/lib/src/utils/app_implementation.dart
Leptopoda: I'ts already on the todo. I made it this way so every single commit is valid and can compile

Leptopoda:
It took some time but we have a lot of edge cases (using the back button on the loginflow when adding another account)

It should finally be ready

packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -186,6 +202,11 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     if (app != null) {
       if (activeApp.valueOrNull?.id != appID) {
         activeApp.add(app);
+
+        // avoid inactive accounts calling the navigator witha potentially unsupported app.
```
provokateurin, line 206:
```suggestion
        // avoid inactive accounts calling the navigator with a potentially unsupported app.
```

provokateurin (approved):
Very nice :heart:

provokateurin:
@Leptopoda any other changes than rebasing? If not then feel free to merge :)

Leptopoda:
The HomePage/HomeRoute handling is different
It's now a StatefulSHellRoute

provokateurin:
Ok, I'll review again

packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -68,6 +69,34 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
     );
   }
 
+  /// Main branch displayed ini the home page.
```
provokateurin, line 72:
```suggestion
  /// Main branch displayed in the home page.
```

packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -68,6 +69,34 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
...
+
+  /// Route for the app.
+  ///
+  /// All pages of the app must be specified as subroutes.
+  /// If this is not [GoRoute] an inital route name must be specified by overriding [initialRouteName].
+  RouteBase get route;
+
+  /// Name of the initial route for this app.
+  ///
+  /// Subclasses that don't provide a [GoRoute] for [route] must ovveride this.
```
provokateurin, line 89:
```suggestion
  /// Subclasses that don't provide a [GoRoute] for [route] must override this.
```

provokateurin (changes requested):
I found some problems:

Opening the settings page still keeps the old app selected and it looks weird because it's suddenly inside the home page and not pushed as another page. Was that intentional?

Starting the app it shows only the app bar and nothing else, even though cached data is present. Should be reproducible with a slow connection or the server being offline.

provokateurin:
Maybe there is more, this is what I found after clicking through it

Leptopoda:
Yeah ...

We might need to bring the appBar under the control of the widgets.
Our structure, tho one adaptive_layout wants, and the one of go_router are very hard to combine and I'll probably drop the adaptive_layout idea for now (maybe even for the entire 1.0).

Leptopoda:
I'll open a new PR once ready as I already archived my fork

PR #426, opened 2023-06-27
neon: cleanup NeonException

Leptopoda:


packages/neon/neon/lib/src/widgets/exception.dart
```diff
@@ -25,88 +25,89 @@ class NeonException extends StatelessWidget {
...
+                ),
+              ),
+            ],
+          ),
+          ElevatedButton(
+            onPressed: () async {
+              if (details.isUnauthorized) {
+                _openLoginPage(context);
+              } else {
+                // TODO: why no retry?
```
Leptopoda, line 102: The old implementation did not call onRetry in this case but the iconButton does.
Is this intended?

Leptopoda: I'm probably blind as the old implementation clearly calls onRetry :woman_facepalming:

provokateurin (approved):

PR #432, opened 2023-06-29
Fix/list cached image

Leptopoda:
Adding the key did the trick :)

I'm not 100% sure whether I like them being factory constructors. What do you think?
I've tested it in the files app in a folder with multiple images, deleting some of them.


provokateurin:
> I'm not 100% sure whether I like them being factory constructors.

I think it's a good move, the two different widgets didn't really make sense.

provokateurin:
Can confirm it works :tada:

packages/neon/neon/lib/src/widgets/cached_image.dart
```diff
@@ -2,18 +2,81 @@ part of '../../neon.dart';
 
 final _cacheManager = DefaultCacheManager();
 
-abstract class NeonCachedImage extends StatefulWidget {
-  const NeonCachedImage({
+typedef APIImageDownloader = Future<Uint8List> Function();
```
provokateurin, line 5: The name looks a bit odd because of the double `I`'s but I don't have a better name

packages/neon/neon/lib/src/widgets/cached_image.dart
```diff
@@ -27,58 +90,60 @@ abstract class NeonCachedImage extends StatefulWidget {
 }
 
 class _NeonCachedImageState extends State<NeonCachedImage> {
-  late Future<File> future = widget.getImageFile();
+  late Future<Uint8List> _future = widget.getImageFile();
```
provokateurin, line 93: I'm not sure if it's good to have the underscore in stateful widgets. I tend to not have them because it's quite irrelevant, but I can also see why you still want to make it private. What do you think?

provokateurin: At least it makes it a bit clearer that the scope is smaller and it's not a global variable (if we wouldn't have all the library thingies).

Leptopoda:
I'll think about how we can make this cleaner.

The problem is that it's hard (currently impossible) to extend this widget.
Factories are useless in that regard and the constructor is private. I thought about this as it might make sense giving `FilePreview` it's own CachedImage.

Maybe we can move the functions into static methods and use them in a normal constructor :thinking: 

Finally most CachedImages are wrappen in `NeonImageWrapper`. And scrolling fast through a listview will flicker the image wrapper background color. This should also be fixed (flickering is always bad. Especially for a11y).

Maybe you can give it a review and I'll make some followup PRs later.

packages/neon/neon/lib/src/widgets/cached_image.dart
```diff
@@ -2,18 +2,81 @@ part of '../../neon.dart';
...
+        fit: fit,
+        svgColor: svgColor,
+        iconColor: iconColor,
+        key: key ?? Key(url),
+      );
+
+  factory NeonCachedImage.api({
+    required final Account account,
+    required final String cacheKey,
+    required final APIImageDownloader download,
```
provokateurin, line 42: I wonder if it would be possible to have a hint based on the returned content type. Currently we don't have access to the raw API response so we can't access it, but maybe you could leave a todo note about that here.

packages/neon/neon/lib/src/widgets/cached_image.dart
Leptopoda: The state is already 'private' so you are right.

packages/neon/neon/lib/src/widgets/cached_image.dart
Leptopoda: I have local changes with a FilesCachedImage where it just takes the FilesDetails and does exactly that. `details.mimeType?.contains('svg')` as a Hint.


And that's the reason why we should think of how we can extend this.

packages/neon/neon/lib/src/widgets/cached_image.dart
Leptopoda: just removed the API :)

Leptopoda:
Ok I think it turned out quite nice.
What do you think @provokateurin.

We currently miss the refresh on error functionality and I'll need to reorder some commits to keep the history clean.

packages/neon/neon_files/lib/widgets/file_preview.dart
```diff
@@ -62,13 +37,82 @@ class FilePreview extends StatelessWidget {
...
+      account: account,
+      file: file,
+      size: size,
+      cacheKey: cacheKey,
+      path: path,
+      width: width,
+      height: height,
+    );
+  }
+  FilePreviewImage._({
```
provokateurin, line 94: Newline between the constructors

packages/neon/neon_files/lib/widgets/file_preview.dart
```diff
@@ -62,13 +37,82 @@ class FilePreview extends StatelessWidget {
...
+          ),
+          writeCache: (final cacheManager, final data) async {
+            await cacheManager.putFile(
+              cacheKey,
+              data,
+              maxAge: const Duration(days: 7),
+              eTag: file.etag,
+            );
+          },
+          isSvgHint: file.mimeType?.contains('svg') ?? false,
```
provokateurin, line 116: Noice

provokateurin (approved):

PR #434, opened 2023-06-30
Fix/flatpak

provokateurin:
The flatpak wasn't really useful without the desktop file and building and running the arm64 version was broken.
Now I can launch the arm64 on my x86_64 trough qemu binfmt using the menu entry of my DE :sunglasses:


Leptopoda (approved):
I've no experience building flatpaks so I need to trust you that it works :)

Issue #433, opened 2023-06-29
Breakup libraries


provokateurin:
Currently we have many libraries that are used to melt everything together.
They should be broken up into multiple libraries where it makes sense (e.g. all blocs together) or just become standalone files.

Leptopoda:
I think there are two sides we should consider.

1. How do we export them?
2. How do we handle them internally?

We should probably export one per folder
- `package:neon/widgets.dart`
- `package:neon/blocs.dart`
- `package:neon/models.dart` (could also container files from interfaces)
- `package:neon/utils.dart`

Internally I don't think we should use `part` directives at all. But we can still decide this after the first one is done.

provokateurin:
So you would like to have a "summary" file that exports all the other files that are in a given directory instead of using the libraries?

Leptopoda:
That's at least how the flutter framework does it but I don't know if that actually fits us as well.

https://github.com/flutter/flutter/tree/master/packages/flutter/lib

provokateurin:
I see, I think for app implementations it makes a lot of sense to have it like that.

How about the app implementations themselves? I think they can still use a single library using part directives.

Leptopoda:
As they don't really provide a public api the visibility doesn't really matter.
Also changing this after the 1.0 milestone wouldn't be breaking so we can leave them for now.

Leptopoda:
We should probably also break up the nextcloud lib into:
- client
- webdav
- api

with the later two also exposing the first one.

provokateurin:
I'll add that after https://github.com/provokateurin/nextcloud-neon/pull/439 is merged and I have rebased https://github.com/provokateurin/nextcloud-neon/pull/436 onto it. It has pretty big changes and already goes a little bit into that direction.

PR #436, opened 2023-07-01
Refactor/libraries

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/433


packages/neon/neon/lib/utils.dart
```diff
@@ -0,0 +1,12 @@
+export 'package:neon/src/utils/app_implementation.dart';
```
Leptopoda, line 1: I never liked this being a util.

It should rather be a model or interface but I don't know how much it makes sense :(

packages/neon/neon/lib/blocs.dart
Leptopoda, line 1: Should this also export bloc?
How much does it make sense to have them separated once the AccountsBloc isn't exported anymore?

packages/neon/neon/lib/utils.dart
```diff
@@ -0,0 +1,12 @@
...
+export 'package:neon/src/utils/branding.dart';
+export 'package:neon/src/utils/confirmation_dialog.dart';
+export 'package:neon/src/utils/exceptions.dart';
+export 'package:neon/src/utils/hex_color.dart';
+export 'package:neon/src/utils/nextcloud_app_specific_options.dart';
+export 'package:neon/src/utils/rename_dialog.dart';
+export 'package:neon/src/utils/request_manager.dart';
+export 'package:neon/src/utils/sort_box_builder.dart';
+export 'package:neon/src/utils/sort_box_order_option_values.dart';
+export 'package:neon/src/utils/storage.dart';
```
Leptopoda, line 11: moved to settings?

Leptopoda (changes requested):
Some random thoughts. They may not all make sense.

- I think that all libs should add the `library` directive with a small little doc comment.
- All private stuff should be annotated with `@internal` so we don't accidentally export it (directly or indirectly). This might be useful for internal widgets and the blocs.

Leptopoda:
I just noticed that neon is directly exposing types defined in nextcloud.

If dynamite or the specs would change some naming this would not only result in breaking changes to nextcloud but also in neon.
We should abstract away all nextcloud types that we expose into their own typedefs. Internally we can still use the normal types.

packages/neon/neon/lib/interfaces.dart
```diff
@@ -0,0 +1 @@
+export 'package:neon/src/interfaces/notifications.dart';
```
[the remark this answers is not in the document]
provokateurin: We could do that, but for me models is something to model concrete data and not to model some abstract interface that gets implemented elsewhere

provokateurin:
I agree with you on most points. Moving to models makes sense :+1:

packages/neon/neon/lib/blocs.dart
provokateurin: Yeah let's expose them together as blocs.

provokateurin:
@Leptopoda  Should we also export file_icons and nextcloud through the neon package?

packages/neon/neon/lib/src/widgets/drawer_destination.dart
Leptopoda, line 1: should be exported as it's gonna be needed for the adaptive layout.

Or should I just expoprt it then?

packages/neon/neon/lib/src/utils/stream_listenable.dart
Leptopoda, line 1: Could be exported (doesn't need to)

packages/neon/neon/lib/settings.dart
Leptopoda, line 1: Could we structure the settings folder in a way similar to all other packages?

/settings
|-> widgets
|-> models
| -> ...

it's currently all over the place

Leptopoda (commented):
I'd really like to have internal annotations on things that we don't (want to ever) export.
[0001-neon-add-internal-annotations.patch.txt](https://github.com/provokateurin/nextcloud-neon/files/11929134/0001-neon-add-internal-annotations.patch.txt)

packages/neon/neon/lib/src/widgets/drawer_destination.dart
provokateurin: Yeah just export it when needed. I only added exports for things that are used outside the neon package right now

packages/neon/neon/lib/src/utils/stream_listenable.dart
provokateurin: Just export it when you need it

Leptopoda:
> @Leptopoda Should we also export file_icons and nextcloud through the neon package?

Yeah this would probably make thing easier for devs.
But I don't know where :(
The neon library is currently not needed for apps (like files, notes ...) and only needed the `app` itself.
We'd need to come up with a better place for this.

provokateurin:
> I'd really like to have internal annotations on things that we don't (want to ever) export.

I looked at your patch and I don't agree with all of it. For example `NeonUserAvatar` will be needed for Talk as we want to display the user avatars there.

provokateurin:
>  We'd need to come up with a better place for this.

Ok let's keep this in mind and do it at some later point

Leptopoda:
> I looked at your patch and I don't agree with all of it. For example NeonUserAvatar will be needed for Talk as we want to display the user avatars there.

I added it to every major internal class. I didn't know you are currently only exporting used things.
I'm fine with not everything having them but things like the theme should probably have them.

Leptopoda (approved):
LGTM

PR #438, opened 2023-07-01
app,dynamite_runtime,neon,neon_files,neon_news: Cleanup unused ignores

provokateurin:
I removed all of the ignores and only added back where the linter complained. These were the ones that got removed and weren't necessary.

### DIFF
--- a/packages/app/test_driver/integration_test.dart
+++ b/packages/app/test_driver/integration_test.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:integration_test/integration_test_driver_extended.dart';
 
 Future<void> main() async {
@@ -15,7 +16,6 @@ Future<void> main() async {
       },
     );
   } catch (e) {
-    // ignore: avoid_print
-    print('Error occurred: $e');
+    debugPrint('Error occurred: $e');
   }
 }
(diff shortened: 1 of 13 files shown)


packages/neon/neon/lib/src/widgets/text_settings_tile.dart
```diff
@@ -1,4 +1,3 @@
-// ignore_for_file: use_late_for_private_fields_and_variables
 // ^ This is a really strange false positive, it goes of at a very random place without any meaning. Hopefully fixed soon?
```
Leptopoda, line 1: leftover

packages/app/pubspec.lock
Leptopoda, line 1: This is my fault, where I forgot to include these changes on my danamite_runtime refactor but I think this should be a separate PR.

packages/app/test_driver/integration_test.dart
```diff
@@ -15,7 +16,6 @@ Future<void> main() async {
       },
     );
   } catch (e) {
-    // ignore: avoid_print
-    print('Error occurred: $e');
+    debugPrint('Error occurred: $e');
```
Leptopoda, line 19: Something for 2.0 could be using a proper logging framework :eyes:

Leptopoda (approved):

Issue #419, opened 2023-06-26
Genrate libraries per spec


Leptopoda:
Far various reasons we should generate libs per spec:
- news.openapi.dart
- notes.openapi.dart
- $app.openapi.dart

depends on #418

provokateurin:
Important for https://github.com/provokateurin/nextcloud-neon/issues/417:

One problem with this will be that the openapi-extractor correctly merges all the capability types when generating the merged spec. We need to deal with that too, otherwise the capabilities become an untyped object.

I'd say we also generate the merged spec and then extract the merged capabilities type from there and inject it into the core spec. That way we don't need to rewrite the capabilities merging again.

provokateurin:
Depending on which of these issues is addressed first, the task of implementing the merging will go to the one that is still open.

Leptopoda:
Should we do this one first so we can find and fix potential issues with dynamite before doing the jump to the automatic specs?

provokateurin:
Yep, let's do that

PR #439, opened 2023-07-01
Refactor/split specs

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/419


packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -247,7 +247,29 @@ class OpenAPIBuilder implements Builder {
...
+                        ${prefix}Client(
+                          client.baseURL,
+                          baseHeaders: client.baseHeaders,
+                          httpClient: client.httpClient,
+                          cookieJar: client.cookieJar,
+                          authentications: client.authentications,
+                        )
+                        '''),
+                    ),
+                  ]);
```
Leptopoda, line 272: This doesn't need to be a factory.

```dart
  UserStatusClient.fromClient(final DynamiteClient client)
      : super(
          client.baseURL,
          baseHeaders: client.baseHeaders,
          httpClient: client.httpClient,
          cookieJar: client.cookieJar,
          authentications: client.authentications,
        );
```

packages/nextcloud/lib/nextcloud.dart
```diff
@@ -1,24 +1,17 @@
...
 export 'package:crypton/crypton.dart' show RSAKeypair, RSAPrivateKey, RSAPublicKey;
 
-export 'src/nextcloud.openapi.dart' hide NextcloudClient;
+export 'src/api/core.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/news.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/notes.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/notifications.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/provisioning_api.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/uppush.openapi.dart' hide jsonSerializers, serializers;
+export 'src/api/user_status.openapi.dart' hide jsonSerializers, serializers;
```
Leptopoda, line 9: we should annotate the hidden fields with @internal so we don't accidentally expose them in the future.

You probably need to add the `meta` package to nextcloud first.

Leptopoda (approved):

Leptopoda (approved):
Would it make sense to just make the serializer private and leave the default name but add a getter or other variable with the public name?

```dart

@SerializersFor([ ])
final Serializers _serializer = ...;

Serializer get userStatusSerializers => _serializer;
```

This way we wouldn't need to pass the name around all the time.

provokateurin:
Oh that's a smart idea

Leptopoda (approved):

provokateurin:
@Leptopoda Codeberg is down and now our CI is failing to build the docker image. I ran the whole pipeline locally, is it ok to merge?

provokateurin:
Actually service seems to be restored just now, I'll give the CI another try

PR #441, opened 2023-07-02
Feature/adaptive layout

Leptopoda:
First step towards: #307 
Implements the layout as described in #329

Now it's up to the app to provide an adaptive layout. I'll migrate them together with the routing changes once this and the routing changes have landed.


provokateurin:
Does this depend on https://github.com/provokateurin/nextcloud-neon/pull/423?

Leptopoda:
Nope :)

But I only want to go through each app once so I'd like them both being merged before I migrate them.

provokateurin:
I like how you adapted the layout of the Android Nextcloud app. Having the account switcher behind the user avatar makes a lot of sense :+1:

provokateurin:
I noticed the animation for switching between drawer and tabbar layout is quite slow, can you make it faster?

provokateurin:
The drawer still contains the settings route, but that can also be accessed from the user avatar now. I'd remove the drawer one to unify the experience (because with the tabbar you don't have another settings entry).

packages/neon/neon/lib/src/widgets/app_bar.dart
```diff
@@ -68,19 +68,103 @@ class NeonAppBar extends StatelessWidget implements PreferredSizeWidget {
...
+          accountList.add(tile);
+        }
+
+        final body = Column(
+          children: [
+            NeonAccountTile(
+              account: aa,
+              trailing: const Icon(Icons.check_circle),
+              onTap: Navigator.of(context).pop,
+            ),
```
provokateurin, line 119: This will always show the current account at the top which is good, but it also changes the order of the displayed accounts depending on which account is active. Not sure which is better/worse to have

packages/neon/neon/lib/src/widgets/app_bar.dart
```diff
@@ -68,19 +68,103 @@ class NeonAppBar extends StatelessWidget implements PreferredSizeWidget {
...
+            const Divider(),
```
provokateurin, line 120: This divider is a bit useless imo. The active account is already indicated by the check mark.

Leptopoda:
> I noticed the animation for switching between drawer and tabbar layout is quite slow, can you make it faster? The animation also plays when you start the app in the tabbar layout. Would be nice to just show it instantly.

I'll have a look.

> The drawer still contains the settings route, but that can also be accessed from the user avatar now. I'd remove the drawer one to unify the experience (because with the tabbar you don't have another settings entry).

I thought I'll need it for the tablet layout where we won't have the tabbar but in that case I'll just reuse the AppBar.
The NC-Android app uses the settings in the account switcher to open an account settings page and doesn't include them in the main settings. Should we go a similar route?

> This will always show the current account at the top which is good, but it also changes the order of the displayed accounts depending on which account is active. Not sure which is better/worse to have

This is how the NC app does it this way and when having more than two (I currently have four setup) makes this approach cleaner.
NC also has the `set status` option below the active one and I thought this might be a good place to add it once we support this.


> This divider is a bit useless imo. The active account is already indicated by the check mark.
https://m3.material.io/components/dialogs/specs

---

I agree that we don't need to follow the android app but I quite like the way they have all account/profile settings in one place.

provokateurin:
>  The NC-Android app uses the settings in the account switcher to open an account settings page and doesn't include them in the main settings. Should we go a similar route?

Yep, let's do it

> This is how the NC app does it this way and when having more than two (I currently have four setup) makes this approach cleaner.

Sure, I usually only have one or two so I can't judge that much :)

>  NC also has the set status option below the active one and I thought this might be a good place to add it once we support this.

Yes, I'll open an issue

Leptopoda:
Should we also remove the account switcher from the drawer?
@provokateurin

provokateurin:
Yes please :) I didn't even see it because I only had one account logged in...

Leptopoda:
The HomePage/HomeRoute handling is different
It's now a StatefulSHellRoute

Issue #447, opened 2023-07-06
Update logo and use adaptive icons


provokateurin:
The accent color is controlled by the server theming and thus can be quite bad as a background color for the logo. It should just use black or whit depending on the device theme.

provokateurin:
Using adaptive icons generated with flutter_launcher_icons should do the trick.

provokateurin:
Color should be changed to be the Nextcloud blue, but the logo should have a blue background with a white icon.

provokateurin:
https://nextcloud.com/trademarks/ describes how the gradient is done. I can try my best to do some inkscape+imagick trickery, but it's probably a lot harder and using a proper tool would help a lot.

Leptopoda:
The problem isn't how to get a rasterized image from the background but rather how to get a drawable from our current icons.

We get SVGs from the apps upstream source but would need an android vector drawable for monochrome icons (and also preferable for adaptive icons). 
pub.dev has https://pub.dev/packages/svg_to_avd but I don't know how good it is.

provokateurin:
The problem with generating multiple icons will be solved with https://github.com/nextcloud/neon/issues/747

PR #448, opened 2023-07-06
neon: Fix Android app switcher background

provokateurin:
Fixes https://github.com/provokateurin/nextcloud-neon/issues/447

### DIFF
--- a/packages/neon/neon/lib/src/app.dart
+++ b/packages/neon/neon/lib/src/app.dart
@@ -287,6 +287,9 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, tray.Tra
                   builder: (final context, final capabilitiesSnapshot) {
                     final nextcloudTheme = capabilitiesSnapshot.data?.capabilities.theming;
                     return MaterialApp.router(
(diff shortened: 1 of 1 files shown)


Leptopoda:
Can't we change the android icon itself to do this?
Why should any flutter code get involved?

If we go the route of changing our icons we should make them wiggle as adaptive icons should :)
https://pub.dev/packages/flutter_launcher_icons could be interesting.

Leptopoda (commented):
Other than that if we go the way of the PR we should probably move this into a separate variable or getter as the build method is already getting a bit cluttered.

provokateurin:
I need to check out flutter_launcher_icons again, the last time I check it was pretty unmaintained and broken.

Leptopoda:
This was just an idea so we don't have to use it (I last used it a month ago in a production app on iOS/Android and it fit our needs).
But we should consider adding a background to the icon no matter how.

provokateurin:
A problem I see with flutter_launcher_icons is that it only allows to generate a single image so we can't use it to generate the launcher icons for the individual apps. I wanted to move some of the platform specific stuff into the apps (like permissions), maybe this is the moment to attempt that.

PR #452, opened 2023-07-07
Refactor/login flow

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/343

Implements the blue flow from https://github.com/provokateurin/nextcloud-neon/blob/main/docs/login.md and makes it possible to easily implement QR login later.

The UI is not perfect yet, but I think it works pretty well. Let me know how you would improve it (especially for larger screens).


packages/neon/neon/lib/src/pages/login_check_server_status.dart
```diff
@@ -0,0 +1,100 @@
...
+        ? (ValidationState.loading, AppLocalizations.of(context).loginCheckingServerVersion)
+        : result.requireData.isSupported
+            ? (
+                ValidationState.success,
+                AppLocalizations.of(context).loginSupportedServerVersion(result.requireData.versionstring),
+              )
+            : (
+                ValidationState.failure,
+                AppLocalizations.of(context).loginUnsupportedServerVersion(result.requireData.versionstring),
+              );
```
Leptopoda, line 82: This looks a bit funny :)

It's probably worth directly returning the ValidationTiles in normal if clauses (I hate the readability of nested `?` conditions).

packages/neon/neon/lib/src/pages/login_check_server_status.dart
```diff
@@ -0,0 +1,100 @@
...
+    return NeonValidationTile(
+      title: title,
+      state: state,
+    );
+  }
+
+  Widget _buildMaintenanceModeTile(final Result<CoreServerStatus> result) {
```
Leptopoda, line 89: See above

packages/neon/neon/lib/src/pages/login_flow.dart
```diff
@@ -0,0 +1,82 @@
...
+  });
+
+  final String serverURL;
+
+  @override
+  State<LoginFlowPage> createState() => _LoginFlowPageState();
+}
+
+class _LoginFlowPageState extends State<LoginFlowPage> {
+  late final LoginFlowBloc bloc;
```
Leptopoda, line 25: disposal?

packages/neon/neon/lib/src/pages/login_flow.dart
```diff
@@ -0,0 +1,82 @@
...
+                  const SizedBox(
+                    height: 10,
+                  ),
+                  ElevatedButton(
+                    onPressed: bloc.refresh,
+                    child: Text(AppLocalizations.of(context).loginOpenAgain),
+                  ),
+                ],
+              ],
+            ),
```
Leptopoda, line 78: It's probably worth directly returning the widgets and only returning the column on success.

Leptopoda (commented):
I can't tell you how much I love this.
I really like the flow where we validate the compatibility and the account tile.
I'm sorry for the picky review but the code is already sooo much cleaner than it was before so let's go the 110% :)

Regarding the design I'd suggest to clamp the max width with a ConstrainedBox and then maybe some centering.
According to m3 BottomSheet(s) should be clamped to `640dp`(https://m3.material.io/components/bottom-sheets/specs) which might also be the same as the tablet breakpoint in an adaptieView.

If this is still to large just play with the value xD

packages/neon/neon/lib/src/pages/login.dart
```diff
@@ -128,129 +57,70 @@ class _LoginPageState extends State<LoginPage> {
...
-                                      ),
-                                    ],
-                                  ],
-                                ),
+          child: Scaffold(
+            resizeToAvoidBottomInset: true,
+            appBar: AppBar(
+              leading: accountsSnapshot.data?.isNotEmpty ?? false
+                  ? BackButton(
+                      onPressed: () {
```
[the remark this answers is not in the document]
Leptopoda: Oh and maybe it's worth not using a back Icon but rather use an IconButton with an `X` to show that this cancels the login rather than going back a page in the loginflow.

There should be a widget called `CloseButton` but I'd have to check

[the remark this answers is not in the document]
provokateurin: The whole manual back button and listener can actually be removed. With the new routing system it works out of the box :)

packages/neon/neon/lib/src/pages/login_check_server_status.dart
provokateurin: Oh yeah, that makes so much more sense!

packages/neon/neon/lib/src/pages/login_flow.dart
provokateurin: We can have a loading and success or error state at the same time, so we need a column. I don't know how you would do it differently?

packages/neon/neon/lib/src/widgets/validation_tile.dart
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  final String title;
+  final ValidationState state;
+
+  @override
+  Widget build(final BuildContext context) {
+    const size = 32.0;
+
+    late Widget leading;
```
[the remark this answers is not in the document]
provokateurin: I think the assign switch looks even nicer

packages/neon/neon/lib/src/blocs/login_check_account.dart
```diff
@@ -0,0 +1,71 @@
...
+  @override
+  Future refresh() async {
+    state.add(Result.loading());
+
+    try {
+      final client = NextcloudClient(
+        serverURL,
+        loginName: loginName,
+        password: password,
+        userAgentOverride: userAgent(_packageInfo),
```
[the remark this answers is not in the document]
provokateurin: Yeah true, it's also quite annoying having to pass the package info everywhere.

packages/neon/neon/lib/src/blocs/login_flow.dart
```diff
@@ -0,0 +1,82 @@
...
+      _pollTimer = Timer.periodic(const Duration(seconds: 1), (final _) async {
+        try {
+          final resultResponse = await _client.core.getLoginFlowResult(token: initResponse.poll.token);
+          _cancelPollTimer();
+          _resultController.add(resultResponse);
+        } catch (e, s) {
+          debugPrint(e.toString());
+          debugPrint(s.toString());
+        }
+      });
```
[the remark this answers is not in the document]
Leptopoda: My bad :)

packages/neon/neon/lib/src/pages/login_flow.dart
Leptopoda: ah yes :woman_facepalming:

Leptopoda (approved):
I didn't re test and only tested on linux (and only adding another account) not the initial login.

Looks really good. using a CloseButton and constraining the width can be done in followups as design changes and not logic changes.

provokateurin:
I implemented the CloseButton and it definitely looks better and makes more sense. I also added the ConstrainedBox and 640 works pretty well.

packages/neon/neon/lib/src/pages/login_check_account.dart
```diff
@@ -0,0 +1,112 @@
...
+  @override
+  void dispose() {
+    bloc.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(final BuildContext context) => Scaffold(
+        appBar: AppBar(
+          leading: const CloseButton(),
```
Leptopoda, line 54: I might have wrongly communicated this. Sorry :)

I thought we only need the close button for the `LoginPage` (the initial page). As using it there will cancel the login. While using it here just goes back one step in the login flow (thus a backbutton might be prefered).

packages/neon/neon/lib/src/pages/login.dart
Leptopoda, line 1: Can we also constrain this page?

packages/neon/neon/lib/src/pages/login_check_server_status.dart
```diff
@@ -0,0 +1,123 @@
...
+  @override
+  void dispose() {
+    bloc.dispose();
+    super.dispose();
+  }
+
+  @override
+  Widget build(final BuildContext context) => Scaffold(
+        appBar: AppBar(
+          leading: const CloseButton(),
```
Leptopoda, line 42: same here.

packages/neon/neon/lib/src/pages/login_check_account.dart
```diff
@@ -0,0 +1,112 @@
...
+        ),
+        body: Center(
+          child: Padding(
+            padding: const EdgeInsets.all(10),
+            child: ConstrainedBox(
+              constraints: const BoxConstraints(
+                maxWidth: 640,
+              ),
```
Leptopoda, line 62: Something for the new NeonTheme?

packages/neon/neon/lib/src/pages/login_check_server_status.dart
```diff
@@ -0,0 +1,123 @@
...
+        ),
+        body: Center(
+          child: Padding(
+            padding: const EdgeInsets.all(10),
+            child: ConstrainedBox(
+              constraints: const BoxConstraints(
+                maxWidth: 640,
+              ),
```
Leptopoda, line 50: also theme?

packages/neon/neon/lib/neon.dart
```diff
@@ -17,6 +17,8 @@ import 'package:package_info_plus/package_info_plus.dart';
 import 'package:provider/provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
+late final String neonUserAgent;
```
Leptopoda, line 20: can we make this private/hidden?

I don't it needs to be part of our api.

packages/neon/neon/lib/src/pages/login_check_account.dart
provokateurin: No these pages also completely cancel the login, so this is correct

packages/neon/neon/lib/neon.dart
provokateurin: You mean annotating with `@internal`? Because making it private wouldn't work as we still need to import it throughout the app

packages/neon/neon/lib/src/pages/login_check_account.dart
provokateurin: What's the best way to add it there?

packages/neon/neon/lib/neon.dart
Leptopoda: Ah yeah I didn't think about this.

Yes in this case `@internal` might make the most sense

packages/neon/neon/lib/src/pages/login_check_account.dart
Leptopoda: I haven't seen full constraints in ThemeData yet and it looks like flutter would rather just store the value (`double 640`) but imo storing the entire constraint would also be ok.

---

For now it might be ok just to add it as a field in the NeonTheme.
You'll need to both add it to the copyWith  and lerp function so flutter will automatically animate transitions.
You can read more at: https://api.flutter.dev/flutter/material/ThemeExtension-class.html
It could look like:
[patch.txt](https://github.com/provokateurin/nextcloud-neon/files/11995166/patch.txt)

---

I think eventually we should organize our fields in the NeonTheme in sub classes (similar to what the flutter ThemeData does with TextTheme, ButtonTheme ...) but we can also change this once we start adding more fields to it.

Leptopoda: Ah and you can access it through `Theme.of(context).extension<NeonTheme>()!` everywhere where we have the context.

provokateurin: Thanks a lot for your explanation! I already read a bit into ThemeExtensions, but I just wanted to make sure that I do it correctly and don't waist time implementing it the wrong way

Leptopoda: And please add doc comments to the property in the NeonTheme.

Ideally the NeonTheme should be exported eventually so other apps can make use of it.

packages/neon/neon/lib/src/pages/login_check_account.dart
provokateurin: You were right, my bad

Leptopoda (approved):
:tada:

PR #454, opened 2023-07-08
neon: make Option listenable

Leptopoda:
Why should we care?
- gone is all the async code to get the value of an option
- gone are StreamBuilders with hasData and RequireData (OptionBuilder already handled that one)
- we have a nice interface for making an option dependant on another one (sadly only if they are of type bool)
- gone are the days of seeded or not seeded values
- events are only fired when the value really changes avoiding rebuilds
- ValueListenableBuilder takes an optional child argument so we can prevent rebuilding the subtree if it doesn't depend on it (similar to what we where doing manually

---

Unrelated:
I also disposed everything and remove listeners where needed.

---
Performance?
There shouldn't be any significant difference when compiled for release. This is mainly a code style question.

---

This is just an idea so if you don't like it don't merge it. I was just having fun with listenables :)

EDIT: And sorry for the single commit


packages/neon/neon/lib/src/blocs/next_push.dart
```diff
@@ -27,19 +27,15 @@ class NextPushBloc extends Bloc implements NextPushBlocEvents, NextPushBlocState
...
       _globalOptions.pushNotificationsEnabled.stream,
       _globalOptions.pushNotificationsDistributor.stream,
-      _globalOptions.pushNotificationsDistributor.values,
       _accountsBloc.accounts,
     ]).debounceTime(const Duration(milliseconds: 100)).listen((final _) async {
-      if (!_globalOptions.pushNotificationsEnabled.enabled.hasValue ||
-          !_globalOptions.pushNotificationsEnabled.enabled.value ||
-          !_globalOptions.pushNotificationsEnabled.hasValue ||
-          !_globalOptions.pushNotificationsEnabled.value) {
+      if (!_globalOptions.pushNotificationsEnabled.enabled || !_globalOptions.pushNotificationsEnabled.value) {
```
provokateurin, line 32: I remember this was a lot of pain, needs to be thoroughly tested

packages/neon/neon/lib/src/blocs/push_notifications.dart
```diff
@@ -54,10 +60,12 @@ class PushNotificationsBloc extends Bloc implements PushNotificationsBlocEvents,
   bool? _pushNotificationsEnabled;
 
   final _notificationsController = StreamController<PushNotification>();
+  late final StreamSubscription _accountsListener;
```
provokateurin, line 63: Needs disposable?

packages/neon/neon/lib/src/blocs/push_notifications.dart
```diff
@@ -32,16 +32,22 @@ class PushNotificationsBloc extends Bloc implements PushNotificationsBlocEvents,
...
+    final enabled = _globalOptions.pushNotificationsEnabled.value;
+    if (enabled != _pushNotificationsEnabled) {
+      _pushNotificationsEnabled = enabled;
+      if (enabled) {
+        // We just use a single RSA keypair for all accounts
+        _keypair = await PushUtils.loadRSAKeypair(_storage);
+        await _setupUnifiedPush();
+      } else {
+        _globalOptions.pushNotificationsDistributor.removeListener(_distributorListener);
+        unawaited(_accountsListener.cancel());
```
provokateurin, line 49: You cancel the listeners here, but what happens if the user enables push notifications again? As far as I can tell the listeners are only created in initState

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -150,17 +149,17 @@ class _HomePageState extends State<HomePage> {
...
+      builder: (final context, final navigationMode, final child) {
         final drawerAlwaysVisible = navigationMode == global_options.NavigationMode.drawerAlwaysVisible;
 
         final body = Scaffold(
           key: _scaffoldKey,
           resizeToAvoidBottomInset: false,
           drawer: !drawerAlwaysVisible ? drawer : null,
           appBar: appBar,
-          body: appView,
+          body: child,
```
provokateurin, line 162: What happened to the appView variable? I don't see that you renamed it or removed it?

packages/neon/neon/lib/src/pages/home.dart
```diff
@@ -176,6 +175,7 @@ class _HomePageState extends State<HomePage> {
 
         return body;
       },
+      child: appView,
```
provokateurin, line 178: Why is this needed/why was it not done before?

packages/neon/neon/lib/src/utils/global_options.dart
```diff
@@ -23,54 +21,43 @@ class GlobalOptions {
...
       }
-    });
+    } else {
+      pushNotificationsDistributor.value = null;
+    }
+  }
+
+  void _themeModeListner() {
+    themeOLEDAsDark.enabled = themeMode.value != ThemeMode.light;
   }
```
provokateurin, line 56: I think removing it is better. The behavior is quite inconsistent because it was still enabled for light theme regardless of the resulting theme

provokateurin (changes requested):
I haven't tested it, but I like the idea a lot! I saw some oversights that need some fixing. I'll test the PR once those are fixed

packages/neon/neon/lib/src/pages/home.dart
Leptopoda: The child parameter let's you cache the subtree but we alreeady cache it manaually in the appView variable.

There is no difference to before and I can see how the original way is easier to understand.

packages/neon/neon/lib/src/pages/home.dart
Leptopoda: the child parameter get's specified below (other comment)

packages/neon/neon/lib/src/blocs/push_notifications.dart
Leptopoda: I haven't tested this (as you mentioned some thorough testing might be needed).
If enabled `_setupUnifiedPush` will be called two lines above wich will register the listener.

I'll have a look we can clean that up/add comments in the code to make that clear.

packages/neon/neon/lib/src/utils/global_options.dart
Leptopoda: I mean that's sorta intended.
I might want to switch back to the dark theme and still keep the oled pref.

We could use a tristate CheckBox in this case and display the undetermined state.
Or only render it checked if enabled (both something we need to change in `CheckBoxSettingsTile`).
Maybe with an option?

What do you prefer?


---

For reference the tristate looks like (just imagine it greyed out if it's also disabled):
unnamed

provokateurin: Ah sorry there was a typo in the comment. I think we can just remove the logic to disable the checkbox if the theme is the light theme. Just make it possible to always change this setting regardless of the selected theme

Leptopoda: but having a setting that doesn't change anything is a bit bad isn't it?

provokateurin: Why wouldn't it change anything? I only want to have the enablement logic removed, not the option or anything else

packages/neon/neon/lib/src/blocs/push_notifications.dart
Leptopoda: I think I found a nice way to clean it up a bit.

I also removed some `cached` variables as it's now possible to synchronously access the real value removing the possibility of potentially outdated values.

provokateurin (changes requested):

packages/neon/neon/lib/src/settings/models/option.dart
```diff
@@ -1,63 +1,114 @@
...
+    final bool enabled = true,
     this.category,
-    final BehaviorSubject<bool>? enabled,
-  }) : enabled = enabled ?? BehaviorSubject<bool>.seeded(true);
+    final T? initialValue,
+  })  : _value = initialValue ?? defaultValue,
+        _enabled = enabled;
+
+  /// Creates an Option depending on the State of another one.
+  Option.depend({
```
provokateurin, line 25: I know I suggested this name, but somehow I still don't like it :see_no_evil: I don't have something better though

provokateurin (approved):

packages/neon/neon/lib/src/settings/models/option.dart
Leptopoda: Maybe rely?

But I not too convinded.

provokateurin: maybe dependent? we can just leave it and fix it later if needed

provokateurin:
Ah I haven't tested this yet, especially the push notifications. Let me do that first and then it's good to go

packages/neon/neon/lib/src/blocs/next_push.dart
provokateurin: Before enabling push notifications it showed me a warning that no UnifiedPush distributor was detected. After enabling it the warning was gone and the distributors were also present in the list.

provokateurin: The warning also appears again when disabling the push notifications. Other than that everything seems to work as expected!

provokateurin:
Alright, it only needs a small fix for the push notifications, then it's ready :)

packages/neon/neon/pubspec.yaml
```diff
@@ -28,6 +28,7 @@ dependencies:
   json_annotation: ^4.8.1
   material_design_icons_flutter: ^7.0.7296
   meta: ^1.9.1
+  mocktail: ^0.3.0
```
Leptopoda, line 31: should be a dev dependency

provokateurin:
@Leptopoda can you rebase and fix the remaining problems?

Leptopoda:
Done.

PR #456, opened 2023-07-10
Feature/account switcher

Leptopoda:
just picking the parts of #441 that are ready.

I'm not entirely sure whether 0c6d49a3fc2eea8a22d2072cba417ab613c6bb5b makes sense.
regarding 919e53264baa976c1cd48fb1f0a2deac31fbc61f We could also move the initial account settings into the manage settings page.

### DIFF
--- a/packages/neon/neon/lib/src/pages/account_settings.dart
+++ b/packages/neon/neon/lib/src/pages/account_settings.dart
@@ -10,13 +10,14 @@ import 'package:neon/src/settings/widgets/custom_settings_tile.dart';
 import 'package:neon/src/settings/widgets/dropdown_button_settings_tile.dart';
 import 'package:neon/src/settings/widgets/settings_category.dart';
 import 'package:neon/src/settings/widgets/settings_list.dart';
+import 'package:neon/src/theme/dialog.dart';
 import 'package:neon/src/utils/confirmation_dialog.dart';
 import 'package:neon/src/widgets/exception.dart';
 import 'package:neon/src/widgets/linear_progress_indicator.dart';
 import 'package:nextcloud/nextcloud.dart';
 
 class AccountSettingsPage extends StatelessWidget {
-  AccountSettingsPage({
+  const AccountSettingsPage({
     required this.bloc,
     required this.account,
     super.key,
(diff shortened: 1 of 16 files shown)


packages/neon/neon/lib/src/widgets/app_bar.dart
```diff
@@ -77,19 +78,103 @@ class NeonAppBar extends StatelessWidget implements PreferredSizeWidget {
...
+            AccountSwitcherButton(),
+          ],
+        ),
+      ),
+    );
+  }
+}
+
+@internal
+class AccountSwitcherButton extends StatelessWidget {
```
provokateurin, line 92: Please move it into a separate file

packages/neon/neon/lib/src/widgets/account_tile.dart
```diff
@@ -47,54 +99,12 @@ class NeonAccountTile extends StatelessWidget {
...
-        ),
-      ),
-      subtitle: Text(
-        account.client.humanReadableID,
-        style: Theme.of(context).textTheme.bodySmall!.copyWith(
-              color: textColor,
-            ),
-        overflow: TextOverflow.ellipsis,
+        showStatus: showStatus && !showActive,
+        decoration: showActive && isActive ? const Icon(Icons.check_circle) : null,
```
provokateurin, line 103: This check mark is quite huge. Please reduce the size to size of the status circle

packages/neon/neon/lib/src/widgets/app_bar.dart
```diff
@@ -77,19 +78,101 @@ class NeonAppBar extends StatelessWidget implements PreferredSizeWidget {
...
+                child: ListBody(children: accountList.toList()),
               ),
+            ListTile(
+              leading: const Icon(Icons.person_add),
+              title: Text(AppLocalizations.of(context).globalOptionsAccountsAdd),
+              onTap: () {
+                Navigator.pop(context);
+                const AddAccountRoute().push(context);
+              },
             ),
```
provokateurin, line 142: This is duplicate with the new manage accounts page. I would remove it from here since it's just one more click away and you don't add accounts that often and it is quite clear where you can do that when the dialog is displayed.

packages/neon/neon/lib/src/pages/account_settings.dart
```diff
@@ -11,12 +11,94 @@ import 'package:neon/src/settings/widgets/dropdown_button_settings_tile.dart';
...
 import 'package:neon/src/widgets/linear_progress_indicator.dart';
 import 'package:nextcloud/nextcloud.dart';
+import 'package:provider/provider.dart';
+
+enum AccountMangeAction {
+  setActive,
+  delete,
+}
+
+class AccountManagePage extends StatelessWidget {
```
provokateurin, line 26: Separate file please

packages/neon/neon/lib/src/pages/account_settings.dart
```diff
@@ -11,12 +11,94 @@ import 'package:neon/src/settings/widgets/dropdown_button_settings_tile.dart';
...
+
+    if (index < accounts.length) {
+      final account = accounts[index];
+
+      final trailing = PopupMenuButton<AccountMangeAction>(
+        itemBuilder: (final context) => [
+          if (activeAccount != account)
+            PopupMenuItem(
+              value: AccountMangeAction.setActive,
+              child: Text(AppLocalizations.of(context).settingsAccountManage),
```
provokateurin, line 64: Wrong text

provokateurin (changes requested):
I feel like there is a bit duplicate logic here between the account switcher and the manage accounts page.
I would remove the popup menu from the manage accounts page and also not show which account is currently active (I also like displaying the status there). For the actions behind the popup menu you still have the account switcher.
Removing an account is also possible from the individual account settings page, so it doesn't need to be on the manage accounts page.

In my opinion we should just provide one way for users to do something, otherwise they can get confused. Duplicate logic also makes it harder to maintain in the future.

Leptopoda:
As mentioned in the other PR I blindly copied the NC design but I forgot to think about our AccountSettingsPage already providing most of the functionality 🫣

So you mean the manage account screen should only be a list of accounts (showing the status instead of the active one)?
But then even this is redundant isn't it?
In that case I'd rather keep them in the normal settings and have the account switcher as a pure switcher (with the ability to add an account).

provokateurin:
>  In that case I'd rather keep them in the normal settings and have the account switcher as a pure switcher (with the ability to add an account).

Yeah that's better. Then we still have two ways to add an account, but I think that is fine (although I would be in favor of removing that from the switcher).

Leptopoda:
How do you like this?

packages/neon/neon/lib/src/widgets/account_switcher.dart
```diff
@@ -0,0 +1,99 @@
...
+                },
+              ),
+            ListTile(
+              leading: const Icon(Icons.person_add),
+              title: Text(AppLocalizations.of(context).globalOptionsAccountsAdd),
+              onTap: () {
+                Navigator.of(context).pop();
+                const AddAccountRoute().push(context);
+              },
+            ),
```
provokateurin, line 61: I'm still not happy with this. I'd remove it since it's duplicate functionality

packages/neon/neon/lib/src/widgets/account_switcher.dart
```diff
@@ -0,0 +1,99 @@
...
+            ListTile(
+              leading: const Icon(Icons.settings),
+              title: Text(AppLocalizations.of(context).settingsAccountManage),
+              onTap: () {
+                Navigator.of(context).pop();
+                const SettingsRoute(category: 'Accounts').push(context);
```
provokateurin, line 67: Externalize these keys into constants

packages/neon/neon/lib/src/pages/settings.dart
```diff
@@ -29,9 +30,12 @@ import 'package:provider/provider.dart';
 
 class SettingsPage extends StatefulWidget {
   const SettingsPage({
+    this.category,
     super.key,
   });
 
+  final String? category;
```
provokateurin, line 37: Better to also name this initialCategory, otherwise the name is not really descriptive

packages/neon/neon/lib/src/router.dart
```diff
@@ -198,8 +198,11 @@ class NextcloudAppSettingsRoute extends GoRouteData {
 
 @immutable
 class SettingsRoute extends GoRouteData {
-  const SettingsRoute();
+  const SettingsRoute({this.category});
+
+  /// The initial category to show.
+  final String? category;
```
provokateurin, line 204: Also initialCategory

provokateurin (changes requested):

provokateurin (approved):

PR #463, opened 2023-07-13
dynamite: Correctly await responses where necessary

provokateurin:


packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -625,20 +626,24 @@ class OpenAPIBuilder implements Builder {
                                     if (mimeType == '*/*' || mimeType.startsWith('image/')) {
                                       dataType = 'Uint8List';
                                       dataValue = 'response.bodyBytes';
```
Leptopoda, line 628: Can't we add the await to the dataValue directly?

Similarly as it's already done in the cases where we don't need to explicitly await it.

provokateurin: Ahhhh this was for a completely different reason. If you do your suggestion then you end up generating `return await response.bodyBytes;` which is a linting error and `dart fix` doesn't fix it for some reason. That's why the switch is needed to only add the await in specific places where it is actually needed. Does that make sense?

Leptopoda (approved):
I somehow have the feeling that this can be done cleaner :thinking: 

But let's get it merged so we can move on with more important stuff; dynamite isn't run in production after all :)

Issue #417, opened 2023-06-26
Use automatic OpenAPI specs


provokateurin:
WIP: https://github.com/provokateurin/nextcloud-neon/tree/feature/openapi-automatic

Issue #1, opened 2022-07-06
Implement all API endpoints of dart-nextcloud


provokateurin:
See https://github.com/jld3103/dart-nextcloud/tree/master/lib/src. There will definitely be breaking changes, but we need all endpoints so it's possible to migrate at all.  

- [x] core: autocomplete user (https://github.com/jld3103/dart-nextcloud/blob/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/autocomplete/client.dart#L19)
- [x] core: avatar (https://github.com/jld3103/dart-nextcloud/blob/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/avatar/client.dart#L18)
- [ ] core: preview by file id (https://github.com/jld3103/dart-nextcloud/blob/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/preview/client.dart#L59)
- [ ] core: thumbnail (https://github.com/jld3103/dart-nextcloud/blob/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/preview/client.dart#L145)
- [ ] files_sharing: sharees (https://github.com/jld3103/dart-nextcloud/blob/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/sharees/client.dart)
- [ ] talk (https://github.com/jld3103/dart-nextcloud/tree/0cc8e1df7ef8b4b4407d4f8c5271453fd6149e24/lib/src/talk)

sonigeez:
what's update on talk APIs?

provokateurin:
They are work in progress: https://github.com/provokateurin/nextcloud-neon/blob/feature/openapi-demo/specs/spreed.json
I assume they will be available in about two months, but no promises given ;)

provokateurin:
@Leptopoda this should be promoted to the 1.0 release milestone as we want to publish the package then and it would be bad if we wouldn't have all APIs implemented in the new version, right? Not sure if Talk will make the cut though, that could take some more time. It doesn't really matter though because the current package only implements an old version of the Talk API which we are never going to implement anyway.

provokateurin:
The missing APIs would all be ticked off once we have https://github.com/provokateurin/nextcloud-neon/issues/417

provokateurin:
[...]

Issue #420, opened 2023-06-26
Dynamite inherritence (AllOf)


Leptopoda:
Might be needed for #417 

This should be fairly trivial and a outdated POC can be found in:
https://github.com/Leptopoda/nextcloud-neon/commit/045aa94bc8bdaedb5205bbd40ca88eedd11bcf16

PR #465, opened 2023-07-13
Feature/openapi automatic

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/1
Closes https://github.com/provokateurin/nextcloud-neon/issues/417
Depends on https://github.com/provokateurin/nextcloud-neon/pull/462 https://github.com/provokateurin/nextcloud-neon/pull/463 https://github.com/provokateurin/nextcloud-neon/pull/464 and https://github.com/nextcloud/server/pull/39223


Leptopoda:
Looks ok at the first glance.
I'll thoroughly review it in the coming days.

Leptopoda:
I might also rebase/redo my changes for #420 so we can land the automatic branch without any big changes afterwards (might be better for devs already using this lib to adjust only once).

provokateurin:
Sounds look a good plan, let's do it 👍

Leptopoda:
The ContentString serialization is fixed with #501 but the same test now fails with a 500 server error.
Let me know if this is also related to built_value

provokateurin:
I'll look into the test failure once your PR is merged

provokateurin:
I assume the problem is that we don't properly deserialize the parameters. OpenAPI can describe many ways to serialize values (https://swagger.io/docs/specification/serialization/) but we just assume something and then spec also doesn't say what it actually needs.

provokateurin:
Yep, server log confirms it: `Argument #5 ($shareTypes) must be of type array, string given`.

Leptopoda:
> The autocomplete test is still failing because we are not properly serializing the parameter.

But this isn't a problem with dynamite? 

> Yep, server log confirms it: Argument #5 ($shareTypes) must be of type array, string given.

The spec says it's a string (ContentString to be exact). After changing this the AutocompleteResult.status appears to also be wrong.

https://github.com/provokateurin/nextcloud-neon/pull/465/files#diff-24dc2feb72ef0ff5bc5c6e6b4fe8f5efd9693883efc2893bd0d1afd9feea73a3R1576-R1592

provokateurin:
I fixed the specs and when https://github.com/provokateurin/nextcloud-neon/pull/522 is merged it will work correctly. The autocomplete tests is still broken because of a bug that was I fixed but that is only in 28+.

Leptopoda:
We can still skip it until we switch to 28

provokateurin:
Shall we wait for the #420 work before merging this? The result should be a lot cleaner.

packages/nextcloud/test/provisioning_api_test.dart
```diff
@@ -14,24 +14,42 @@ void main() {
...
+      test('Get apps', () async {
+        final apps = await client.provisioningApi.apps.getApps();
+        expect(apps.ocs.data.apps, hasLength(39));
+      });
+
+      test('Get app infos', () async {
+        for (final app in (await client.provisioningApi.apps.getApps()).ocs.data.apps) {
+          await client.provisioningApi.apps.getAppInfo(app: app);
+        }
+      });
```
Leptopoda, line 52: I'm not a fan of tests without any expect (I can see why it's done though)

But I really dislike `(await client.provisioningApi.apps.getApps()).ocs.data.apps`
The result of getApps should be in a separate variable raising thje question why this isn't merged with the test above it.

packages/nextcloud/test/core_test.dart
```diff
@@ -69,8 +69,10 @@ void main() {
       expect(navigationApps.ocs.data[5].id, 'news');
     });
 
+    /*
+    TODO: This test only works on 28+ due to a bug fix with the status
```
Leptopoda, line 73: I'd like to leave the test implemented and just skip it.

I hate commented tests as it's harder to see them from the UI of the IDE and I tend to forget that we even have it

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -41,6 +41,7 @@ class Account implements Credentials {
         );
 
   factory Account.fromJson(final Map<String, dynamic> json) => _$AccountFromJson(json);
+
```
Leptopoda, line 44: Intened?

packages/nextcloud/lib/src/helpers/news.dart
```diff
@@ -0,0 +1,16 @@
...
+  folder(1),
+  starred(2),
+  allItems(3),
+  shared(4),
+  explore(5),
+  unread(6);
+
+  const NewsListType(this.code);
+
+  final int code;
```
Leptopoda, line 15: Is the code even needed?

`NewsListType.feed.index` should also yield the same (yes this requires correct ordering).

packages/nextcloud/lib/src/helpers/core.dart
```diff
@@ -0,0 +1,22 @@
+// ignore_for_file: public_member_api_docs
+
+enum ShareType {
+  user(0),
+  group(1),
+  usergroup(2),
+  link(3),
+  email(4),
+  // 5 was contact, is no longer used
```
Leptopoda, line 9: I'd rather have the deprecated ones be implemented with a `deprecated` annotation.

That way `ShareType.values[index]` will always match the requested index.

packages/nextcloud/lib/src/client.dart
```diff
@@ -19,23 +21,31 @@ class NextcloudClient extends DynamiteClient {
...
     final String? password,
+    final String? appPassword,
     final String? language,
     final AppType appType = AppType.unknown,
     final String? userAgentOverride,
     super.cookieJar,
-  })  : assert(loginName != null || username == null, 'Provide loginName instead of username or both'),
+  })  : assert(
+          ((loginName != null) == (username != null)) || (loginName != null || username == null),
+          'Provide loginName instead of username or both',
```
Leptopoda, line 31: So `loginName` is always needed and `username` is optional?

In this case we should make the first one non nullable.

Leptopoda (commented):
Just reviewing the not generated code :sweat_smile: 

I'll still need to test it but it looks quite good.

packages/nextcloud/test/core_test.dart
provokateurin: Ah I forgot that you can skip tests

packages/nextcloud/lib/src/client.dart
provokateurin: The username variable was used for the WebDAV endpoint but that is no longer the case. It can just be removed. loginName still has to stay nullable because you can create clients without credentials.

provokateurin: Fixed it in https://github.com/provokateurin/nextcloud-neon/pull/526

Leptopoda:
I don't see any changes from my last review.
What should I review?

provokateurin:
Just approve :) Will push my rebase now

provokateurin:
Ah I forgot that I haven't pushed my changes yet :upside_down_face:

provokateurin:
@Leptopoda now you can review

Leptopoda (approved):
:tada:

provokateurin:
Actually to also close https://github.com/provokateurin/nextcloud-neon/issues/1 we need files_sharing

provokateurin:
Woop woop! A dream came true! :heart:

Issue #358, opened 2023-06-03
Use Conventional Commits


provokateurin:
https://www.conventionalcommits.org/en/v1.0.0/

This will also help melos to create changelogs.
A pre-commit git hook and CI (https://github.com/webiny/action-conventional-commits) should be added to ensure mistakes are caught.

provokateurin:
We could use https://pub.dev/packages/commitlint_cli (both locally and in CI; I don't like using two different tools to achieve the same and possible introducing differences).

provokateurin:
@Leptopoda I think this is important enough to bump it to the milestone. It will affect the development process so it's easier to do before we release a first version

PR #468, opened 2023-07-14
Use conventional commits

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/358

### DIFF
--- a/tool/setup.sh
+++ b/tool/setup.sh
@@ -6,3 +6,4 @@ dart pub global activate melos 3.1.0
 dart pub global activate fvm 2.4.1
 echo "y" | fvm install
 melos exec --concurrency=1 flutter pub get
+dart run husky install
--- a/.github/workflows/conventional_commits.yaml
+++ b/.github/workflows/conventional_commits.yaml
@@ -1,0 +1,22 @@
+name: Conventional commits
+on:
+  pull_request:
+
+jobs:
+  conventional-commits:
+    name: Conventional commits
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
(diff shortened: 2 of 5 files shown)


provokateurin:
For some reason the CI passes but if I execute the exact same command locally it fails as it should :thinking:

tool/setup.sh
```diff
@@ -6,3 +6,4 @@ dart pub global activate melos 3.1.0
 dart pub global activate fvm 2.4.1
 echo "y" | fvm install
 melos exec --concurrency=1 flutter pub get
+dart run husky install
```
Leptopoda, line 9: does it make sense to use `fvm dart ...`?

Leptopoda (approved):
We are now using a fairly complex stack (fvm, melos, husky ...)

We should add a `CONTRIBUTING.MD` either with this or as a followup so new devs know how to get started

tool/setup.sh
provokateurin: I don't think it matters since it's only for the workspace package and that has a very loose constraint. Also  we are using the direct dart command earlier in the script so it wouldn't fail here

provokateurin:
I think the CI might not be working because GH actions is doing a sparse checkout with only the branch. I guess that's a bug in commitlint_cli as well

provokateurin:
Meh I still can't get the CI to work. Using https://github.com/nektos/act locally also doesn't fail properly, so I will investigate that later

provokateurin:
Wow I can not get it working at all. I was able to get it working with act by using a quirk, but that doesn't work here.

provokateurin:
I opened https://github.com/hyiso/commitlint/issues/7, hopefully someone can help us :/

provokateurin:
Yay CI is failing correctly!

Leptopoda (approved):

Issue #460, opened 2023-07-12
Configure concurrency for release workflow


provokateurin:
Prevents to releases potentially race conditioning each other: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

PR #474, opened 2023-07-15
ci: Fix publish workflow concurrency

provokateurin:
Fixes https://github.com/provokateurin/nextcloud-neon/issues/460

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -5,6 +5,10 @@ on:
     branches:
       - main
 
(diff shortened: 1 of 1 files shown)


Leptopoda (approved):
Wouldn't it make sense to just run it on a cron job?

Our 'nightly' runs on every PR and I usually know them as once a day (would also save us some CI time)

provokateurin:
Yeah that would make sense. Can you create an issue for that?

PR #486, opened 2023-07-15
Fix/web server

provokateurin:
>  The web server runs only one single-threaded process, so PHP applications will stall if a request is blocked. 

https://www.php.net/manual/en/features.commandline.webserver.php

This makes it impossible to use with Talk. The dev container uses apache by default and tests can opt-in to use apache, but the built-in php web server is still used by default because it is a lot faster to start up.

### DIFF
--- a/packages/nextcloud/test/news_test.dart
+++ b/packages/nextcloud/test/news_test.dart
@@ -25,12 +25,12 @@ Future run(final DockerImage image) async {
     tearDown(() => container.destroy());
 
     Future<NewsListFeeds> addWikipediaFeed([final int? folderID]) => client.news.addFeed(
-          url: 'http://localhost:${rssServer.port}/wikipedia.xml',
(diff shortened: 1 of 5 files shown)


packages/nextcloud/test/news_test.dart
```diff
@@ -22,15 +22,18 @@ Future run(final DockerImage image) async {
       container = await getDockerContainer(image);
       client = await getTestClient(container);
     });
-    tearDown(() => container.destroy());
+    tearDown(() async {
+      print(await container.collectLogs());
```
Leptopoda, line 26: how chatty is this going to be?
Looks like it currently only outputs linebreaks.

Leptopoda (approved):

packages/nextcloud/test/news_test.dart
provokateurin: Ah forgot to remove it again. It would be perfect if the test package would have a way to execute code when a test fails

Issue #470, opened 2023-07-14
Cleanup login UI


provokateurin:
Error handling in the UI is not great and the main login page should roughly have the layout of the Android client.

PR #489, opened 2023-07-16
Cleanup/login

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/470
Addresses problems found in https://github.com/provokateurin/nextcloud-neon/pull/291

### DIFF
--- a/packages/neon/neon/lib/src/pages/login.dart
+++ b/packages/neon/neon/lib/src/pages/login.dart
@@ -56,16 +56,16 @@ class _LoginPageState extends State<LoginPage> {
                     branding.name,
                     style: Theme.of(context).textTheme.titleLarge,
                   ),
-                  const SizedBox(
-                    height: 20,
-                  ),
                   if (branding.showLoginWithNextcloud) ...[
+                    const SizedBox(
+                      height: 20,
+                    ),
                     Text(AppLocalizations.of(context).loginWorksWith),
                     const SizedBox(
                       height: 20,
                     ),
+                    const NextcloudLogo(),
                   ],
-                  const NextcloudLogo(),
                   const SizedBox(
                     height: 40,
                   ),
--- a/packages/neon/neon/lib/src/pages/login.dart
+++ b/packages/neon/neon/lib/src/pages/login.dart
@@ -20,16 +20,21 @@ class LoginPage extends StatefulWidget {
 class _LoginPageState extends State<LoginPage> {
   final _formKey = GlobalKey<FormState>();
   final _focusNode = FocusNode();
-
-  @override
-  void initState() {
-    super.initState();
-  }
+  final _controller = TextEditingController();
 
   @override
   void dispose() {
     _focusNode.dispose();
+    _controller.dispose();
     super.dispose();
+  }
+
+  void login(final String url) {
+    if (_formKey.currentState!.validate()) {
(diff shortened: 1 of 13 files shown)


packages/neon/neon/lib/src/pages/login.dart
```diff
@@ -56,48 +56,27 @@ class _LoginPageState extends State<LoginPage> {
...
                   ],
+                  const SizedBox(
+                    height: 60,
+                  ),
                   Form(
                     key: _formKey,
                     child: TextFormField(
                       focusNode: _focusNode,
-                      decoration: const InputDecoration(
+                      decoration: InputDecoration(
```
Leptopoda, line 76: We should also provide a submit button as not everyone knows the submit functionality of the keyboard (a11y)

Leptopoda (approved):
The UX is better but I still don't like it; I imagined something more like:

[patch.txt](https://github.com/provokateurin/nextcloud-neon/files/12063549/patch.txt)

Leptopoda:
https://github.com/provokateurin/nextcloud-neon/assets/25266387/a165f997-b8a7-4c1d-9890-34bf3bcb49fa

provokateurin:
@Leptopoda I applied your suggestions but didn't apply the patch. Using an animation is a good idea but out of scope for this PR imo. I saw you tried adding better error handling, but I'm a bit confused by that because we already use NeonException which will display the most accurate error message possible :thinking:

packages/neon/neon/lib/src/widgets/nextcloud_logo.dart
```diff
@@ -12,5 +13,6 @@ class NextcloudLogo extends StatelessWidget {
         package: 'neon',
         width: 100,
         height: 100,
+        semanticsLabel: AppLocalizations.of(context).nextcloudLogo,
```
Leptopoda, line 16: I think the normal nextcloud brand name `Nextcloud` would be enough.

My Idea was to have something like "Neon \n works with \n Nextcloud" while we now get "Works with \n Nextcloud logo".

(The neon logo should also get a semantic node in a followup)

Leptopoda:
>  I saw you tried adding better error handling, but I'm a bit confused by that because we already use NeonException which will display the most accurate error message possible thinking

I only did this for design reasons and not for the text we show. The above video shows how I'd like the design (my opinion and I'm not a designer by any means).

- If an error occurs the checks should be grayed out to show they weren't run/completed.
- the retry button should replace the continue one: A user already expects to press this button and not a separate one to pop up somewhere else
- When the failed checks are still shown (why shouldn't they) the "NeonException" should follow the design of the other checks and not be centered.

Does this make sense?

packages/neon/neon/lib/src/widgets/nextcloud_logo.dart
provokateurin: No the logo on the login page is wrapped with extra semantics exactly because of this.

provokateurin:
Now I understand it better. The patch was a bit hard to understand. I'll try to implement your ideas :+1:

provokateurin:
I haven't implemented the size transition, but everything else.

Leptopoda:
I think we could move the login submit button into the server URL field as a suffix icon (similar to what the NC app does).

Leptopoda (approved):

Issue #469, opened 2023-07-14
Create CONTRIBUTING.md


provokateurin:
Explain what tools we use, how they interact with each other and how to set everything up.

For reference: https://github.com/provokateurin/nextcloud-neon/pull/468#pullrequestreview-1529809114

PR #490, opened 2023-07-16
feat(docs): Add contributing docs

provokateurin:
Closes https://github.com/provokateurin/nextcloud-neon/issues/469

@Leptopoda is this everything we need right now? I feel like there could be more explanation. We could add a sentence about the architecture and link to the other doc page.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,0 +1,46 @@
+# Contributing
+
+## Setup
+This project uses an assortment of tools for the development.
+Currently included are:
+- [fvm](https://pub.dev/packages/fvm) for managing Flutter versions
+- [melos](https://pub.dev/packages/melos) for managing packages in this monorepo
+- [husky](https://pub.dev/packages/husky) for managing git hooks
+- [commitlint_cli](https://pub.dev/packages/commitlint_cli) for validating commits according to the conventional commits specification
+
+To set up all these tools run the `./tool/setup.sh` script.
+Note that you need to have Dart installed and [`~/.pub-cache/bin/` needs to be in your PATH](https://dart.dev/tools/pub/cmd/pub-global#running-a-script-from-your-path) before running the script.
+
+## Conventional commits
+We use conventional commits to have meaningful commit messages and be able to generate changelogs.
+A non-breaking feature contribution to `neon_notes` could look like this:
+```bash
+git commit -m "feat(neon_notes): Add a super cool feature."
+```
+You can read the full documentation at https://www.conventionalcommits.org.
+
+## Tools
+We maintain a collection of scripts in `./tool/`.
+They range from setting up a local Nextcloud server (`./tool/dev.sh`) to generating assets.
+
+## Monorepo
+For easier development we use a monorepo structure.
+This means that we have multiple packages in one git repository.
+We use [melos](https://pub.dev/packages/melos) to manage the packages in this repository.
+
+Take a look at our [melos.yaml](../melos.yaml) to find useful commands for running commands like build_runner or the analyzer in all packages.
+
+## Linting
+We use very strict static code analysis (also known as linting) rules.
+This enables us to maintain and verify a consistent code style throughout the repository.
+Please make sure your code passes linting.
+
+You can read more about it on [dart.dev](https://dart.dev/tools/linter-rules).
+
+## Testing
+If you found a bug and are here to fix it, please make sure to also submit a test that validates that the bug is fixed.
+This way we can make sure it will not be introduced again.
+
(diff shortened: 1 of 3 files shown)


docs/contributing.md
```diff
@@ -0,0 +1,12 @@
...
+## Setup
+This project uses an assortment of tools for the development.
+Currently included are:
+- [fvm](https://pub.dev/packages/fvm) for managing Flutter versions
+- [melos](https://pub.dev/packages/melos) for managing packages in this monorepo
+- [husky](https://pub.dev/packages/husky) for managing git hooks
+- [commitlint_cli](https://pub.dev/packages/commitlint_cli) for validating commits according to the conventional commits specification
+
+To set up all these tools run the `./tool/setup.sh` script.
+Note that you need to have Dart installed before running the script.
```
Leptopoda, line 12: pub-cache also needs to be in the path of a user.

docs/contributing.md
Leptopoda, line 1: I'd also add something like:

## Conventional commits:

We make use of conventional commits which allows us to automatically generate changelogs.
A non breaking feature contribution to `neon_notes` could looks like:
```sh
git commit -m "feat(neon_notes): Add a super cool feature."
``` 
You can read the [documentation](https://www.conventionalcommits.org) to get familiar.

## Tools

We maintain a collection of scripts in `./tool`. They range from setting up a local test server (`./tool/dev.sh`) to generating assets for us.

## Monorepo

For easier development we have use monorepo structure. This means that we have multiple packages in one git repo. We use [melos](https://pub.dev/packages/melos) to make working with it easier.
Take a look at our [melos.yaml](melos.yaml) to find other useful commands for running commands like build_runner or the analyzer in all packages at once.

## Linting

We make use of very strict static code analysis (also known as linting). This enables us to maintain and verify a consistent code style throughout  the project. Please make sure your code passes all tests.
You can read more about it on [dart.dev](https://dart.dev/tools/linter-rules).

## Testing

When you found a bug and are here to fix it please make sure to also submit a test that validates the bug is fixed.
This way we can make sure it won't be introduced again.

## Documentation

Whenever you are submitting new features please make sure to also add doc comments.
Please adhere to the [effective-dart](https://dart.dev/effective-dart/documentation) documentation guidlines.

---

This makes it clear to all developers :)
The links in the list under `##Setup` should then link to the individual sections where appropriate.

We can also steel some stuff from https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md like the `Testing` case.

Leptopoda (changes requested):
We should make it as easy as possible for new devs and devs new to the dart/flutter ecosystem.

Leptopoda:
GH also defaults to having it in the root of the repo. Is there any benefit like other tools finding it easier or github promoting it a bit more in the UI :thinking:

docs/contributing.md
provokateurin: But keep the existing section or replace it?

provokateurin:
I know that it's usually in the root, but I don't see any benefits. Also cluttering the root seems bad. When there are many differently named files in the root you end up not finding anything. I can also move it to the root if you think that it is very important.

docs/contributing.md
Leptopoda: I think we can keep the existing as a sort of table of contents.
Especially as we won't be talking about everything like fvm

provokateurin: I mostly copied your texts, but improved and reprashed a few things.

docs/contributing.md
```diff
@@ -0,0 +1,46 @@
...
+## Setup
+This project uses an assortment of tools for the development.
+Currently included are:
+- [fvm](https://pub.dev/packages/fvm) for managing Flutter versions
+- [melos](https://pub.dev/packages/melos) for managing packages in this monorepo
+- [husky](https://pub.dev/packages/husky) for managing git hooks
+- [commitlint_cli](https://pub.dev/packages/commitlint_cli) for validating commits according to the conventional commits specification
+
+To set up all these tools run the `./tool/setup.sh` script.
+Note that you need to have Dart installed and `~/.pub-cache/` needs to be in your PATH before running the script.
```
Leptopoda, line 12: `.pub-cache/bin` as you only want to add packages explicitly activated with `dart pub global activate`

Maybe a link to https://dart.dev/tools/pub/cmd/pub-global#running-a-script-from-your-path might be good as it also describes the craziness of windows (`%LOCALAPPDATA%\Pub\Cache\bin`)

Leptopoda (changes requested):

Leptopoda (approved):

PR #491, opened 2023-07-16
Cleanup/tests

provokateurin:
Lately the tests have been a bit flaky due to network problems. Retry should make the CI more stable

### DIFF
--- a/packages/nextcloud/test/core_test.dart
+++ b/packages/nextcloud/test/core_test.dart
@@ -3,12 +3,11 @@ import 'package:test/test.dart';
 
 import 'helper.dart';
 
-Future main() async {
-  await run(await getDockerImage());
-}
-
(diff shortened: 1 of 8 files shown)


packages/nextcloud/test/core_test.dart
```diff
@@ -1,14 +1,16 @@
+@Retry(3)
+library core_test;
```
Leptopoda, line 2: Why this?

can't we add the annotation to the main method?

Leptopoda (approved):
Thanks :)
it was very annoying having a flaky CI

packages/nextcloud/test/core_test.dart
provokateurin: The annotation can only be set on a library directive. I also was confused by this at first.

PR #495, opened 2023-07-16
Refactor/bloc arguments propagation

provokateurin:
The global instances of RequestManager and NeonPlatform are used because we pass the same instance all the time.

### DIFF
--- a/packages/app/lib/apps.dart
+++ b/packages/app/lib/apps.dart
@@ -1,6 +1,5 @@
 import 'package:neon/models.dart';
 import 'package:neon/platform.dart';
-import 'package:neon/utils.dart';
 import 'package:neon_files/neon_files.dart';
 import 'package:neon_news/neon_news.dart';
 import 'package:neon_notes/neon_notes.dart';
@@ -9,12 +8,11 @@ import 'package:shared_preferences/shared_preferences.dart';
 
 List<AppImplementation> getAppImplementations(
   final SharedPreferences sharedPreferences,
-  final RequestManager requestManager,
   final NeonPlatform platform,
 ) =>
     [
-      FilesApp(sharedPreferences, requestManager, platform),
-      NewsApp(sharedPreferences, requestManager, platform),
-      NotesApp(sharedPreferences, requestManager, platform),
-      NotificationsApp(sharedPreferences, requestManager, platform),
+      FilesApp(sharedPreferences, platform),
+      NewsApp(sharedPreferences, platform),
+      NotesApp(sharedPreferences, platform),
+      NotificationsApp(sharedPreferences, platform),
     ];
--- a/packages/app/lib/apps.dart
+++ b/packages/app/lib/apps.dart
@@ -1,5 +1,4 @@
 import 'package:neon/models.dart';
-import 'package:neon/platform.dart';
 import 'package:neon_files/neon_files.dart';
 import 'package:neon_news/neon_news.dart';
 import 'package:neon_notes/neon_notes.dart';
@@ -8,11 +7,10 @@ import 'package:shared_preferences/shared_preferences.dart';
 
 List<AppImplementation> getAppImplementations(
   final SharedPreferences sharedPreferences,
-  final NeonPlatform platform,
 ) =>
     [
-      FilesApp(sharedPreferences, platform),
-      NewsApp(sharedPreferences, platform),
-      NotesApp(sharedPreferences, platform),
-      NotificationsApp(sharedPreferences, platform),
+      FilesApp(sharedPreferences),
+      NewsApp(sharedPreferences),
+      NotesApp(sharedPreferences),
+      NotificationsApp(sharedPreferences),
     ];
--- a/packages/neon/neon/lib/neon.dart
+++ b/packages/neon/neon/lib/neon.dart
@@ -19,8 +19,7 @@ import 'package:provider/provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 Future runNeon({
-  required final Iterable<AppImplementation> Function(SharedPreferences, RequestManager, NeonPlatform)
-      getAppImplementations,
+  required final Iterable<AppImplementation> Function(SharedPreferences, NeonPlatform) getAppImplementations,
   required final NeonTheme theme,
   @visibleForTesting final WidgetsBinding? bindingOverride,
   @visibleForTesting final Account? account,
@@ -33,10 +32,8 @@ Future runNeon({
   final sharedPreferences = await SharedPreferences.getInstance();
 
   final platform = await getNeonPlatform();
-  final cache = Cache(platform);
-  await cache.init();
-  final requestManager = RequestManager(cache);
-  final allAppImplementations = getAppImplementations(sharedPreferences, requestManager, platform);
+  await RequestManager.instance.initCache(platform);
+  final allAppImplementations = getAppImplementations(sharedPreferences, platform);
 
   final packageInfo = await PackageInfo.fromPlatform();
   buildUserAgent(packageInfo);
@@ -47,7 +44,6 @@ Future runNeon({
   );
 
   final accountsBloc = AccountsBloc(
-    requestManager,
     platform,
     sharedPreferences,
     globalOptions,
@@ -85,9 +81,6 @@ Future runNeon({
         ),
         Provider<GlobalOptions>(
           create: (final _) => globalOptions,
-        ),
-        Provider<RequestManager>(
-          create: (final _) => requestManager,
         ),
         Provider<AccountsBloc>(
           create: (final _) => accountsBloc,
--- a/packages/neon/neon/lib/neon.dart
+++ b/packages/neon/neon/lib/neon.dart
@@ -9,7 +9,6 @@ import 'package:neon/src/blocs/next_push.dart';
 import 'package:neon/src/blocs/push_notifications.dart';
 import 'package:neon/src/models/account.dart';
 import 'package:neon/src/models/app_implementation.dart';
-import 'package:neon/src/platform/platform.dart';
 import 'package:neon/src/theme/neon.dart';
 import 'package:neon/src/utils/global_options.dart';
 import 'package:neon/src/utils/request_manager.dart';
@@ -19,7 +18,7 @@ import 'package:provider/provider.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 Future runNeon({
-  required final Iterable<AppImplementation> Function(SharedPreferences, NeonPlatform) getAppImplementations,
+  required final Iterable<AppImplementation> Function(SharedPreferences) getAppImplementations,
   required final NeonTheme theme,
   @visibleForTesting final WidgetsBinding? bindingOverride,
   @visibleForTesting final Account? account,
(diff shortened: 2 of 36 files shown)


packages/neon/neon/lib/src/platform/platform.dart
```diff
@@ -5,22 +5,24 @@ import 'package:meta/meta.dart';
...
+      throw UnimplementedError('No implementation for platform ${Platform.operatingSystem} found');
+    }
+
+    await _platform.init();
   }
 
-  await platform.init.call();
-  return platform;
-}
+  static NeonPlatform get instance => _platform;
```
Leptopoda, line 24: `_platform` should be nullable instead of late and we should throw if it's null (not yet set up).

And doc comments for class, setup function and getter please.

packages/neon/neon/lib/src/blocs/accounts.dart
```diff
@@ -293,10 +280,7 @@ class AccountsBloc extends Bloc implements AccountsBlocEvents, AccountsBlocState
       return _userStatusesBlocs[account.id]!;
     }
 
-    return _userStatusesBlocs[account.id] = UserStatusesBloc(
-      _platform,
-      account,
-    );
+    return _userStatusesBlocs[account.id] = UserStatusesBloc(account.client);
```
Leptopoda, line 283: now it might be cleaner to do 
```dart
=> _userStatusesBlocs[account.id] ??= UserStatusesBloc(account.client);
```

packages/neon/neon/lib/src/blocs/user_statuses.dart
```diff
@@ -61,15 +58,15 @@ class UserStatusesBloc extends InteractiveBloc implements UserStatusesBlocEvents
       _updateStatus(username, Result.loading());
 
       UserStatusPublicStatus? data;
-      if (_account.username == username) {
-        final isAway =
-            _platform.canUseWindowManager && (!(await windowManager.isFocused()) || !(await windowManager.isVisible()));
-        final response = await _account.client.userStatus.heartbeat(
+      if (_client.username == username) {
+        final isAway = NeonPlatform.instance.canUseWindowManager &&
+            (!(await windowManager.isFocused()) || !(await windowManager.isVisible()));
```
Leptopoda, line 63: :confounded:

packages/neon/neon/lib/src/platform/platform.dart
Leptopoda: Unit testing this will probably be hard :thinking:

packages/neon/neon/lib/src/utils/request_manager.dart
```diff
@@ -10,13 +10,14 @@ import 'package:sqflite/sqflite.dart';
 import 'package:xml/xml.dart' as xml;
 
 class RequestManager {
-  RequestManager([
-    this.cache,
-  ]);
+  static Cache? _cache;
 
-  final Cache? cache;
+  // ignore: avoid_setters_without_getters
```
Leptopoda, line 15: I like this lint.

packages/neon/neon_news/lib/pages/article.dart
```diff
@@ -65,7 +65,7 @@ class _NewsArticlePageState extends State<NewsArticlePage> {
 
   Future _startMarkAsReadTimer() async {
     if (await widget.bloc.unread.first) {
-      if (widget.articlesBloc.options.articleDisableMarkAsReadTimeoutOption.value) {
+      if (widget.articlesBloc._options.articleDisableMarkAsReadTimeoutOption.value) {
```
Leptopoda, line 68: I don't like accessing private fields like that.
We should use a getter for this.

Maybe we should reconsider having our apps as a single library and also split them

packages/neon/neon_notes/lib/blocs/notes.dart
```diff
@@ -24,16 +24,14 @@ abstract class NotesBlocStates {
...
+    this._client,
     this.options,
-    this.requestManager,
-    this.client,
   ) {
     unawaited(refresh());
   }
 
+  final NextcloudClient _client;
   final NotesAppSpecificOptions options;
```
Leptopoda, line 34: Here the options are just public :woman_shrugging:

Leptopoda (commented):
Just some initial code style.
Not tested yet.

packages/neon/neon_news/lib/pages/article.dart
provokateurin: Oh that was a mistake, I tried hiding the options where it is possible.

packages/neon/neon/lib/src/utils/request_manager.dart
provokateurin: Do you want me to keep it or fix it?

packages/neon/neon/lib/src/blocs/user_statuses.dart
provokateurin: hm?

packages/neon/neon/lib/src/utils/request_manager.dart
Leptopoda: Please fix it

packages/neon/neon/lib/src/blocs/user_statuses.dart
Leptopoda: I don't like those inline awaits.

Both should be assigned to a variable that's then used in the assignment

packages/neon/neon/lib/src/platform/platform.dart
provokateurin: Can't we just use the same mocking approach as we used for TimerBloc?

provokateurin: @Leptopoda ping

Leptopoda: Shure this makes sense for mocking.

I still think that we should throw if it hasn't been set up with a meaningful message like `NeonPlatform hasn't been set up yet. Please make sure NeonPlatform.setup() has been called before.`

Now if someone want's to access `NeonPlatform.instance` without having called setup before they just get a late init error which is harder to debug.

provokateurin: Yes that's good!

packages/app/lib/apps.dart
```diff
@@ -9,12 +7,10 @@ import 'package:shared_preferences/shared_preferences.dart';
...
 ) =>
     [
-      FilesApp(sharedPreferences, requestManager, platform),
-      NewsApp(sharedPreferences, requestManager, platform),
-      NotesApp(sharedPreferences, requestManager, platform),
-      NotificationsApp(sharedPreferences, requestManager, platform),
+      FilesApp(sharedPreferences),
+      NewsApp(sharedPreferences),
+      NotesApp(sharedPreferences),
+      NotificationsApp(sharedPreferences),
```
Leptopoda, line 15: Maybe as a followup and but once rebased I'd love to also remove the sharedPrefs from here.

Especially as we use the same one throughout the app a static field somewhere would also work (maybe in the NeonPlatform).

packages/neon/neon/lib/neon.dart
```diff
@@ -36,11 +35,11 @@ Future runNeon({
...
   final sharedPreferences = sharedPreferencesOverride ?? await SharedPreferences.getInstance();
 
-  final platform = await getNeonPlatform();
-  final cache = Cache(platform);
+  await NeonPlatform.setup();
+  final cache = Cache();
   await cache.init();
-  final requestManager = RequestManager(cache);
-  final allAppImplementations = getAppImplementations(sharedPreferences, requestManager, platform);
+  RequestManager.cache = cache;
```
Leptopoda, line 41: the Cache is only used  by `RequestManager` thus the requestmanager should handle the setup of the cache.

The cache should also be marked as internal and hidden from the export.

packages/neon/neon/lib/src/platform/platform.dart
```diff
@@ -5,22 +5,43 @@ import 'package:meta/meta.dart';
...
 
-Future<NeonPlatform> getNeonPlatform() async {
-  final NeonPlatform platform;
-  if (Platform.isAndroid) {
-    platform = const AndroidNeonPlatform();
-  } else if (Platform.isLinux) {
-    platform = const LinuxNeonPlatform();
-  } else {
-    throw UnimplementedError('No implementation for platform ${Platform.operatingSystem} found');
+@immutable
```
Leptopoda, line 8: Below the doc comment please.

packages/neon/neon/lib/src/platform/platform.dart
Leptopoda, line 1: Maybe as a followup:
we should probably make `getApplicationCachePath` and `getUserAccessibleAppDataPath` getters.
This would make the api a bit cleaner.

  FutureOr<String> get applicationCachePath;

  FutureOr<String> get userAccessibleAppDataPath;

packages/neon/neon/lib/src/platform/platform.dart
```diff
@@ -5,22 +5,43 @@ import 'package:meta/meta.dart';
...
-  return platform;
-}
+  /// Gets the current instance of [NeonPlatform].
+  ///
+  /// Make sure [NeonPlatform.setup] has been called before accessing the instance.
+  static NeonPlatform get instance {
+    if (_platform == null) {
+      throw Exception(
+        'NeonPlatform has not been set up yet. Please make sure NeonPlatform.setup() has been called before',
+      );
```
Leptopoda, line 39:
```suggestion
      throw StateError(
        'NeonPlatform has not been set up yet. Please make sure NeonPlatform.setup() has been called and completed.',
      );
```

packages/neon/neon/lib/src/utils/request_manager.dart
Leptopoda, line 1: I don't like that RequestManager only has static fields but still has a state stored in `cache`.

We should probably make this similar to the `NeonPlatform` with the instance getter. This would also allow us to handle the entire cache initialization in here (as mentioned above).

Regarding the cache itself it should also throw if it hasn't yet been setup.

Leptopoda (commented):
Some of the comments are not entirely related to this PR.
If you don't want to implement them yet I'll do a followup addressing them.

packages/neon/neon/lib/src/utils/request_manager.dart
provokateurin: The cache doesn't need to be initialized, but I agree with the other points.

Leptopoda: then what is `Cache.init` doing :thinking:

provokateurin: Ahhh, I meant the RequestManager doesn't need to have a Cache. Of course the Cache itself needs to be initialized.

packages/app/lib/apps.dart
provokateurin: Let's do this as a follow up

Leptopoda (approved):
I really like this cleanup :)

Leptopoda:
Too little too late; I can't seem to find where `NeonPlatform.setup()` is called :thinking:

provokateurin:
Ah ffs I accidentally removed it in a rebase and didn't test afterwards again.

PR #504, opened 2023-07-19
Fix/option persistence

### LINKED ISSUE #173: Importing settings with a deleted account set as the initial account results in login page being shown

Leptopoda:
Forgot to push this earlier.
Fixes the loading of `SelectOption` value when no values have been set yet.

### DIFF
--- a/packages/neon/neon/lib/src/settings/models/select_option.dart
+++ b/packages/neon/neon/lib/src/settings/models/select_option.dart
@@ -1,8 +1,8 @@
 import 'dart:async';
 
 import 'package:collection/collection.dart';
+import 'package:neon/src/models/label_builder.dart';
 import 'package:neon/src/settings/models/option.dart';
-import 'package:neon/src/settings/widgets/label_builder.dart';
 
 class SelectOption<T> extends Option<T> {
   /// Creates a SelectOption
--- a/packages/neon/neon/lib/src/settings/models/select_option.dart
+++ b/packages/neon/neon/lib/src/settings/models/select_option.dart
@@ -12,10 +12,16 @@ class SelectOption<T> extends Option<T> {
     required super.label,
     required super.defaultValue,
     required final Map<T, LabelBuilder> values,
+
+    /// Force loading the stored value.
+    ///
+    /// This is needed when [values] is empty but the stored value should still be loaded.
+    /// This only works when [T] is of type String?.
+    final bool forceLoadValue = true,
     super.category,
     super.enabled,
   })  : _values = values,
-        super(initialValue: _fromString(values, storage.getString(key)));
+        super(initialValue: loadValue(values, storage.getString(key), forceLoad: forceLoadValue));
 
   /// Creates a SelectOption depending on the State of another [Option].
   SelectOption.depend({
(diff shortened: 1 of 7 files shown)


Leptopoda:
Still missing a unit test

packages/neon/neon/lib/src/settings/models/select_option.dart
```diff
@@ -25,9 +32,27 @@ class SelectOption<T> extends Option<T> {
...
+    /// This only works when [T] is of type String.
+    final bool forceLoadValue = false,
     super.category,
   })  : _values = values,
-        super.depend(initialValue: _fromString(values, storage.getString(key)));
+        super.depend(initialValue: loadValue(values, storage.getString(key), forceLoad: forceLoadValue));
+
+  static T? loadValue<T>(final Map<T, LabelBuilder> vs, final String? stored, {final bool forceLoad = false}) {
+    if (forceLoad) {
+      assert(vs.isEmpty, 'Force loading should only be used when no initial valus are specified.');
```
provokateurin, line 48:
```suggestion
      assert(vs.isEmpty, 'Force loading should only be used when no initial values are specified.');
```

packages/neon/neon/lib/src/settings/models/select_option.dart
```diff
@@ -12,10 +12,17 @@ class SelectOption<T> extends Option<T> {
     required super.label,
     required super.defaultValue,
     required final Map<T, LabelBuilder> values,
+
+    /// Force loading the stored value.
+    ///
+    /// This is needed when [values] is empty but the stored value should still
+    /// be loaded.
+    /// This only works when [T] is of type String.
+    final bool forceLoadValue = false,
```
provokateurin, line 21: Can't we just always force load? I fear that future devs will forget to set this option manually

provokateurin (changes requested):

packages/neon/neon/lib/src/settings/models/select_option.dart
Leptopoda: What about keeping forceLoadValue and defaulting to true?

provokateurin: Is the false value useful for anything if it is broken without it?

Leptopoda: It's not strictly broken.
The problem only occurs when `values` is empty when initializing.

Force loading will only work when the generic is of type `String` as we wouldn't know how to desetialize it without values being set.

It just so happens that we mostly use Strings (the ID of something) and use late loading quite some times

provokateurin: Ah, makes sense. How about only defaulting to true if the type is String?

Leptopoda:
I couldn't test/verify the updated branch yet.

Leptopoda:
@provokateurin  I verified it with the initial app option

packages/neon/neon/lib/src/settings/models/select_option.dart
```diff
@@ -12,10 +12,17 @@ class SelectOption<T> extends Option<T> {
     required super.label,
     required super.defaultValue,
     required final Map<T, LabelBuilder> values,
+
+    /// Force loading the stored value.
+    ///
+    /// This is needed when [values] is empty but the stored value should still
+    /// be loaded.
```
provokateurin, line 19: Can this be on one line? (same for the comment further down)

packages/neon/neon/lib/src/settings/models/select_option.dart
```diff
@@ -25,9 +32,24 @@ class SelectOption<T> extends Option<T> {
     required super.defaultValue,
     required final Map<T, LabelBuilder> values,
     required super.enabled,
+
+    /// Force loading the stored value.
+    ///
+    /// This is needed when [values] is empty but the stored value should still
+    /// be loaded.
+    /// This only works when [T] is of type String?.
```
provokateurin, line 40: Here it says String? while further up it says String. I assume you meant String here as well?

provokateurin (approved):

packages/neon/neon/lib/src/settings/models/select_option.dart
Leptopoda: `String?` is right although it's debatable wether a type definition for a generic does always implicitly include the null value as well. 

Where did I say plain `String` ?

provokateurin: The plain String is in the comment in the other constructor. I'm fine with both ways, but they should say the same thing

provokateurin (approved):

provokateurin:
The news app "How to open article" option is still broken :/

Leptopoda:
I'll take a look :sweat_smile:

provokateurin:
But other options in the news settings seem fine

PR #507, opened 2023-07-29
refactor(neon): Use image instead of bitmap library to encode bitmaps

provokateurin:
The bitmap library randomly broke my build so I checked the alternatives and the image library can actually encode bitmaps. Now the encoding doesn't use any platform specific code anymore and we pull in one less dependency (at least in the app). I didn't check for performance, but it works as expected.

### DIFF
--- a/packages/neon/neon/lib/src/utils/push_utils.dart
+++ b/packages/neon/neon/lib/src/utils/push_utils.dart
@@ -1,13 +1,13 @@
 import 'dart:convert';
 import 'dart:ui';
 
-import 'package:bitmap/bitmap.dart';
 import 'package:crypto/crypto.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 import 'package:flutter_svg/flutter_svg.dart' show SvgFileLoader, vg;
+import 'package:image/image.dart' as img;
 import 'package:meta/meta.dart';
 import 'package:neon/src/blocs/accounts.dart';
 import 'package:neon/src/models/account.dart';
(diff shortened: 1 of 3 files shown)


packages/neon/neon/lib/src/utils/push_utils.dart
```diff
@@ -132,8 +132,7 @@ class PushUtils {
               final image = recorder.endRecording().toImageSync(scaledSize.width.toInt(), scaledSize.height.toInt());
               final bytes = await image.toByteData(format: ImageByteFormat.png);
 
-              final bitmap = await Bitmap.fromProvider(MemoryImage(bytes!.buffer.asUint8List()));
-              largeIconBitmap = ByteArrayAndroidBitmap(bitmap.buildHeaded());
+              largeIconBitmap = ByteArrayAndroidBitmap(img.encodeBmp(img.decodePng(bytes!.buffer.asUint8List())!));
```
Leptopoda, line 135: I think this can be done in a more performant way.  Maybe:

```suggestion
final imageStuff = img.Image.fromBytes(
  width: scaledSize.width.toInt(),
  height: scaledSize.height.toInt(),
  bytes: bytes!.buffer,
  numChannels: 4,
);

largeIconBitmap = ByteArrayAndroidBitmap(img.encodeBmp(imageStuff));
```

I can further investigate this if needed

provokateurin: Ah I see, I was already wondering if the png step was really necessary

provokateurin: How is this more performant?

Leptopoda: I just thought it would be as we wouldn't need to encode and decode again.

I didn't test it yet

provokateurin: But it still does the conversion from bytes to image? I'd think it's the same, but maybe your suggestion is more performant :woman_shrugging:

provokateurin: Can we just merge this and optimize it later?

Leptopoda (approved):

PR #509, opened 2023-07-29
fix(neon): Hide useless push notifications popups

provokateurin:
Popups were shown on platforms that don't even support push notifications.

### DIFF
--- a/packages/neon/neon/lib/src/utils/global_popups.dart
+++ b/packages/neon/neon/lib/src/utils/global_popups.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
 import 'package:neon/l10n/localizations.dart';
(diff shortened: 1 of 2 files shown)


packages/neon/neon/lib/src/utils/global_popups.dart
```diff
@@ -31,27 +32,31 @@ class GlobalPopups {
     final globalOptions = Provider.of<GlobalOptions>(context, listen: false);
     final firstLaunchBloc = Provider.of<FirstLaunchBloc>(context, listen: false);
     final nextPushBloc = Provider.of<NextPushBloc>(context, listen: false);
+    final platform = Provider.of<NeonPlatform>(context, listen: false);
 
     firstLaunchBloc.onFirstLaunch.listen((final _) {
-      if (globalOptions.pushNotificationsEnabled.enabled) {
-        if (!context.mounted) {
```
Leptopoda, line 37: Are we sure that the context is always mounted when the the listener gets called?

Leptopoda (changes requested):
We should make GlobalPopups disposable (canceling the listeners) and while being at it also update the context when `register` gets called while already registered.

packages/neon/neon/lib/src/utils/global_popups.dart
provokateurin: The home page is always kept alive once it has been created, right? Then the context is always mounted.

Leptopoda: Than we should add an assert.

While this is true for now I'd like to prevent this popping up in the future

provokateurin: I implemented the dispose logic now, but then it's not even needed, right? Or shall we just keep it so we don't forget it in the future if things may change?

Leptopoda: I think we should keep it.

PR #511, opened 2023-07-30
Cleanup/docs

provokateurin:
The CoC is in the top level so I figured moving the contributing docs there makes sense. We can still list it in the ToC of the docs.


Leptopoda (approved):
Makes sense :+1:

Issue #435, opened 2023-06-30
Improve test/code coverage


provokateurin:
Currently we test most of the endpoints we use, but don't cover all the different options.
Add more tests to cover all endpoints and all available options.

provokateurin:
This won't be feasible anymore with the automatic specs, but we should at least try to cover the endpoints used by neon.

provokateurin:
For WebDAV I would like to replicate all the tests of https://github.com/notroj/litmus. It is a test suite for testing WebDAV servers for compliance of the spec, but it will help us to cover most WebDAV features and also find possible problems in our implementation.

provokateurin:
For the normal APIs I would like to switch to [gherkin](https://cucumber.io/docs/gherkin/reference/) as apps like [Talk](https://github.com/nextcloud/spreed/tree/master/tests/integration/features) already do it. That will make it possible to write tests independent of the platform/language/client/framework. This way we can share some efforts and don't even need to write any tests for Talk manually since they already exists. There is also a chance that this will become a task I will work on officially in the future.

provokateurin:
There is also [tcases](https://github.com/Cornutum/tcases/tree/master/tcases-openapi) which we can use to generate tests based on the specs. The output is a JSON and we would need to write another generator that can convert that format into Dart tests using the generated API client.
I'm not sure how useful this will be, but I'm pretty sure we could uncover at least some bugs with this and properly scale our code coverage since it will test everything without someone having to write manual tests.

provokateurin:
And we could also think about using test fixtures for the HTTP calls (similarly to how golden tests work for widget rendering). This way we would catch any possible change on the server or client side. I searched for a package that already implements this but couldn't find any.

Leptopoda:
> And we could also think about using test fixtures for the HTTP calls

What exactly do you mean?

provokateurin:
Basically logging all HTTP parameters like request url, headers, body and response headers and body. The tests would compare the values to the saved ones and fail if they differ. Some of the stuff would need to be done using regex since some tests involve fields that contain a timestamp.

It won't help much, just increase the confidence of our testing.

provokateurin:
tcases is an interesting tool, but i don't think it will be of much use since it relies on example data defined in the spec to generate the happy cases, but there is no way for us to inject it other than manual patching which will never scale with generated specs.

provokateurin:
[...]

PR #513, opened 2023-07-30
Feature/litmus tests

provokateurin:
https://github.com/provokateurin/nextcloud-neon/issues/435

Implements most of the applicable tests from https://github.com/notroj/litmus.
Sadly the tests are not as extensive as I had hoped. Some of the existing tests have to stay. Though it helped to uncover some problems with litmus and the Nextcloud implementation and also in our client code.

### DIFF
--- a/packages/nextcloud/test/webdav_test.dart
+++ b/packages/nextcloud/test/webdav_test.dart
@@ -2,7 +2,6 @@
 library webdav_test;
 
 import 'dart:convert';
-import 'dart:io';
 import 'dart:math';
 import 'dart:typed_data';
 
@@ -23,12 +22,6 @@ void main() {
       client = await getTestClient(container);
     });
     tearDown(() => container.destroy());
-
-    test('Get status', () async {
-      final options = await client.webdav.options();
-      expect(options.capabilities, containsAll(['1', '3']));
-      expect(options.searchCapabilities, hasLength(0));
-    });
 
     test('List directory', () async {
       final responses = (await client.webdav.propfind(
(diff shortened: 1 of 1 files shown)


packages/nextcloud/test/litmus_test.dart
```diff
@@ -0,0 +1,246 @@
...
+          final downloadedContent = await client.webdav.get(path);
+          expect(downloadedContent, equals(content));
+        });
+      }
+
+      test('put_no_parent', () async {
+        expect(
+          () => client.webdav.put(Uint8List(0), '409me/noparent.txt'),
+          // https://github.com/nextcloud/server/issues/39625
+          throwsA(predicate((final e) => (e! as DynamiteApiException).statusCode == 409)),
```
Leptopoda, line 57:
```suggestion
          throwsA(predicate<DynamiteApiException>((final e) => e.statusCode == 409)),
```

Same for all the other tests

packages/nextcloud/test/litmus_test.dart
```diff
@@ -0,0 +1,246 @@
...
+      test('mkcol_no_parent', () async {
+        expect(
+          () => client.webdav.mkcol('409me/noparent'),
+          throwsA(predicate((final e) => (e! as DynamiteApiException).statusCode == 409)),
+        );
+      });
+
+      test('mkcol_with_body', () async {
+        // This test is not applicable because we only write valid request bodies
+      });
```
Leptopoda, line 109: Can we skip this?
Or even implement and skip it.

Same for the other empty tests. This would exclude them from visually from the other tests in my IDE and probably also remove the need to set up and destroy the docker container for them.

packages/nextcloud/test/litmus_test.dart
```diff
@@ -0,0 +1,246 @@
...
+        expect(response.statusCode, 201);
+      });
+
+      test('large_get', () async {
+        final response = await client.webdav.put(Uint8List(largefileSize), 'test.txt');
+        expect(response.statusCode, 201);
+
+        final downloadedContent = await client.webdav.get('test.txt');
+        expect(downloadedContent, hasLength(largefileSize));
+      });
```
Leptopoda, line 239: Do you think it makes sense to unify both tests?

We could have one single test and add the `get` or `put` information in the `reason` field of the `expect`.
I think that's better as a failing put will also result in a failing get thus I don't see a reason to separate them.

Leptopoda (commented):
I'm not a WebDAV expert by any means but other than the cosmetic changes it looks reasonable.

packages/nextcloud/test/litmus_test.dart
provokateurin: Sure, I'd still like to keep the comment so later on we can see why the test doesn't exist

packages/nextcloud/test/litmus_test.dart
provokateurin: I was thinking the same. The official litmus tests are not independent of each other so they need this but we don't

Leptopoda (approved):

PR #517, opened 2023-07-30
refactor(neon,neon_files,neon_news,neon_notes,neon_notifications): Use shared constants for app ids

provokateurin:


packages/neon/neon/lib/src/models/app_ids.dart
```diff
@@ -0,0 +1,6 @@
+abstract class AppIDs {
```
Leptopoda, line 1: Enum?

Or because other apps need to extend this?

packages/neon/neon_news/lib/neon_news.dart
```diff
@@ -55,9 +55,7 @@ class NewsApp extends AppImplementation<NewsBloc, NewsAppSpecificOptions> {
   NewsApp(super.sharedPreferences, super.requestManager, super.platform);
 
   @override
-  String id = appId;
-
-  static const String appId = 'news';
+  String id = AppIDs.news;
```
Leptopoda, line 58: I think we can remove this now.

Leptopoda (approved):

packages/neon/neon/lib/src/models/app_ids.dart
provokateurin: You need to have it as a string value for comparison and with an enum using a field it would always be `AppIDs.news.value` or something like that. I don't see the benefit there

Leptopoda: Makes sense.

Can we at least make the class `final` or even `sealed`?
I don't think an app needs to extend it.

Leptopoda (approved):

PR #519, opened 2023-07-31
fix(dynamite): Skip comparing nullable in type results

provokateurin:
In https://github.com/provokateurin/nextcloud-neon/commit/7a71e010b6fe934a92238720d9bae159570d760f the whole type was added to check if the type is already generated. This breaks https://github.com/provokateurin/nextcloud-neon/pull/465 as it has enums that are both use as nullable and non-nullable type.


Leptopoda (approved):
we could now also cleanup this:

[patch.txt](https://github.com/provokateurin/nextcloud-neon/files/12222565/patch.txt)

PR #520, opened 2023-07-31
Refactor/webdav methods

provokateurin:


Leptopoda (changes requested):
Looks good but I'd like the doc comments to follow:
https://dart.dev/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs

Leptopoda (commented):
Still missing punctuation in some places (sentences should end with a period).

We should also document all parameters. Especially the optional ones are currently undocumented.

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -76,24 +76,48 @@ class WebDavClient {
...
 
-  /// make a dir with [remotePath] under current dir
-  Future<HttpClientResponse> mkdir(
-    final String remotePath, {
+  /// Creates a collection at [path].
+  ///
+  /// See http://www.webdav.org/specs/rfc2518.html#METHOD_MKCOL for more information.
+  Future<HttpClientResponse> mkcol(
+    final String path, {
     final bool safe = true,
```
Leptopoda, line 121: Safe needs to be documented

Leptopoda (approved):
Just a small thing

packages/nextcloud/lib/src/webdav/client.dart
provokateurin: Honestly I would just remove it. We never use it and it feels odd to have such an option. The application itself should decide how it handles the failure

provokateurin: In the future we should just create a few exceptions that model the different status codes, then it will be much easier for applications to handle specific errors.

Leptopoda (approved):

PR #523, opened 2023-08-03
chore(neon_files): Update icon

provokateurin:


provokateurin:
Yep, the icon was updated on the server.

PR #533, opened 2023-08-05
Fix/coverage

provokateurin:
Adds some coverage ignores to exclude code that is not relevant.


packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -970,6 +966,11 @@ TypeResult resolveObject(
                   )
                   ..redirect = refer('_\$${state.prefix}$identifier'),
               ),
+              Constructor(
+                (final b) => b
+                  ..name = '_'
+                  ..constant = true,
+              ),
```
Leptopoda, line 973: Is there a reason this one moved down?

Leptopoda (approved):
:+1:

packages/dynamite/dynamite/lib/src/openapi_builder.dart
provokateurin: Yeah because the formatter would move it down and then break the coverage ignores :roll_eyes:

PR #534, opened 2023-08-06
refactor(neon): Align colors with design guidelines

provokateurin:


Leptopoda:
:thinking:

The same document also mentions:
> Android: Material Design guidelines

provokateurin:
Hm, shall we open an issue in the documentation repo and tag the designers?

Leptopoda:
Yeah please.

I wondered the same thing when I redid the design of the cookbook app.

https://m3.material.io/styles/color/the-color-system/key-colors-tones

For reference

provokateurin:
I just read the guidelines again and I think my changes are not correct. As far as I understand it the color codes in the guidelines are meant for web, but we should probably stick to the Material Design colors. Of course we don't only run on Android, but since we use a Material design we should stick to the colors. @Leptopoda do you agree?

Leptopoda:
So just follow the material guidelines everywhere?
Sounds good :+1: 

We could still leave the colors in the NcColors class but not use them in the theme.
I don't know why but maybe someone wants to use them?

provokateurin:
Hm I don't think people really care to use this or that green tone and it's just awkward if some elements are colored slightly different.

provokateurin:
@Leptopoda done

Leptopoda (approved):
There now is some duplication with multiple red colors.

I think it's ok as that's somewhat the reason we have this file just wanted to mention it.

PR #538, opened 2023-08-10
Feature/update screenshots

provokateurin:


Leptopoda (commented):
Can we do this in a different way?
I don't like the hacks needed to make keys work.

Maybe finding it by the label (text)? This might get harder if we want to make multiple screenshots for the app stores but it would at least keep the app code cleaner

provokateurin:
I was using the keys before, but searching for the labels makes so much more sense! I'll implement it that way

packages/app/integration_test/screenshot_test.dart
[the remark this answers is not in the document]
provokateurin: Generally yes, but I don't know if it will work for this as it has to actually function I believe. I will check it out though

packages/neon/neon/lib/src/utils/user_agent.dart
```diff
@@ -0,0 +1,9 @@
+import 'package:meta/meta.dart';
+
+late String? _userAgent;
```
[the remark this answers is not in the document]
provokateurin: The tests were setting this every time `runNeon` was called and the variable was never reset. Before it was `late` which would throw errors

provokateurin:
> This might get harder if we want to make multiple screenshots for the app stores (in different languages)

No we can just use the AppLocalizations and search for the text we get from there. Then it doesn't matter which language the text actually has

packages/neon/neon/lib/src/settings/widgets/settings_list.dart
```diff
@@ -1,9 +1,7 @@
 import 'package:flutter/material.dart';
-import 'package:meta/meta.dart';
 import 'package:neon/src/settings/widgets/settings_category.dart';
 import 'package:scrollable_positioned_list/scrollable_positioned_list.dart';
 
-@internal
 class SettingsList extends StatelessWidget {
```
Leptopoda, line 5: @visibleForTesting ?

packages/neon/neon/lib/src/utils/user_agent.dart
```diff
@@ -0,0 +1,9 @@
...
+
+@internal
+set neonUserAgent(final String value) => _userAgent = value;
```
Leptopoda, line 6: What do you think about making this a method like:

```dart
void buildNeonUseragent(PackageInfo packageInfo){
  var buildNumber = packageInfo.buildNumber;
  if (buildNumber.isEmpty) {
    buildNumber = '1';
  }
  _userAgent = 'Neon ${packageInfo.version}+$buildNumber';
}
```

or maybe using a stringbuffer. This would cleanup neon.dart

packages/neon/neon/lib/neon.dart
```diff
@@ -14,27 +13,24 @@ import 'package:neon/src/platform/platform.dart';
...
-@internal
-late final String neonUserAgent;
-
 Future runNeon({
   required final Iterable<AppImplementation> Function(SharedPreferences, RequestManager, NeonPlatform)
       getAppImplementations,
   required final NeonTheme theme,
   final WidgetsBinding? bindingOverride,
-  final SharedPreferences? sharedPreferencesOverride,
   final Account? account,
```
Leptopoda, line 26: can we mark those two (and the other ones as needed) as `visibleForTesting`?

Leptopoda (approved):

PR #547, opened 2023-08-13
Fix/dependabot

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/488

It works indeed only when pointing to each package individually (the documentation says so and I checked that it works locally).

Since we don't have any cyclic dependencies everything should work fine. I guess they will all be a bit delayed as the updates will follow after another (due to the changing dependency constraints).

I added a script to generate this config as it is easier to update if we have new packages and also keeps all the entries in line (like using the same label).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,13 +7,73 @@ updates:
     labels:
       - "dependencies"
   - package-ecosystem: "gradle"
-    directory: "/"
+    directory: "/packages/app/android"
     schedule:
       interval: "daily"
     labels:
       - "dependencies"
   - package-ecosystem: "pub"
-    directory: "/"
+    directory: "/packages/app"
+    schedule:
+      interval: "daily"
+    labels:
+      - "dependencies"
+  - package-ecosystem: "pub"
+    directory: "/packages/dynamite/dynamite"
+    schedule:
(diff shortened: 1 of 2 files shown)


provokateurin:
Can't wait for the PR flood :sweat_smile:

Leptopoda:
Should we also document the process of adding a new package?

It currently involves updating the melos config and now this.

Probably also a note about static analysis.

Leptopoda (approved):
Changes themselves look good

provokateurin:
> Should we also document the process of adding a new package?

Good idea

> It currently involves updating the melos config and now this.

What melos config? It is picked up automatically

> Probably also a note about static analysis.

What do you mean? How to add the correct config for static analysis?

Leptopoda:
> What melos config? It is picked up automatically

to my understanding we need to re run `melos bootstrap`

provokateurin:
> to my understanding we need to re run melos bootstrap

Ah I forgot that part

> Just mention that we need to add nit_pick to a new package.

The license and a proper .gitignore should be added as well

provokateurin:
I think we should setup a CI workflow that checks that everything is generated correctly. I has to run all the different generate scripts and melos bootstrap as well.

Issue #445, opened 2023-07-05
Performance profiling


Leptopoda:
we should do some performance profiling before we go live.

I just found a few issues on some older devices

Possible optimizations:
- [x] lazily build lists #585 
- [x] pre compile svg images (similar to #589)
- [x] reduce rebuilds #377 
- [x] https://github.com/nextcloud/neon/blob/69d69657907b937866b9716cbdcb361fd91e1589/packages/neon/neon_files/lib/widgets/browser_view.dart#L72-L79

provokateurin:
I always experience the most jank when starting the app. I've also seen that the active app and the active account is set multiple times (both are not distinct but they shouldn't have to be) which probably causes a lot of that. The other thing is the network requests that are all fired at the same time and the deserialization of the responses (including the cached data). The later could be solved by queueing all requests (which will be partialy done with https://github.com/nextcloud/neon/issues/543 anyway).

Leptopoda:
I have a local branch that reworks the files rendering and I cleaned up the code a lot and made it more performant by multiple orders of magnitude.

It currently depends on #377 and maybe the go_router work.

provokateurin:
That sounds great!

Issue #282, opened 2023-04-24
[RFC]: NeonListView performance


Leptopoda:
As talked about on matrix I found that the layout phase of `NeonListView` takes to long.
This is probably the case because the main content isn't built lazily.

Just asking about how we would approach a refactor of this (and related) widget/s.
IMO we should only expose a lazy way for building this Widget. Consumers of neon shouldn't be able to go the wrong route.

Also I'd love to remove the `withFloatingActionButton` as we could just search the the WidgetTree ourselves (thinking of `Scaffold.of(context).hasFloatingActionButton`).

### 1. How should the API of the widget be?
Currently all implementations expect a list of all items and the builder function to return the context and an item while the flutter style would be to get an itemCount and a builderfunction that provides the context and an index (see `NullableIndexedWidgetBuilder`). 

Both are possible and the current code is tailored towards the first one.

### 2. Move the Children into their own Widget?
Currently the children are all build in the builder provided to the NeonListView. Making them their own Widget would allow us to also make a PrototypeItem that we could provide to the list. Most (if not all) implementations of the List could profit of only needing to layout one child.

### 3. Provide a second NeonListView.ordered factory
Most implementations of the NeonListView are wraped by SortBoxBuilder (makes sense as lists are sortable). This would just be a convenience method and the old implementation would still be possible.

We could also just make a separate convenience Widget like `NeonSortedListView` that does this.

### 4. Improve SortBoxBuilder nullability.
When the list provided to the SortBoxBuilder is null we will still call the builder just without a list value. I think we should just return a container if the input is null and not build the children.

Currently this would lead to not rendering the entire NeonListView wich might be used to display an error in this case so this is only desirable when we have a merged version of both (see above).

### 5. Avoid nested SortBoxBuilder
The notes_view currently nests two SortBoxBuilder(s). One for the favorites and one for nonFavorites. This means we couldn't profit form such a mentioned `NeonSortedListView`. In theory the nesting is only done to sort based on the favorite state and then the notesSortPropertyOptions.

Recursive SortBox anyone :eyes: #276
(I already have an idea of how we could expose a nice api for the sortbox to support this but don't know if it would work yet)

----

Those are just ideas and I'd love some feedback on them. Obviously this would need separate PRs probably from the bottom up.
[...]

PR #556, opened 2023-08-13
Reactor/files app

Leptopoda:
improves the files app performance as mentioned in #445.
#282  will still gain us a bit more.

@provokateurin please review thoroughly :)


provokateurin:
Why is it in draft if I should already review? :eyes:

Leptopoda:
> Why is it in draft if I should already review? eyes

Because I love the UX of github :roll_eyes:

provokateurin:
Ah yeah it remembers what type of PR you last submitted. I also hate it so much lol

packages/neon/neon_files/lib/widgets/actions.dart
```diff
@@ -0,0 +1,163 @@
...
+    final bloc = Provider.of<FilesBloc>(context, listen: false);
+    final browserBloc = bloc.browser;
+    switch (action) {
+      case FilesFileAction.toggleFavorite:
+        if (details.isFavorite ?? false) {
+          bloc.removeFavorite(details.path);
+        } else {
+          bloc.addFavorite(details.path);
+        }
+        break;
```
provokateurin, line 26: With Dart 3 we don't need breaks anymore (also for the cases below)

packages/neon/neon_files/lib/widgets/actions.dart
```diff
@@ -0,0 +1,163 @@
...
+          value: details.name,
+        );
+        if (result != null) {
+          bloc.rename(details.path, result);
+        }
+        break;
+      case FilesFileAction.move:
+        final b = bloc.getNewFilesBrowserBloc();
+        final originalPath = details.path.sublist(0, details.path.length - 1);
+        b.setPath(originalPath);
```
provokateurin, line 51: I know you didn't introduce this code, but could you make it so that it is possible to pass an optional initial path to the files browser bloc instead of having to call setPath directly after it?

packages/neon/neon_files/lib/widgets/actions.dart
```diff
@@ -0,0 +1,163 @@
...
+        );
+        b.dispose();
+        if (result != null) {
+          bloc.move(details.path, result..add(details.name));
+        }
+        break;
+      case FilesFileAction.copy:
+        final b = bloc.getNewFilesBrowserBloc();
+        final originalPath = details.path.sublist(0, details.path.length - 1);
+        b.setPath(originalPath);
```
provokateurin, line 68: Same as above

packages/neon/neon_files/lib/models/file_details.dart
```diff
@@ -11,7 +11,53 @@ class FileDetails {
...
+        etag = file.etag,
+        mimeType = file.mimeType,
+        lastModified = file.lastModified,
+        hasPreview = file.hasPreview,
+        isFavorite = file.favorite,
+        progress = null,
+        isUploading = false,
+        isDownloading = false;
+
+  FileDetails.uploading({
```
provokateurin, line 33:
```suggestion
  FileDetails.fromUploadTask({
```

packages/neon/neon_files/lib/models/file_details.dart
```diff
@@ -11,7 +11,53 @@ class FileDetails {
...
+        progress = task.progress,
+        isUploading = true,
+        isDownloading = false,
+        isDirectory = false,
+        etag = null,
+        mimeType = null,
+        hasPreview = null,
+        isFavorite = null;
+
+  FileDetails.downloading({
```
provokateurin, line 47:
```suggestion
  FileDetails.fromDownloadTask({
```

packages/neon/neon_files/lib/widgets/browser_view.dart
```diff
@@ -96,39 +83,32 @@ class _FilesBrowserViewState extends State<FilesBrowserView> {
...
+                                      } else if (matchingUploadTasks != null) {
+                                        details = FileDetails.uploading(
+                                          task: matchingUploadTasks,
+                                        );
+                                      } else {
+                                        details = FileDetails.fromWebDav(
+                                          file: file,
+                                          path: widget.bloc.path.value,
+                                        );
+                                      }
```
provokateurin, line 105: I wonder if it would make sense to put this logic into a factory constructor :thinking:

packages/neon/neon_files/lib/models/file_details.dart
```diff
@@ -30,4 +76,10 @@ class FileDetails {
   final bool? hasPreview;
 
   final bool? isFavorite;
+
+  final Stream<double>? progress;
+  final bool isUploading;
+  final bool isDownloading;
+
+  bool get isLoading => isUploading || isDownloading;
```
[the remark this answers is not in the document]
provokateurin: I've read the code further down and I think it would make sense to use a nullable enum for this. That's way easier to understand and using 2 booleans is a bit ugly especially since they are exclusive of each other.

provokateurin (changes requested):
As far as I understand the performance improvements mostly come from caching widgets or am I wrong here?

packages/neon/neon_files/lib/utils/task.dart
provokateurin, line 1: Can you prefix all of these classes with `Files` please. The sound so generic that they should belong into neon, but they shouldn't (at least right now).

packages/neon/neon_files/lib/widgets/file_list_tile.dart
```diff
@@ -0,0 +1,133 @@
...
+    final bloc = Provider.of<FilesBloc>(context);
+
+    Widget icon = Center(
+      child: details.hasTask
+          ? StreamBuilder<double>(
+              stream: details.task!.progress,
+              builder: (final context, final progress) => Column(
+                children: [
+                  Icon(
+                    details.task is UploadTask ? MdiIcons.upload : MdiIcons.download,
```
provokateurin, line 95: Please us a switch for this. This code was exactly the reason why I wanted to move away from two booleans as I think it is not good to only check one condition for both options.

provokateurin (changes requested):

packages/neon/neon_files/lib/widgets/file_list_tile.dart
Leptopoda: It's really awesome in such situations :)

packages/neon/neon_files/lib/utils/task.dart
```diff
@@ -0,0 +1,66 @@
+part of '../neon_files.dart';
+
+sealed class FileTask {
```
provokateurin, line 3:
```suggestion
sealed class FilesTask {
```

Because of the app name :grimacing:

provokateurin (changes requested):

provokateurin (approved):

PR #558, opened 2023-08-13
feat(tool): Install setup tools from workspace

provokateurin:


provokateurin:
No idea what the error message means, it works locally. Maybe yj/jq version mismatch?

Leptopoda:
I'm no CI expert but to my understanding this runs on ubuntu 22.04

https://github.com/nextcloud/neon/blob/main/.github/workflows/dart.yml#L17

And afaict ubuntu doesn't even ship yq. So we'd need to first install it.

provokateurin:
They install some additional tools on the runner images. And if yq wasn't installed then there would be a command not found error

provokateurin:
The action uses https://github.com/mikefarah/yq while I have https://github.com/kislyuk/yq. I changed the syntax so it works on both variants.

Leptopoda (approved):

PR #560, opened 2023-08-13
chore(app): Update pubspec.lock

provokateurin:
We have to do it manually as dependabot doesn't respect our dependency_overrides and doesn't even update this lockfile if one of the other packages is updated.

Maybe we should add a CI workflow that runs once a day and submits the updated lockfile?
OTOH from now on we should update the lockfile directly, I just didn't do it as there were so many updates at once.

### DIFF
--- a/packages/app/pubspec.lock
+++ b/packages/app/pubspec.lock
@@ -320,10 +320,10 @@ packages:
     dependency: transitive
     description:
       name: flutter_cache_manager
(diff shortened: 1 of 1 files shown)


provokateurin:
I'll pickup https://github.com/nextcloud/neon/pull/561 as well

Leptopoda (approved):

PR #580, opened 2023-08-17
fix(tool): Fix building app in docker

provokateurin:
https://github.com/nextcloud/neon/actions/runs/5888861326/job/15970958322

This step should be added to https://github.com/nextcloud/neon/issues/575, but I want to switch it against something based on `melos list` as that is way more dynamic and won't need to be update.

### DIFF
--- a/tool/build-app.sh
+++ b/tool/build-app.sh
@@ -45,7 +45,7 @@ if [[ "$target" == "linux/arm64" ]] || [[ "$target" == "linux/amd64" ]]; then
   ./tool/build
 
   run_args=()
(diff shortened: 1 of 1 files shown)


Leptopoda:
Can't we just mount the entire working directory?
Or why are we doing this in the first place.

Leptopoda (approved):
I mean LGTM as it fixes the issue but I don't really like what we are doing.

Is it that hard to switch to melos or mount everything in this PR?

provokateurin:
I tried switching to melos once but then messed it up. Shouldn't be too hard to pull off :thinking:

PR #581, opened 2023-08-17
Feature/more apis

provokateurin:
Depends on https://github.com/nextcloud/openapi-extractor/issues/8 as I had to patch the files and core specs.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,6 +19,3 @@
 [submodule "external/flathub-shared-modules"]
 	path = external/flathub-shared-modules
 	url = https://github.com/flathub/shared-modules.git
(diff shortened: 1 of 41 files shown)


external/nextcloud-openapi-extractor
Leptopoda, line 1: Why did you remove the extractor?

provokateurin: The extractor is now a dev dependency of the server. That way we can be sure the server code is compatible with the used extractor version. The same will apply for apps as they will also have their own extractor dependency pinned.

provokateurin:
Will be ready once https://github.com/nextcloud/neon/pull/597 is merged

Leptopoda (approved):

PR #582, opened 2023-08-18
Fix/dynamite conflicting variables

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/572

### DIFF
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -276,7 +276,7 @@ class OpenAPIBuilder implements Builder {
                   ..fields.add(
                     Field(
                       (final b) => b
-                        ..name = 'rootClient'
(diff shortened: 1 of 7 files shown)


provokateurin:
@Leptopoda do you know what the best way is for creating tests for a build_runner? i would like to add a spec that intentionally uses the internal variable names as parameters to validate that correct code is generated.

Leptopoda:
I don't know nut iirc when I contributed to built_value they just have a example file in the test dir and let the build_runner run.

This would mean just adding a spec into the test folder and generating a client but maybe there is a better way :shrug:

provokateurin:
There is https://github.com/google/json_serializable.dart/blob/master/json_serializable/test/ensure_build_test.dart which seems to just build the example and check that the code is working. I will try to implement that as well

provokateurin:
Bleh it's not that easy, I would just like to fix this now

Leptopoda (approved):

PR #583, opened 2023-08-18
chore(tool): Update color to Nextcloud blue

provokateurin:
Uses the Nextcloud blue. Changing the logos so that they are white on a blue circle background proves to be trickier than I anticipated because flutter_svg doesn't support percentage heights and widths. We need to fix that later.


Leptopoda:
If wee need to change it later on what's the point of this PR?

Diff looks good but again untested.
Could you please update #447 or document the needed changes elsewhere?

Leptopoda (approved):

provokateurin:
Hm true

PR #589, opened 2023-08-19
Feature/user status icons

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/399.
The server icons will also be used there.
According to https://docs.nextcloud.com/server/latest/developer_manual/html_css_design/icons.html this is a deprecated feature, but there is no drop-in replacement and many apps still rely on these icons.


provokateurin:
I think the SVGs should be [precompiled](https://github.com/dnfield/flutter_svg/tree/master/packages/flutter_svg#precompiling-and-optimizing-svgs) to save on size and increase performance.

Leptopoda (commented):
1. I fear that this will bloat the apk size quite significantly.

2. A get it on the apple store badge slipped into this.
I doubt this is needed

provokateurin:
> I fear that this will bloat the apk size quite significantly.

The folder is about 2.1mb, I haven't checked the resulting apk. I already thought about limiting the icons we include, but the only way I see would be parsing https://github.com/nextcloud/server/blob/master/core/src/icons.js which might be viable though.

> A get it on the apple store badge slipped into this. 

I forgot to remove those icons as they are indeed not available.

Leptopoda:
We might be able to get it down by pre processing it with the svg build_runner although this could also increase it; testing needed.

Another thing is that visually some icons look identical (e.g. I didn't see a difference between light and dark of the sun with cloud icon).

Finally there should be an easy way for single app apps to opt out of this.

2.1mB might be acceptable especially if the above  can bring it down further :thinking:

provokateurin:
I tried the preprocessing that flutter_svg can do but I think it's not meant for optimizing the size. Also plain text is probably better compressable in apks.

Filtering the dark icons makes sense, we render them with a different color anyway.

provokateurin:
I will do some testing, but this feature is definitely needed for useful unified search (and other places as well).

provokateurin:
I removed the duplicate and unused icons and precompiled the remaining ones and brought the size down to 1664kB. I think this is quite acceptable, but maybe it can be optimized even more in some way?

On another note: It should be possible to load these icons from the server just like the web interface does, but we would need to have the mapping between the icon name and it's path. That way we wouldn't need to add any assets which seems like a good move.

Leptopoda (commented):
Overall looks good.

I still think that it should be easy to exclude the assets. Mabe just a document in the docs folder mentioning something like:

```
android {
  aaptOptions {
    ignoreAssetsPattern "assets/icons"
  }
}
```

If that would even work :thinking:

Leptopoda:
> It should be possible to load these icons from the server just like the web interface does

I'm not that sure how this would impact the performance. Especially if they aren't precompiled.

provokateurin:
> I'm not that sure how this would impact the performance. Especially if they aren't precompiled.

I don't think the performance hit would be too big as we only show icons that are actually used and there usually aren't a lot of icons displayed at the same time

Leptopoda (approved):

PR #592, opened 2023-08-19
Refactor/nextcloud

Leptopoda:


packages/nextcloud/lib/src/client.dart
```diff
@@ -18,7 +18,7 @@ import 'package:nextcloud/src/webdav/client.dart';
 class NextcloudClient extends DynamiteClient {
   // ignore: public_member_api_docs
   NextcloudClient(
-    super.baseURL, {
+    final String baseURL, {
```
provokateurin, line 21: Why not use Uri here?

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -56,24 +54,22 @@ class WebDavClient {
...
-            while (part.endsWith('/')) {
-              part = part.substring(0, part.length - 1); // coverage:ignore-line
-            }
-            return part;
-          })
-          .where((final part) => part.isNotEmpty)
-          .join('/');
+  Uri _constructUri([final Uri? path]) {
+    assert(
+      path == null || path.path == '/' || !path.path.startsWith('/') || !path.path.endsWith('/'),
```
provokateurin, line 59: Why is `/` as path not allowed?

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -56,24 +54,22 @@ class WebDavClient {
...
+
+    final pathBuffer = StringBuffer(webdavBasePath);
+    if (path != null && path.path != '/') {
+      pathBuffer
+        ..write('/')
+        ..write(path.path);
+    }
+
+    final builder = UriBuilder.fromUri(rootClient.baseURL)..path = pathBuffer.toString();
+    return builder.build();
```
provokateurin, line 71: Wouldn't using `Uri(pathSegments: ...)` be better?

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -56,24 +54,22 @@ class WebDavClient {
...
+    );
+
+    final pathBuffer = StringBuffer(webdavBasePath);
+    if (path != null && path.path != '/') {
+      pathBuffer
+        ..write('/')
+        ..write(path.path);
+    }
+
+    final builder = UriBuilder.fromUri(rootClient.baseURL)..path = pathBuffer.toString();
```
provokateurin, line 70: Just setting to path will break instances hosted at a subfolder. The baseURL might already contain a path.

provokateurin (changes requested):

packages/nextcloud/lib/src/webdav/client.dart
Leptopoda: `UriBuilder` sadly doesn't support pathSegments :(

packages/nextcloud/lib/src/webdav/client.dart
Leptopoda: I just copied the old behavior. '/' is ollwed but everything else shouldn't start or end with a `/`.

Otherwise we could get a `cloud.example.tld/remote.php/webdav/something/` with a trailing `/`. If this is ok I'll happily simplify the code.

packages/nextcloud/lib/src/client.dart
Leptopoda: It's a pain to migrate our login stuff and as it is fragile enough I didn't change this.
Should this also be included in this PR?

packages/nextcloud/lib/src/webdav/client.dart
provokateurin: I think that was only done to correctly put together the path. I assume your changes will work with leading and trailing `/`?

Leptopoda: IIRC it does; I'll check this later.

packages/nextcloud/lib/src/client.dart
provokateurin: I think it would be better to do so

Leptopoda:
I just pulles out the URI stuff and will make a followup PR after this one is merged

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -33,8 +33,8 @@ class WebDavClient {
     for (final header in {
       HttpHeaders.contentTypeHeader: 'application/xml',
       ...?rootClient.baseHeaders,
-      if (headers != null) ...headers,
-      if (rootClient.authentications.isNotEmpty) ...rootClient.authentications.first.headers,
```
provokateurin, line 37: This check needs to stay or you have to use `firstOrNull`

provokateurin (changes requested):

provokateurin (approved):

Issue #568, opened 2023-08-15
refactor FilesBloc initial path 


Leptopoda:
I know you didn't introduce this code, but could you make it so that it is possible to pass an optional initial path to the files browser bloc instead of having to call setPath directly after it?

_Originally posted by @provokateurin in https://github.com/nextcloud/neon/pull/556#discussion_r1292826379_

Leptopoda:
I hate all the path handling.

I was thinking to write a custom FilesUri class that implements uri and gives us these features (like removing trailing `/`, joining path + filename ...

What do you think? This is more a general question that would fit with #568

_Originally posted by @Leptopoda in https://github.com/nextcloud/neon/pull/600#discussion_r1305537413_

PR #600, opened 2023-08-24
Feature/sync

provokateurin:
Closes https://github.com/nextcloud/neon/issues/21

Still missing
- [ ] Settings for mappings: Auto sync, decide for each upload/download etc.
- [ ] Auto sync (not just auto status update)
- [ ] Native file watching (Android background)
- [ ] Optional: Display sync status in the files browser

@Leptopoda early review would be nice if you have the time.
Probably contains a bunch of stuff that should be done in separate PRs, let me know where I should split out stuff (I'll try to do that myself already).


packages/synchronize/lib/synchronize.dart
```diff
@@ -0,0 +1,14 @@
+library synchronize;
+
+import 'package:collection/collection.dart';
+import 'package:json_annotation/json_annotation.dart';
+import 'package:meta/meta.dart';
+
+part 'src/action.dart';
```
Leptopoda, line 7: I'm still not a fan of part directives in library exports

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
+part of '../synchronize.dart';
+
+/// The source the sync uses to sync from and to.
+abstract class SyncSource<T1, T2> {
```
Leptopoda, line 4:
```suggestion
abstract interface class SyncSource<T1, T2> {
```

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Writes the given [object].
+  Future<SyncObject<T1>> writeObject(final SyncObject<T2> object);
+
+  /// Deletes the given [object].
+  Future deleteObject(final SyncObject<T1> object);
+}
+
+/// The sources the sync uses to sync from and to.
+abstract class SyncSources<T1, T2> {
```
Leptopoda, line 21:
```suggestion
abstract interface class SyncSources<T1, T2> {
```

packages/synchronize/lib/src/journal_entry.dart
```diff
@@ -0,0 +1,44 @@
+part of '../synchronize.dart';
+
+/// Stores a single entry in the [SyncJournal].
+///
+/// It contains an [id] and ETags for each object, [etagA] and [etagB] respectively.
+@immutable
+@JsonSerializable()
```
Leptopoda, line 7: This is the reason I don't like the part stuff.

I was just searching for the generated file and forgot that it's all in one.

Leptopoda (commented):
A few initial structure nitpicks (leftovers of our old code).

I like the general code structure/style but haven't looked ad the logic of everything.

packages/synchronize/lib/src/object.dart
```diff
@@ -0,0 +1,22 @@
...
+  final String id;
+
+  /// Actual data of the object, can be anything.
+  final T data;
+}
+
+// ignore: public_member_api_docs
+extension SyncObjectsFind<T> on List<SyncObject<T>> {
+  // ignore: public_member_api_docs
+  SyncObject<T>? find(final String id) => firstWhereOrNull((final object) => object.id == id);
```
Leptopoda, line 21:
```suggestion
  SyncObject<T>? tryFind(final String id) => firstWhereOrNull((final object) => object.id == id);
```

We only use find when the result is non nullable

packages/synchronize/lib/src/journal_entry.dart
```diff
@@ -0,0 +1,44 @@
...
+  bool operator ==(final Object other) => other is SyncJournalEntry && other.id == id;
+
+  @override
+  int get hashCode => id.hashCode;
+}
+
+// ignore: public_member_api_docs
+extension SyncJournalEntriesFind on Iterable<SyncJournalEntry> {
+  // ignore: public_member_api_docs
+  SyncJournalEntry? find(final String id) => firstWhereOrNull((final entry) => entry.id == id);
```
Leptopoda, line 43:
```suggestion
  SyncJournalEntry? tryFind(final String id) => firstWhereOrNull((final entry) => entry.id == id);
```

packages/synchronize/lib/src/object.dart
```diff
@@ -0,0 +1,22 @@
+part of '../synchronize.dart';
+
+/// Wraps the actual data contained on each side.
+class SyncObject<T> {
```
Leptopoda, line 4: I always think that classes with only two members can also be represented as records.
```dart
typedef SyncObject<T> = (String, T);
```

What do you think? I'm not sure whether this makes sense here

packages/synchronize/lib/src/journal.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  /// Update an [entry].
+  void updateEntry(final SyncJournalEntry entry) {
+    for (var i = 0; i < entries.length; i++) {
+      if (entries.elementAt(i) == entry) {
+        entries.remove(entries.elementAt(i));
+        break;
+      }
+    }
+    entries.add(entry);
```
Leptopoda, line 30: Isn't this the same as:

```suggestion
   entries..remove(entry)..add(entry);
```

packages/synchronize/lib/src/action.dart
```diff
@@ -0,0 +1,40 @@
+part of '../synchronize.dart';
+
+/// Action to be executed in the sync process
+abstract class SyncAction<T1, T2> {}
```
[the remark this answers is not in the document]
Leptopoda: As they all have an object this should be part of the interface.

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,263 @@
...
+    } else if (action is SyncActionDeleteFromB<T1, T2>) {
+      removeActions.add(action);
+    } else {
+      throw Exception('illegal action for sorting');
+    }
+  }
+  return [
+    ..._innerSortActions(addActions),
+    ..._innerSortActions(removeActions).reversed,
+  ];
```
Leptopoda, line 245:
```suggestion
  return _innerSortActions(addActions)..addAll(_innerSortActions(removeActions).reversed);
```

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,263 @@
...
+    } else if (action is SyncActionWriteToB<T1, T2>) {
+      addActions.add(action);
+    } else if (action is SyncActionDeleteFromA<T1, T2>) {
+      removeActions.add(action);
+    } else if (action is SyncActionDeleteFromB<T1, T2>) {
+      removeActions.add(action);
+    } else {
+      throw Exception('illegal action for sorting');
+    }
+  }
```
Leptopoda, line 241: Once SyncAction is sealed you should use a switch statement. They are guaranteed to be exhaustive.

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
+part of '../synchronize.dart';
+
+/// The source the sync uses to sync from and to.
+abstract class SyncSource<T1, T2> {
+  /// List all the objects.
+  Future<List<SyncObject<T1>>> listObjects();
```
Leptopoda, line 6:
```suggestion
  FutureOr<List<SyncObject<T1>>> listObjects();
```

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Calculates the ETag of a given [object].
+  ///
+  /// Should be something easy to compute like the mtime of a file and preferably not the hash of the whole content in order to be fast
+  Future<String> getObjectETag(final SyncObject<T1> object);
```
Leptopoda, line 11:
```suggestion
  FutureOr<String> getObjectETag(final SyncObject<T1> object);
```

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Writes the given [object].
+  Future<SyncObject<T1>> writeObject(final SyncObject<T2> object);
```
Leptopoda, line 14:
```suggestion
  FutureOr<SyncObject<T1>> writeObject(final SyncObject<T2> object);
```

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,36 @@
...
+
+  /// Deletes the given [object].
+  Future deleteObject(final SyncObject<T1> object);
```
Leptopoda, line 17:
```suggestion
  FutureOr deleteObject(final SyncObject<T1> object);
```

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,263 @@
...
+                objectA: conflict.objectA,
+                objectB: conflict.objectB,
+                skipped: true,
+              ),
+            );
+          }
+          break;
+      }
+    } else {
+      unsolvedConflicts.add(conflict);
```
Leptopoda, line 216: add this as a `case null:` to the switch

packages/synchronize/lib/src/sources.dart
provokateurin: Not possible because interface classes can't be extended outside the package

packages/synchronize/lib/src/sources.dart
provokateurin: Same as above

packages/synchronize/lib/src/sources.dart
Leptopoda: Ah yes I've run into this before :(

I wonder how it's possible to extend dart interfaces.
```dart
class CustomUri implements Uri{}
```

is valid code :thinking:

provokateurin: Maybe it's just allowed for the sdk?

Leptopoda: RTFM 🤦‍♀️
https://dart.dev/language/class-modifiers#interface

We can't extend it but we can implement it.
Makes sense xD

packages/neon/neon/lib/src/sync/mapping.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  final String appId;
+
+  final SyncJournal journal;
+
+  String get title;
+  String get subtitle;
+
+  /// This method can be implemented to watch local or remote changes and update the status accordingly.
+  void watch(final Function() onUpdated) {}
```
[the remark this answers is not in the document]
provokateurin: I agree, although the mapping is already in scope because you call the watch method on it.

packages/synchronize/lib/src/object.dart
provokateurin: I think it makes a lot of sense. This type is only used to hold some data and has no logic so a class is not necessary.

provokateurin: For SyncAction it's better to have it as the sealed class for exhaustive matching

packages/synchronize/lib/src/journal.dart
provokateurin: Yep :woman_facepalming:

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,245 @@
...
+      journal.updateEntry(
+        SyncJournalEntry(
+          action.object.id,
+          await sources.sourceA.getObjectETag(action.object),
+          await sources.sourceB.getObjectETag(objectB),
+        ),
+      );
+    }
+  }
+}
```
Leptopoda, line 80: this could now also become a switch case

packages/neon/neon_files/lib/sync/sources.dart
```diff
@@ -0,0 +1,161 @@
...
+
+  final props = WebDavPropWithoutValues.fromBools(
+    davgetetag: true,
+    davgetlastmodified: true,
+    nchaspreview: true,
+    ocsize: true,
+    ocfavorite: true,
+  );
+
+  String _path(final SyncObject object) => [...webdavBaseDir, object.id].join('/');
```
Leptopoda, line 47: I hate all the path handling.

I was thinking to write a custom FilesUri class that implements uri and gives us these features (like removing trailing `/`, joining path + filename ...

What do you think? This is more a general question that would fit with #568

Leptopoda (commented):
Looks good overall. I'll take a closer look and test this later.

packages/neon/neon_files/lib/sync/sources.dart
provokateurin: Yeah, it's pretty ugly... A helper class or something like that would be pretty useful

provokateurin:
@Leptopoda if you have better names for the new types please let me know. I feel I'm not good at choosing/inventing new names

provokateurin:
I've put the optional items as follow ups to tackle them later after the basic functionality is done.

provokateurin:
@Leptopoda Would it make sense to move the webdav<->io sync implementation to the nextcloud package? I don't like that it adds another dependency, but people might want to use it without depending on the whole neon package. Maybe a nextcloud_sync package would be more fitting? OTOH when we add other sync things that depend on platform specific stuff (not just io) then they can't be in the nextcloud package.

Leptopoda:
I like the current way :)

packages/neon/neon/lib/src/pages/app_implementation_settings.dart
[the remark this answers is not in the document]
Leptopoda: It might also make sense to add a `OptionsCollectionSettings` widget that handles the AppBar with the reset button although that might also be unnecessary

packages/neon/neon/lib/src/widgets/resolve_sync_conflicts_dialog.dart
```diff
@@ -0,0 +1,150 @@
...
+                child: Text(AppLocalizations.of(context).actionPrevious),
+              ),
+            ] else ...[
+              OutlinedButton(
+                onPressed: () {
+                  Navigator.of(context).pop();
+                },
+                child: Text(AppLocalizations.of(context).actionCancel),
+              ),
+            ],
```
Leptopoda, line 80:
```suggestion
bool get isLast => index == widget.conflicts.conflicts.length;
bool get isFirst => index == 0;

void movePrevious() {
  if (isFirst){
    if (_solutions.length != widget.conflicts.conflicts.length) {
      return;
    }
    Navigator.of(context).pop(_solutions);
  } else {
    setState(() {
      index--;
    });
  }
}

[ ... ]

              OutlinedButton(
                onPressed: movePrevious,
                child: Text( isFirst? 
                  AppLocalizations.of(context).actionCancel
                  : AppLocalizations.of(context).actionPrevious,
                ),
              ),
```

I hope this is understandable :)

packages/neon/neon/lib/src/widgets/resolve_sync_conflicts_dialog.dart
```diff
@@ -0,0 +1,150 @@
...
+        child: Container(
+          padding: const EdgeInsets.all(24),
+          constraints: NeonDialogTheme.of(context).constraints,
+          child: body,
+        ),
+      ),
+    );
+  }
+
+  Widget _buildConflictCard({
```
Leptopoda, line 117: Make this a separate widget please.
Should be more performant.

packages/neon/neon/lib/src/widgets/sync_status_icon.dart
```diff
@@ -0,0 +1,29 @@
...
+  final SyncMappingStatus status;
+  final double? size;
+
+  @override
+  Widget build(final BuildContext context) {
+    final (icon, color) = switch (status) {
+      SyncMappingStatus.unknown => (MdiIcons.cloudQuestion, NcColors.error),
+      SyncMappingStatus.incomplete => (MdiIcons.cloudSync, NcColors.warning),
+      SyncMappingStatus.complete => (MdiIcons.cloudCheck, NcColors.success),
+    };
```
Leptopoda, line 22: cosmetic stuff but a newline between the return and the assignment would be nice

packages/neon/neon_files/lib/sync/implementation.dart
```diff
@@ -0,0 +1,82 @@
+part of '../neon_files.dart';
+
+class FilesSync implements SyncImplementation<FilesSyncMapping, WebDavFile, FileSystemEntity> {
+  @override
+  String appId = AppIDs.files;
```
Leptopoda, line 5:
```suggestion
  final String appId = AppIDs.files;
```

packages/neon/neon_files/lib/sync/mapping.dart
```diff
@@ -0,0 +1,64 @@
...
+  final String localPath;
+
+  @override
+  late String title = remotePath.join('/');
+
+  @override
+  late String subtitle = localPath;
+
+  @override
+  late String id = '${Uri.encodeComponent(remotePath.join('/'))}-${Uri.encodeComponent(localPath)}';
```
Leptopoda, line 45:
```suggestion
  late String id = '${Uri.encodeComponent(title)}-${Uri.encodeComponent(localPath)}';
```

packages/neon/neon_files/lib/widgets/file_tile.dart
```diff
@@ -0,0 +1,89 @@
...
+                        ],
+                      )
+                    : FilePreview(
+                        bloc: filesBloc,
+                        details: details,
+                        withBackground: true,
+                        borderRadius: const BorderRadius.all(Radius.circular(8)),
+                      ),
+              ),
+              if (details.isFavorite ?? false) ...[
```
Leptopoda, line 74: Stacks can be expensive. I'd like to use them conditionally:

```dart
Widget icon = Center( ... );

if(details.isFavorite){
icon = Stack(
  children [
    icon,
    starOverlay,
  ]
);
}

// use the icon as needed
```

packages/synchronize/lib/src/action.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:meta/meta.dart';
+import 'package:synchronize/src/object.dart';
+
+/// Action to be executed in the sync process
+@internal
+sealed class SyncAction<T> {
```
Leptopoda, line 6: could be immutable couldn't it?

packages/synchronize/lib/src/journal.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  // ignore: public_member_api_docs
+  factory SyncJournal.fromJson(final Map<String, dynamic> json) => _$SyncJournalFromJson(json);
+  // ignore: public_member_api_docs
+  Map<String, dynamic> toJson() => _$SyncJournalToJson(this);
+
+  /// All entries contained in the journal.
+  final Set<SyncJournalEntry> entries;
+
+  /// Update an [entry].
```
Leptopoda, line 25:
```suggestion
  /// Updates an [entry].
```

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,39 @@
+import 'dart:async';
+
+import 'package:synchronize/src/conflict.dart';
+import 'package:synchronize/src/object.dart';
+
+/// The source the sync uses to sync from and to.
+abstract interface class SyncSource<T1, T2> {
```
Leptopoda, line 7: could be immutable

packages/synchronize/lib/src/sources.dart
```diff
@@ -0,0 +1,39 @@
...
+
+  /// Writes the given [object].
+  FutureOr<SyncObject<T1>> writeObject(final SyncObject<T2> object);
+
+  /// Deletes the given [object].
+  FutureOr deleteObject(final SyncObject<T1> object);
+}
+
+/// The sources the sync uses to sync from and to.
+abstract interface class SyncSources<T1, T2> {
```
Leptopoda, line 24: immutable and interfaces shouldn't have a constructor

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,246 @@
...
+import 'package:synchronize/src/sources.dart';
+
+/// Sync between two [SyncSources]s.
+///
+/// This implementation follows https://unterwaditzer.net/2016/sync-algorithm.html in a generic and abstract way
+/// and should work for any two kinds of sources and objects.
+Future<List<SyncConflict<T1, T2>>> sync<T1, T2>(
+  final SyncSources<T1, T2> sources,
+  final SyncJournal journal,
+  final Map<String, SyncConflictSolution> conflictSolutions, {
```
Leptopoda, line 15: looks like the solution is more or less optional as wherever we don't have one an empty map is passed around.
I'd rather like the parameter to be optional in the first place.

This also applies to all other methods

packages/synchronize/lib/src/journal.dart
```diff
@@ -0,0 +1,31 @@
...
+part 'journal.g.dart';
+
+/// Contains the journal.
+///
+/// Used for detecting changes and new or deleted files.
+@JsonSerializable()
+class SyncJournal {
+  // Note: This must not be const as otherwise the entries are not modifiable when a const set is used!
+  // ignore: public_member_api_docs
+  SyncJournal(
```
Leptopoda, line 13: I don't know how much it makes sense but a journal is probably always created empty.

I therefore suggest adding a named constructor like `SyncJournal.empty()` that initializes the the `entries` with an empty map.

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,246 @@
...
+Future executeSyncDiff<T1, T2>(
+  final SyncSources<T1, T2> sources,
+  final SyncJournal journal,
+  final SyncDiff<T1, T2> diff,
+) async {
+  for (final action in diff.actions) {
+    switch (action) {
+      case SyncActionDeleteFromA():
+        await sources.sourceA.deleteObject(action.object as SyncObject<T1>);
+        journal.entries.removeWhere((final entry) => entry.id == action.object.id);
```
Leptopoda, line 57: In theory we don't need to add the comparison as the `SyncJournalEntry` already overrides the euqality check by checking the id

```suggestion
        journal.entries.remove(action.object);
```

Leptopoda (changes requested):
Only tested the basics yet and didn't check all the doc comments for style/correctness.

I also think that a lot of interfaces can still be improved with more doc comments and it would be great to add unit tests to the syncBloc (I'll also need to check the coverage).

Overall just a few smaller cleanups and I really love the UI in the settings. It's just beautiful

packages/neon/neon/lib/src/pages/app_implementation_settings.dart
```diff
@@ -0,0 +1,183 @@
...
+                          .where((final mappingStatus) => mappingStatus.key.appId == appImplementation.id)) ...[
+                        CustomSettingsTile(
+                          title: Text(mappingStatus.key.title),
+                          subtitle: Text(mappingStatus.key.subtitle),
+                          leading: NeonUserAvatar(
+                            account: accountsBloc.accounts.value
+                                .singleWhere((final account) => account.id == mappingStatus.key.accountId),
+                            showStatus: false,
+                          ),
+                          trailing: IconButton(
```
Leptopoda, line 98: IconButtons should have a tooltip for a11y

packages/neon/neon/lib/src/blocs/sync.dart
```diff
@@ -0,0 +1,274 @@
...
+  SyncImplementation _getImplementationForMapping(final SyncMapping mapping) =>
+      _syncImplementations.singleWhere((final implementation) => implementation.appId == mapping.appId);
+
+  void _loadMappings() {
+    debugPrint('Loading mappings');
+    final loadedMappings = <SyncMapping>[];
+
+    if (_storage.containsKey(_keyMappings)) {
+      final serializedMappings = (json.decode(_storage.getString(_keyMappings)!) as Map<String, dynamic>)
+          .map((final key, final value) => MapEntry(key, (value as List).map((final e) => e as Map<String, dynamic>)));
```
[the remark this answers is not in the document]
provokateurin: This is what I hate about Dart type casting. You can't simply cast this stuff directly but you need to do it manually on every level :/

packages/neon/neon/lib/src/blocs/sync.dart
```diff
@@ -0,0 +1,274 @@
...
+    if (_watchControllers[mapping.id] != null) {
+      return;
+    }
+
+    // ignore: close_sinks
+    final controller = StreamController();
+    // Debounce is required to stop bulk operations flooding the sync and potentially creating race conditions.
+    controller.stream.debounceTime(const Duration(seconds: 1)).listen((final _) async {
+      await _updateMapping(mapping);
+    });
```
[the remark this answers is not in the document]
provokateurin: Nope no periodic. This basically filters all events that happen until one second has passed and then only sends of one event. That way we don't get flooded e.g. when you drop 20 files at once into a folder.

packages/neon/neon/lib/src/blocs/sync.dart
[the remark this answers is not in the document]
Leptopoda: Ah I thought that might be the problem.
Can we at least use the `.cast()` method that iterables (list and map) provide? I think that might be cleaner.

packages/neon/neon/lib/src/blocs/sync.dart
```diff
@@ -0,0 +1,274 @@
...
+
+    _watchControllers[mapping.id] = controller;
+
+    mapping.watch(() async {
+      controller.add(null);
+    });
+  }
+
+  SyncMappingOptions getSyncMappingOptionsFor(final SyncMapping mapping) {
+    final id = '${mapping.accountId}-${mapping.appId}-${mapping.id}';
```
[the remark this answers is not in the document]
provokateurin: It already has, that's the problem. This id is basically the global unique id, shall I use that as the name?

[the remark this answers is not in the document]
Leptopoda: Yes please

packages/synchronize/lib/src/journal.dart
provokateurin: How about named parameter with default empty value? IMO that is even cleaner

packages/synchronize/lib/src/sync.dart
provokateurin: But `action.object.data` is not a `SyncJournalEntry`, that's why the explicit comparison is needed.

provokateurin:
@Leptopoda thanks for the review. I picked up all the easy changes and will do the more complex ones in the next days.

packages/synchronize/lib/src/sync.dart
```diff
@@ -0,0 +1,246 @@
...
+import 'package:synchronize/src/sources.dart';
+
+/// Sync between two [SyncSources]s.
+///
+/// This implementation follows https://unterwaditzer.net/2016/sync-algorithm.html in a generic and abstract way
+/// and should work for any two kinds of sources and objects.
+Future<List<SyncConflict<T1, T2>>> sync<T1, T2>(
+  final SyncSources<T1, T2> sources,
+  final SyncJournal journal, {
+  final Map<String, SyncConflictSolution> conflictSolutions = const {},
```
Leptopoda, line 15: I was more thinking of making it nullable as I think const collections can be dangerous.
I can see me 100% calling conflictSolutions.add down the road and wondering why it's not working.

I briefly tested it this morning and it should be possible with null.

packages/synchronize/lib/src/journal.dart
Leptopoda: also possible.
But as the set can't be null it will probably have to be something like:
```dart
SyncJournal({
 Set<SyncJournalEntry>? entries, 
}): entries = entries ?? {};
```

packages/synchronize/lib/src/sync.dart
provokateurin: Ah true, that could be a problem indeed

packages/neon/neon/lib/src/blocs/sync.dart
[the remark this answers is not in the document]
provokateurin: Not applicable here because the casts are from `dynamic` and thus `.cast()` is not available (since the objects aren't iterables or maps already).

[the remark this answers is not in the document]
Leptopoda: I meant
```dart
(json.decode(_storage.getString(_keyMappings)!) as Map).cast<String, List<Map<String, dynamic>>>();
```
or 
```dart
(json.decode(_storage.getString(_keyMappings)!) as Map<String, dynamic>)
          .map((final key, final value) => MapEntry(key, (value as List).cast<Map<String, dynamic>()));
```

packages/neon/neon/lib/src/pages/app_implementation_settings.dart
provokateurin: Also added labels to the icon itself.

provokateurin:
Based on https://github.com/nextcloud/neon/pull/698

Leptopoda (commented):
Code style is good (nothing obvious left).

I just tested the conflict resolution and I have a few concerns:
1. there is no UI indicating a conflict. The `Sync now` button should get red with an error icon.
2. there should be a list of all failed syncs easily accessible (I suppose in the sync settings). Followup expected.
3. The left TextButton of the conflict resolution dialog (showing previous and cancel) always has an outline. while the right one doesn't. I'd like that only the cancel and finish ones to have those while previous/next should stay not outlined.
4. the local/remote text on the SyncConflictCard could do with some padding.
5. the Card widget within the SyncConflictCard should have the `clip` property set to properly clip the splash of it's InkWell (I suppose Clip.hardEdge is enough)
6. the SyncConflictDialog should not be shown again for files I've already skipped by hitting next or cancel/finish. Skipping something means I don't want to deal with it right away but "Sync automatically" pushes the dialog every few seconds.
7. related to (6) if I have a conflict and "Sync automatically" enabled the automatic pushes will push an infinite number of dialogs over time. This makes the app unusable.
8. If I disable the "Sync automatically" option, create a conflicting file and re enable it the sync isn't made automatically. In fact the manual sync also doesn't work until I hot restart.

provokateurin:
I agree with your concerns, I will try to address them soon

provokateurin:
We should implement one way sync (for both directions)

Leptopoda:
> We should implement one way sync (for both directions)

Is this related to what we discussed for #779

provokateurin:
No I thought of this for other reasons, but it will help with that as well

provokateurin:
(Just rebased to solve all conflicts and new linting problems)

Leptopoda:
I think you missed it but I got some great feedback from simon and nimsha. I'll post them tomorrow.

provokateurin:
Yes I had to go when the topic came up, so please let me know what the ideas where :)

provokateurin:
Maybe it would make more sense to wait for our plugin architecture to properly encapsulate this whole feature?

Leptopoda:
Whatever you want but I agree that it's probably not that important for the 1.0

Regarding the design feedback:
Simon was really against burriing it in the settings and said that we could either display it on the notifications page (probably also involving the unread counter).
Another good idea was to add a static tile at the top of the files list similar to what we do when uploading or downloading files. Pressing this tile would then open a more detailed page similar to what the NC android client does.

provokateurin:
Another TODO: Use better methods for generating ETags on the client, since currently it just hashes the file content which is really bad. The Desktop client has some documentation about the way they do it (and it should work us as well).

provokateurin:
@Leptopoda can you note down the feedback from the design meeting?

Leptopoda:
I can write it down on the weekend/next week.

provokateurin:
@Leptopoda it would be great to have the notes from the meeting :)

I think we should introduce a "sync" entry in the drawer like the android app does as it is currently hidden in the depths of the settings. That entry could lead to a new page that shows the current queue and has the ability to add new sync mappings and remove them as well.
I can't remember much, but we might have discussed this idea already at the meeting :woman_shrugging:

provokateurin:
Fixed the long lists of conflicts so it is now up-to-date again. I didn't give it any testing.

provokateurin:
Unit tests for the webdav/io sync source need to be added.

Leptopoda:
> @Leptopoda it would be great to have the notes from the meeting :)
 
#946

provokateurin:
Ah I forgot about that and it wasn't linked here.

I llike the direct integration into the files app, but what do you think about my idea with the separate section for sync? I like that there is a centralized place to check for there current queue (regardless of which app is syncing something, but only files for now) and to add new sync mappings. Instead of burying it in the app settings it would be pretty visible. I fear some people would never find out about sync support if they don't check through all the settings.

PR #603, opened 2023-08-25
Refactor/nextcloud uri

Leptopoda:
The rest of #592

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
@@ -3,6 +3,7 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:cookie_jar/cookie_jar.dart';
+import 'package:dynamite_runtime/src/uri.dart';
 import 'package:universal_io/io.dart';
 
 export 'package:cookie_jar/cookie_jar.dart';
@@ -110,15 +111,16 @@ class DynamiteHttpBearerAuthentication extends DynamiteAuthentication {
 
 class DynamiteClient {
   DynamiteClient(
-    this.baseURL, {
+    final Uri baseURL, {
     this.baseHeaders,
     final String? userAgent,
     final HttpClient? httpClient,
     this.cookieJar,
     this.authentications = const [],
-  }) : httpClient = (httpClient ?? HttpClient())..userAgent = userAgent;
+  })  : httpClient = (httpClient ?? HttpClient())..userAgent = userAgent,
+        baseURL = baseURL.normalizeEmptyPath();
 
-  final String baseURL;
+  final Uri baseURL;
 
   final Map<String, String>? baseHeaders;
 
--- a/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/http_client.dart
@@ -132,11 +132,18 @@ class DynamiteClient {
 
   Future<HttpClientResponse> doRequest(
     final String method,
-    final String path,
+    final Uri path,
     final Map<String, String> headers,
     final Uint8List? body,
   ) async {
-    final uri = Uri.parse('$baseURL$path');
+    final uri = baseURL.replace(
+      path: '${baseURL.path}${path.path}',
+      queryParameters: {
+        ...baseURL.queryParameters,
+        ...path.queryParameters,
+      },
+    );
+
     final request = await httpClient.openUrl(method, uri);
     for (final header in {...?baseHeaders, ...headers}.entries) {
       request.headers.add(header.key, header.value);
--- a/packages/neon/neon/lib/src/models/account.dart
+++ b/packages/neon/neon/lib/src/models/account.dart
@@ -13,7 +13,7 @@ part 'account.g.dart';
 @immutable
 abstract interface class Credentials {
   /// Url of the server
-  abstract final String serverURL;
+  abstract final Uri serverURL;
 
   /// Username
   abstract final String username;
@@ -44,7 +44,7 @@ class Account implements Credentials {
   Map<String, dynamic> toJson() => _$AccountToJson(this);
 
   @override
-  final String serverURL;
+  final Uri serverURL;
   @override
   final String username;
   @override
@@ -70,19 +70,18 @@ class Account implements Credentials {
     return _idCache[key] ??= sha1.convert(utf8.encode(key)).toString();
   }
 
+  /// A human readable representation of [username] and [serverURL].
   String get humanReadableID {
-    final uri = Uri.parse(serverURL);
-
     // Maybe also show path if it is not '/' ?
     final buffer = StringBuffer()
       ..write(username)
       ..write('@')
-      ..write(uri.host);
+      ..write(serverURL.host);
 
-    if (uri.hasPort) {
+    if (serverURL.hasPort) {
       buffer
         ..write(':')
-        ..write(uri.port);
+        ..write(serverURL.port);
     }
 
     return buffer.toString();
@@ -96,7 +95,7 @@ class Account implements Credentials {
   /// The paths of the [serverURL] and the [uri] need to be join to get the full path, unless the [uri] path is already an absolute path.
   /// In that case an instance hosted at a sub folder will already contain the sub folder part in the [uri].
   Uri completeUri(final Uri uri) {
-    final result = Uri.parse(serverURL).replace(
+    final result = serverURL.replace(
       queryParameters: uri.queryParameters,
       fragment: uri.fragment,
     );
@@ -108,7 +107,7 @@ class Account implements Credentials {
   /// Removes the [serverURL] part from the [uri].
   ///
   /// Should be used when trying to push a [uri] from an API to the router as it might contain the scheme, host and sub path of the instance which will not work with the router.
-  Uri stripUri(final Uri uri) => Uri.parse(uri.toString().replaceFirst(serverURL, ''));
+  Uri stripUri(final Uri uri) => Uri.parse(uri.toString().replaceFirst(serverURL.toString(), ''));
 }
 
 Map<String, String> _idCache = {};
@@ -134,7 +133,7 @@ class LoginQrcode implements Credentials {
   });
 
   @override
-  final String serverURL;
+  final Uri serverURL;
   @override
   final String username;
   @override
@@ -176,7 +175,7 @@ class LoginQrcode implements Credentials {
       }
 
       return LoginQrcode(
-        serverURL: match.group(3)!,
+        serverURL: Uri.parse(match.group(3)!),
         username: match.group(1)!,
         password: match.group(2)!,
       );
--- a/packages/neon/neon/lib/src/pages/login.dart
+++ b/packages/neon/neon/lib/src/pages/login.dart
@@ -30,7 +30,9 @@ class _LoginPageState extends State<LoginPage> {
 
   void login(final String url) {
     if (_formKey.currentState!.validate()) {
-      LoginCheckServerStatusRoute(serverUrl: url).go(context);
+      final uri = Uri.parse(url);
+
+      LoginCheckServerStatusRoute(serverUrl: uri).go(context);
     } else {
       _focusNode.requestFocus();
     }
--- a/packages/neon/neon/lib/src/utils/validators.dart
+++ b/packages/neon/neon/lib/src/utils/validators.dart
@@ -9,8 +9,9 @@ String? validateHttpUrl(
   if (input == null || input.isEmpty) {
     return AppLocalizations.of(context).errorInvalidURL;
   }
-  try {
-    final uri = Uri.parse(input);
+  final uri = Uri.tryParse(input);
+
+  if (uri != null) {
     if (uri.isScheme('https')) {
       return null;
     }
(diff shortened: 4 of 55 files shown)


packages/dynamite/dynamite_runtime/lib/src/http_client.dart
```diff
@@ -134,7 +135,8 @@ class DynamiteClient {
     final Map<String, String> headers,
     final Uint8List? body,
   ) async {
-    final uri = Uri.parse('$baseURL$path');
+    final uri = (UriBuilder.fromUri(baseURL)..path = '${baseURL.path}$path').build();
```
provokateurin, line 138: Why not just use `Uri.replace`? I can see how this is nicer to work with, but I don't see it as necessary (at least in this case).

packages/neon/neon/lib/src/pages/login.dart
```diff
@@ -31,7 +31,9 @@ class _LoginPageState extends State<LoginPage> {
 
   void login(final String url) {
     if (_formKey.currentState!.validate()) {
-      LoginCheckServerStatusRoute(serverUrl: url).go(context);
+      final uri = Uri.parse(url.endsWith('/') ? url.substring(0, url.length - 1) : url);
```
provokateurin, line 34: I think creating an extension for Uri that removes the trailing slash would make this a lot better.

packages/neon/neon/lib/src/utils/validators.dart
```diff
@@ -9,16 +9,14 @@ String? validateHttpUrl(
...
-    if (uri.isScheme('http') && !httpsOnly) {
-      // TODO: Maybe make a better error message for http URLs if only https is allowed
-      return null;
-    }
-  } catch (_) {}
+  final uri = Uri.tryParse(input);
+  if (uri != null && uri.isScheme('https')) {
+    return null;
+  }
+  if (uri != null && uri.isScheme('http') && !httpsOnly) {
```
provokateurin, line 16: Unify the null check

packages/nextcloud/lib/src/client.dart
```diff
@@ -26,7 +26,8 @@ class NextcloudClient extends DynamiteClient {
     final AppType appType = AppType.unknown,
     final String? userAgentOverride,
     super.cookieJar,
-  }) : super(
+  })  : assert(!baseURL.path.endsWith('/'), "The baseURL should not end with a '/'"),
```
provokateurin, line 29: Should we move this check into the generated dynamite code?

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -62,24 +61,27 @@ class WebDavClient {
...
+
+    final pathBuffer = StringBuffer(baseURL.path)..write(webdavBasePath);
+    if (path != null && path.path != '/') {
+      pathBuffer
+        ..write('/')
+        ..write(path.path);
+    }
+
+    final builder = UriBuilder.fromUri(baseURL)..path = pathBuffer.toString();
+    return builder.build();
```
provokateurin, line 83: Again, just using `Uri.replace` would work as well.

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/http_client.dart
Leptopoda: AFAICT this will lead to side effects mutating the baseURL object.

Or am I wrong :thinking:

provokateurin: No you get a new Uri returned

packages/nextcloud/lib/src/client.dart
Leptopoda: I think we should rather move it into the dynamite client. And all other clients would "inherit" the assertion.

packages/neon/neon/lib/src/pages/login.dart
Leptopoda: I'll leave this open for now and we'll see once the mentioned FilesUri becomes a thing

provokateurin (approved):

provokateurin:
Please wait for https://github.com/nextcloud/neon/pull/581

packages/dynamite/dynamite_runtime/lib/src/http_client.dart
```diff
@@ -116,9 +116,11 @@ class DynamiteClient {
     final HttpClient? httpClient,
     this.cookieJar,
     this.authentications = const [],
-  }) : httpClient = (httpClient ?? HttpClient())..userAgent = userAgent;
+  })  : httpClient = (httpClient ?? HttpClient())..userAgent = userAgent,
+        assert(!baseURL.path.endsWith('/'), "The baseURL should not end with a '/'"),
+        assert(baseURL.queryParameters.isEmpty, 'The baseURL can not contain query parameters.');
```
provokateurin, line 121: Is there any reason why we should not allow it? I don't think it could break anything and maybe this is useful somewhere else? :woman_shrugging:

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/http_client.dart
```diff
@@ -130,11 +132,12 @@ class DynamiteClient {
 
   Future<HttpClientResponse> doRequest(
     final String method,
-    final String path,
+    final Uri path,
```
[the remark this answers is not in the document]
provokateurin: Same throughout the webdav code, although it makes more sense there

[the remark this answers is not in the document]
Leptopoda: I wanted to keep it to indicate it shouldn't contain a scheme/host although they would just get ignored.

Maybe the planned FilesUri should be a PathsUri that can only handle a path and a queryParam.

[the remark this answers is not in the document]
provokateurin: Sounds good

provokateurin (approved):

provokateurin:
Tests were stuck with your new contructUri test. Is there something wrong with them (infinite loop maybe)?

Leptopoda:
```dart
  test('constructUri', () {
    var baseURL = Uri.parse('http://cloud.example.com');
    expect(WebDavClient.constructUri(baseURL).toString(), '$baseURL$webdavBasePath');
    expect(WebDavClient.constructUri(baseURL, Uri(path: '/')).toString(), '$baseURL$webdavBasePath');
    expect(WebDavClient.constructUri(baseURL, Uri(path: 'test')).toString(), '$baseURL$webdavBasePath/test');

    baseURL = Uri.parse('http://cloud.example.com/subdir');
    expect(WebDavClient.constructUri(baseURL, Uri(path: 'test')).toString(), '$baseURL$webdavBasePath/test');

    expect(() => WebDavClient.constructUri(baseURL, Uri(path: '/test')), throwsA(isA<AssertionError>()));
    baseURL = Uri.parse('http://cloud.example.com/');
    expect(() => WebDavClient.constructUri(baseURL), throwsA(isA<AssertionError>()));
  });
```

looks like a normal test to me (also finishes quite fast on my side

provokateurin:
Ok let's run again, maybe it was just a hiccup.

provokateurin:
> 2023-08-25T16:05:03.1638456Z nextcloud:
2023-08-25T16:05:03.8827861Z 
2023-08-25T16:05:11.1742404Z ✅ /home/runner/work/neon/neon/packages/nextcloud/test/webdav_test.dart: constructUri
2023-08-25T17:34:22.3073817Z ##[error]The operation was canceled.

There seems to be a real problem :thinking:

Leptopoda:
@provokateurin should we kill it again?
It's already double the expected time.

provokateurin:
Yeah, let's not waist CI time. I'll try it on my end as well

provokateurin:
I find it confusing that the log says the job failed but it still runs

provokateurin:
Ahhh I think the `constructUri` test passes just fine, but then the next test (`List directory`) just runs forever. At least that's what happens on my machine. I just ran the tests individually by name.

Leptopoda:
Thanks for the info; I can repro this on my side.
I'm fairly sure I did run all tests before pushing :thinking: 

I think we should specify timeouts of the tests

provokateurin:
Yes, 30 seconds should be absolutely enough

Leptopoda:
I also added a unified timeout to all tests and unified the retryCount.

packages/nextcloud/test/core_test.dart
```diff
@@ -1,160 +1,162 @@
...
-    });
-
-    test('Delete app password', () async {
-      await client.core.appPassword.deleteAppPassword();
-      expect(
-        () => client.core.appPassword.deleteAppPassword(),
-        throwsA(predicate((final e) => (e! as DynamiteApiException).statusCode == 401)),
+  group(
+    'core',
+    () {
```
provokateurin, line 9: I don't like this formatting change. Doesn't seem necessary to me?

packages/nextcloud/test/helper.dart
```diff
@@ -127,12 +134,18 @@ Future<TestNextcloudClient> getTestClient(
...
   while (true) {
-    // Test will timeout after 30s
     try {
       await client.core.getStatus();
       break;
-    } catch (_) {}
+    } on CoreApiException catch (error) {
+      i++;
+      await Future.delayed(const Duration(microseconds: 100));
+      if (i >= 3) {
```
provokateurin, line 145:
```suggestion
      await Future.delayed(const Duration(milliseconds: 100));
      if (i >= 30) {
```

This is 3 seconds in total. If there really is an infinite loop then that doesn't really matter, but we will have a lot less false positives.

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -72,11 +72,9 @@ class Account implements Credentials {
...
   }
 
-  String get humanReadableID {
-    final uri = Uri.parse(serverURL);
-    // Maybe also show path if it is not '/' ?
-    return '$username@${uri.port != 443 ? '${uri.host}:${uri.port}' : uri.host}';
-  }
+  // Maybe also show path if it is not '/' ?
+  String get humanReadableID =>
+      '$username@${serverURL.port != 443 ? '${serverURL.host}:${serverURL.port}' : serverURL.host}';
```
provokateurin, line 77:
```suggestion
      '$username@${serverURL.host}${serverURL.port != 443 ? ':${serverURL.port}' : ''}';
```

How about also using a string buffer for this? I think that would be a lot cleaner.

packages/nextcloud/test/core_test.dart
provokateurin: Ah I see what you did. I don't like it, but I guess we'll have to live with it.

provokateurin (changes requested):

packages/nextcloud/test/core_test.dart
Leptopoda: We could also specify the timeout at the lib level (similar to what we did with the retry before) but I don't really like having them global.

The new uri test for example doesn't need a retry at all.

packages/nextcloud/test/helper.dart
Leptopoda: You might have noticed that I restricted the catch on dynamite exceptions.

This means that an error caused by false code (as was the problem before) won't even cause a loop anymore.

provokateurin: Yes, but the container might still be slow to start. Apache is really bad with this, that's why most tests use php instead because the startup time is so much better.

packages/neon/neon/lib/src/models/account.dart
```diff
@@ -72,10 +72,24 @@ class Account implements Credentials {
...
+      ..write('@')
+      ..write(serverURL.host);
+
+    if (serverURL.hasPort) {
+      buffer
+        ..write(':')
+        ..write(serverURL.host);
+    }
+
+    if (!serverURL.hasEmptyPath) {
```
provokateurin, line 88: How does this handle `/` as path?

Leptopoda: We already assert in the dynamite client that the path should not end with a `/`

Leptopoda: https://github.com/nextcloud/neon/pull/603/files#diff-8dde2cc07e164437d15007e4dfc0d39721404c1370d1b4008087877d6f0d9c83R120

provokateurin: Hm, now I wonder how the UI will handle it if some puts in a URL with a trailing slash on the login screen

Leptopoda: We allow trailing slashes and manually strip it before passing it into the bloc.

I'll need to check for the loginQrCode though

Leptopoda: We also only assert this meaning that it will get compiled away for production.

We might want to throw an exception

Leptopoda: @provokateurin what do you think?

provokateurin: In our login logic yes, but I wouldn't do it in the nextcloud package.

provokateurin (approved):
Good cleanup :)

PR #604, opened 2023-08-25
Feat/neon files/sort folders

Leptopoda:
- sort folders first (not configurable)
- hide hidden files (configurable)
- cleanup of some widgets

### DIFF
--- a/packages/neon/neon_files/lib/widgets/file_list_tile.dart
+++ b/packages/neon/neon_files/lib/widgets/file_list_tile.dart
@@ -1,7 +1,9 @@
 import 'package:filesize/filesize.dart';
 import 'package:flutter/material.dart';
 import 'package:material_design_icons_flutter/material_design_icons_flutter.dart';
+import 'package:neon/utils.dart';
 import 'package:neon/widgets.dart';
+import 'package:neon_files/l10n/localizations.dart';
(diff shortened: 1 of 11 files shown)


packages/neon/neon_files/lib/widgets/file_list_tile.dart
```diff
@@ -10,66 +12,72 @@ class FileListTile extends StatelessWidget {
...
-              width: 10,
-            ),
-            Text(
-              filesize(details.size, 1),
-              style: DefaultTextStyle.of(context).style.copyWith(
-                    color: Colors.grey,
-                  ),
-            ),
+  @override
+  // When the ETag is null it means we are uploading this file right now
```
provokateurin, line 45: Comment should be one line below

provokateurin (changes requested):
I would be nice to have a warning that shows up when you create a hidden file or folder but you have hidden files not shown. Otherwise the user might get confused why the file/folder was not created.

Leptopoda:
What about doing the warning as a followup (maybe together with https://github.com/nextcloud/neon/issues/535)?

It would currently require adding  the warning to two separate dialogs (folder creation and rename dialog). And I think we should first unify them and them

Leptopoda:
Ah forgot to push one sec

PR #611, opened 2023-08-27
Feature/renovate

provokateurin:
Closes https://github.com/nextcloud/neon/issues/574

This simple config file should already get us very far. I checked locally and it correctly recognizes all packages.
If we find that it doesn't do exactly what we want we can still change it.


Leptopoda (approved):
I don't know anything about renovate but if it works

provokateurin:
Apparently it still needs to be "installed" for this repository. I made the request and hope that someone will accept it.

provokateurin:
Ah someone accepted it!

PR #654, opened 2023-08-29
Refactor/shared preferences

### LINKED ISSUE #609: Make sharedPreferences instance static.
Maybe as a followup and but once rebased I'd love to also remove the sharedPrefs from here.

Especially as we use the same one throughout the app a static field somewhere would also work (maybe in the NeonPlatform).

_Originally posted by @Leptopoda in https://github.com/nextcloud/neon/pull/495#discussion_r1306648141_

Leptopoda:
fixes #609

### DIFF
--- a/packages/app/lib/apps.dart
+++ b/packages/app/lib/apps.dart
@@ -3,14 +3,10 @@ import 'package:neon_files/neon_files.dart';
 import 'package:neon_news/neon_news.dart';
 import 'package:neon_notes/neon_notes.dart';
 import 'package:neon_notifications/neon_notifications.dart';
-import 'package:shared_preferences/shared_preferences.dart';
 
-List<AppImplementation> getAppImplementations(
-  final SharedPreferences sharedPreferences,
-) =>
-    [
-      FilesApp(sharedPreferences),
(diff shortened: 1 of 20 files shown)


packages/app/lib/apps.dart
```diff
@@ -3,14 +3,10 @@ import 'package:neon_files/neon_files.dart';
...
-List<AppImplementation> getAppImplementations(
-  final SharedPreferences sharedPreferences,
-) =>
-    [
-      FilesApp(sharedPreferences),
-      NewsApp(sharedPreferences),
-      NotesApp(sharedPreferences),
-      NotificationsApp(sharedPreferences),
-    ];
+final List<AppImplementation> appImplementations = [
```
provokateurin, line 7: Since there is no logic left I would just inline this

provokateurin (approved):
Just one small nit pick, very nice cleanup!

packages/app/lib/apps.dart
Leptopoda: This isn't possible as we still need it for the screenshot tests

provokateurin (approved):
Wrong formatting

PR #675, opened 2023-08-29
feat(ci): Check generated code

provokateurin:


provokateurin:
@Leptopoda Ok this doesn't make sense. The jobs will take forever to complete and until now the problem hasn't been too huge. If it becomes a big problem in the future then we can try this again.

PR #690, opened 2023-08-31
feat(neon_lints): use lint_maker package

Leptopoda:
lint_maker is just a poc. Ideas to improve it are welcome and I might also publish it on pub.dev in the future.


packages/neon_lints/lib/dart.yaml
```diff
@@ -9,219 +9,219 @@ analyzer:
...
-    use_rethrow_when_possible: true
-    use_setters_to_change_properties: true
-    use_string_buffers: true
-    use_string_in_part_of_directives: true
-    use_super_parameters: true
-    use_test_throws_matchers: true
-    use_to_and_as_if_applicable: true
-    valid_regexps: true
-    void_checks: true
+    - always_declare_return_types: true
```
provokateurin, line 12: Does this work with the `-`? I better like it without :)

packages/neon_lints/lib/flutter.yaml
```diff
@@ -1,9 +1,228 @@
-include: package:neon_lints/dart.yaml
```
provokateurin, line 1: I liked how this was extending the other one. I would like to keep it this way instead of duplicating almost everything.

packages/neon_lints/lib/dart.yaml
Leptopoda: I can have a look but I'm not sure

Leptopoda:
My idea was to add more rules in the future to differentiate between publishable and not publishable packages:
- dart.yaml
- dart_package.yaml
- flutter.yaml
- flutter_package.yaml

and let the package ones have stricter doc requirements (we currently don't require any documentation for flutter packages like neon).
We can still extend lint_maker in the future if we want further custom stuff.

packages/neon_lints/lint_maker.yaml
```diff
@@ -0,0 +1,29 @@
+dart:
+  commitHash: 7f40f11
```
provokateurin, line 2: I'm wondering if there is any way to extract this information from the Dart SDK that currently executes the script. That would be pretty neat (as it would also avoid rules that don't work yet) but it's not very realistic that this is possible in any way :thinking:

Leptopoda: I mean the analyzer must have the ability the question is whether it exposes it

provokateurin: The ability to know from which commit hash it was built? I'm not sure about that

Leptopoda: Not that but what rules are available in the used release.

I mean it has to know it otherwise it couldn't warn us of of the unreleased rules.

Leptopoda:
not directly related but should we enable more rules?
I was thinking of:

- avoid_annotating_with_dynamic
- avoid_catching_errors
- avoid_classes_with_only_static_members

they all only require minor changes as we mostly follow then already.

provokateurin:
Sure, although I don't like avoid_annotating_with_dynamic as you then implicitly type something as dynamic :thinking:

Leptopoda:
Yeah that's bad.
I was thinking about changing them to something like `Object?` but didn't think about the implicit default :(

provokateurin (approved):
Please update to https://github.com/Leptopoda/lint_maker/pull/2 and then we are good to go

PR #701, opened 2023-09-02
Feature/neon lints/package rules

Leptopoda:
as initially discussed in #690


packages/neon_lints/lib/flutter.yaml
```diff
@@ -1,4 +1,5 @@
-include: package:neon_lints/dart.yaml
+include: package:neon_lints/flutter_package.yaml
```
provokateurin, line 1: I think flutter_package should import flutter and not the other way around.

packages/neon_lints/lib/src/base.yaml
```diff
@@ -0,0 +1,227 @@
+analyzer:
+  language:
+    strict-casts: true
+  errors:
+    flutter_style_todos: ignore
+    todo: ignore
```
provokateurin, line 6: I think we should eventually enable these. Leaving instead of leaving todos in the code we should just create issues. Of course you can still use todos during development and then the CI will also be able to warn you about it.

provokateurin (changes requested):
The second commit has a typo in the name

packages/neon_lints/lib/flutter.yaml
Leptopoda: I thought the strictest rules are used in a package and a non package can relax those rules.

Otherwise we'd have strict base -> relax for flutter non package -> tighten them again for flutter package

provokateurin: I didn't think of flutter non package as relaxing the rules but it makes sense

Leptopoda: we can also base all rules on the base rule

provokateurin: Hmm, but then we have duplicate stuff again. Let's just leave it like it is

packages/neon_lints/lib/flutter.yaml
provokateurin, line 1: This should also include the `lib/generated_plugin_registrant.dart` exclusion.

provokateurin (changes requested):

packages/neon_lints/lib/flutter.yaml
```diff
@@ -1,9 +1,6 @@
-include: package:neon_lints/dart.yaml
+include: package:neon_lints/src/base.yaml
+
 linter:
   rules:
     public_member_api_docs: false
     library_private_types_in_public_api: false
-    avoid_print: true
```
provokateurin, line 6: I think we should keep this as one should use `debugPrint` instead.

provokateurin (changes requested):

packages/neon_lints/lib/flutter.yaml
Leptopoda: sure but isn't this already required by the depending on flutter_package.yaml?

provokateurin: Currently it depends on base and not flutter_package

provokateurin (approved):
LGTM, shouldn't we move more packages to flutter_package?

Leptopoda:
more documentation is always better and I'd love to have everything documented one day but for now I think only those published (pub.dev or not) do need it. I also think that we should restructure our apps first (split them into proper packages).

But feel free to open an issue

PR #702, opened 2023-09-03
Fix/screenshots

provokateurin:
The news folders and notifications screenshots are broken for unknown reasons. I just needed to fix the debug banner for the lightning talk slides :sweat_smile:


packages/neon/neon/lib/src/settings/models/storage.dart
```diff
@@ -36,7 +36,6 @@ enum StorageKeys implements Storable {
   final String value;
 }
 
-@internal
```
Leptopoda, line 39: Is this really needed?

The test calls `runNeon` wich should setup the sharedPrefs with the same `await SharedPreferences.getInstance()` you are now doing.

I initally only added this function so we can pass a `Mock` object (from mocktail) into it just to unit test the `SingleValueStorage` and `AppStorage`. I don't think it's needed outside of this.

nishp77:
Could you elaborate here? I tried adding some rss feeds and no issue

provokateurin:
It works fine, but not in our screenshot tests. Nothing to worry about for you

PR #704, opened 2023-09-03
Feature/dynamite/all of

Leptopoda:
finally implements #420 

How it works:
1. We generate an interface class for every `TypeResultObject` as they are the only thing allowed in allOfs.
2. we simply use that interface in the normal class and the allOf class.

Yes this means that we generate interface classes that might only be used in one built_value one. We could traverse all schemes of the spec and find the interfaces we need but I consider this outside the scope of this PR and with more important stuff to work on something that I don't plan to change any time soon.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
@@ -1,5 +1,6 @@
 import 'package:built_collection/built_collection.dart';
 import 'package:code_builder/code_builder.dart';
+import 'package:dynamite/src/builder/resolve_interface.dart';
 import 'package:dynamite/src/builder/resolve_type.dart';
 import 'package:dynamite/src/builder/state.dart';
 import 'package:dynamite/src/helpers/built_value.dart';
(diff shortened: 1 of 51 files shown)


packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
```diff
@@ -8,13 +9,72 @@ import 'package:dynamite/src/models/open_api.dart';
...
+          spec,
+          state,
+          identifier,
+          s,
+          nullable: nullable,
+        );
+
+        if (object is! TypeResultObject) {
+          throw StateError('allOf does only allow objects. Please change $identifier');
+        }
```
provokateurin, line 40: Why not? I think from the spec sind this is totally allowed?

packages/dynamite/dynamite/lib/src/builder/resolve_object.dart
```diff
@@ -39,52 +38,108 @@ TypeResult resolveObject(
...
-    );
+    ]);
     if (isHeader) {
       state.output.add(buildHeaderSerializer(state, identifier, spec, schema));
     }
   }
   return result;
 }
+
+TypeResultObject resolveInterface(
```
provokateurin, line 65: Shouldn't this be in a new file called `resolve_interface.dart`?

packages/dynamite/dynamite/lib/src/helpers/built_value.dart
```diff
@@ -1,65 +1,73 @@
...
-            hiddenConstructor,
-            fromJsonConstructor,
-          ])
-          ..methods.addAll([
-            toJsonMethod,
-            ...methods,
-            buildSerializer(className, isCustom: customSerializer),
-          ]);
+}) {
+  assert(interfaces == null || methods == null, 'Either provide an interface or methods.');
```
provokateurin, line 15:
```suggestion
  assert((interfaces == null) != (methods == null), 'Either provide an interface or methods.');
```

I believe this should be XOR? Otherwise the wording is confusing.

provokateurin (changes requested):
Awesome work :heart:

packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
Leptopoda: > OpenAPI lets you combine and extend model definitions using the allOf keyword. **allOf takes an array of object definitions** that are used for independent validation but together compose a single object.

https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#allof

packages/dynamite/dynamite/lib/src/helpers/built_value.dart
Leptopoda: In theory a normal or should also be possible but I don't see why we should allow this.
I'll change it to an XOR accidentally mess anything up in the future.

packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
provokateurin: Ah ffs I oversaw the `!` and didn't properly read the comment... Sorry for the noise

provokateurin:
Let me do a quick check of the coverage if we need some new coverage ignores.

provokateurin:
Coverage still works as expected.

provokateurin (approved):
Can you open a follow up issue for reducing the number of interfaces to the ones that are actually used?

provokateurin:
:tada:

Issue #306, opened 2023-05-08
neon: dispose state


Leptopoda:
We should propperly dispose states like:

- [x] FocusNode
- [x] TextEditingControllers
- [ ] remove and dispose cached blocs on `removeAccount` and `updateAccount` (AccounsBloc.accounts changes)

provokateurin:
I think there are even more things that are not disposed right now, maybe turn this into a broader issue?

PR #714, opened 2023-09-05
Fix/dispose state

Leptopoda:
closes: #306

### DIFF
--- a/packages/neon/neon/lib/src/blocs/apps.dart
+++ b/packages/neon/neon/lib/src/blocs/apps.dart
@@ -164,9 +164,7 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     unawaited(openNotifications.close());
     unawaited(appVersions.close());
 
-    for (final app in _allAppImplementations) {
-      app.dispose();
-    }
+    super.dispose();
   }
 
   @override
--- a/packages/neon/neon/lib/neon.dart
+++ b/packages/neon/neon/lib/neon.dart
@@ -67,21 +67,31 @@ Future runNeon({
       providers: [
         Provider<GlobalOptions>(
           create: (final _) => globalOptions,
+          dispose: (final _, final globalOptions) => globalOptions.dispose(),
         ),
         Provider<AccountsBloc>(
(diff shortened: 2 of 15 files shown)


packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -164,9 +164,7 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
     unawaited(openNotifications.close());
     unawaited(appVersions.close());
 
-    for (final app in _allAppImplementations) {
-      app.dispose();
-    }
```
provokateurin, line 169: Why is this removed?

provokateurin (approved):
Last night I thought about fixing this today :sweat_smile: I'm surprise that there weren't more places where dispose wasn't called correctly or at all :thinking:

packages/neon/neon/lib/src/blocs/apps.dart
Leptopoda: The `AppsBloc` is not really the owner of the `appImplementations`.
I added a disposal for them at the provider level in neon.dart.

I don't know why we should ever need the `appImplementations` without the `AppsBloc` (or even without the `AccountsBloc`) but it's now possible. Otherwise the `AppsBloc` could dispose our apps while we still need them elsewhere.

Leptopoda:
> I'm surprise that there weren't more places where dispose wasn't called correctly or at all

I thought of
- `TexteditingController`s
- `FocusNode`s
- `ValueListenable`s (=> OptionsCollection)
- `Bloc`s

Our blocs are fairly long lived and we even cache them for multiple accounts that's why the widgets themselves didn't need to be changed that much. 
But I might have missed something :sweat_smile:

Leptopoda:
Aaaah I missed something.

Removing an account doesn't remove and dispose the old blocs. This should probably also be done with `updateAccount` to ensure a fresh initialization of them.
But this will require more testing. @provokateurin  followup PR or should I do it with this one?

provokateurin:
Let's get this merged and then a follow up can fix that part.

PR #715, opened 2023-09-05
Feature/dynamite multiple status codes

provokateurin:
Partially fixes https://github.com/nextcloud/neon/issues/711
Fixes failures in https://github.com/nextcloud/neon/pull/710

I'm not sure if just comparing the responses is the best way, but I wouldn't know any other way that makes sense :thinking:


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -497,111 +511,107 @@ final _response = await ${isRootClient ? 'this' : '_rootClient'}.doRequest(
...
 
-          if (operation.responses != null) {
-            if (operation.responses!.length > 1) {
-              throw Exception('Can not work with multiple status codes right now');
+          for (final responseEntry in responses.entries) {
+            final response = responseEntry.key;
+            final statusCodes = responseEntry.value;
+            code.write(
+              'if (${statusCodes.map((final statusCode) => '_response.statusCode == $statusCode').join(' || ')}) {',
+            );
```
Leptopoda, line 519: Can we use a switch here?
Would make lookup in constant time and I think the code would be cleaner.

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -297,13 +298,26 @@ Iterable<Method> buildTags(
...
+            for (final responseEntry in operation.responses!.entries) {
+              final statusCode = int.parse(responseEntry.key);
+              final response = responseEntry.value;
+
+              responses[response] ??= [];
+              responses[response]!.add(statusCode);
+            }
+
+            // We currently do not support multiple different schemas
+            responses = Map.fromEntries([responses.entries.first]);
```
Leptopoda, line 312: can we print an information that we omit the other ones if we do

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -497,111 +511,107 @@ final _response = await ${isRootClient ? 'this' : '_rootClient'}.doRequest(
...
-                  identifier,
-                  Schema(
-                    properties: response.headers!.map(
-                      (final headerName, final value) => MapEntry(
-                        headerName.toLowerCase(),
-                        value.schema!,
-                      ),
-                    ),
+                  toDartName(
+                    '$operationId-response-${statusCodes.join('-')}-$mimeType',
```
Leptopoda, line 559: Does it even make sense to include the status code in the name or do we need it to avoid collisions?
I think `200201400401404` is quite bad to read.

provokateurin: Need to do that in order to avoid collision. The other option would be using the index of the response, that would at least be more readable.

packages/dynamite/dynamite/lib/src/builder/client.dart
provokateurin: But in the switch case it would just be ` a || b` again, right? We don't have the fall through cases with Dart 3 anymore.
I don't see how that would benefit us (until we support multiple schemas).

Leptopoda: > We don't have the fall through cases with Dart 3 anymore.

:woman_facepalming:  I forgot that.

Leptopoda (approved):

Leptopoda:
Please remove the `fixes` statement in the PR body to avoid auto closing of the issue

provokateurin:
That would only happen if there wasn't another word in front.

PR #721, opened 2023-09-06
fix(neon): remove and dispose dagnling blocs after logout

Leptopoda:
fixes: #306

### DIFF
--- a/packages/neon/neon/lib/src/models/account_cache.dart
+++ b/packages/neon/neon/lib/src/models/account_cache.dart
@@ -1,0 +1,41 @@
+import 'package:neon/models.dart';
+import 'package:neon/src/models/disposable.dart';
+
+/// Cache for [Account] specific [Disposable] objects.
+class AccountCache<T extends Disposable> implements Disposable {
+  AccountCache();
+
+  final Map<String, T> _cache = {};
+
+  /// Disposes all entries and clears the cache.
+  @override
+  void dispose() {
+    _cache
+      ..disposeAll()
+      ..clear();
+  }
+
+  /// Updates the cache against the given [accounts].
+  ///
+  /// Every cache entry with an account no longer in [accounts] is removed and
+  /// disposed. This method is in O(N²).
+  void pruneAgainst(final Iterable<Account> accounts) {
+    _cache.removeWhere((final key, final value) {
+      if (accounts.tryFind(key) == null) {
+        value.dispose();
+        return true;
+      }
+
+      return false;
+    });
+  }
+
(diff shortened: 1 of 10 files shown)


provokateurin:
I think we need something similar for disposing the blocs of an app implementation when the list of available apps is updated and the server no longer has that app. It is very rare, but right now we don't dispose correctly in that case.

Leptopoda:
I also added unit tests for `Disposable` and `AccountCache`

packages/neon/neon/lib/src/models/account_cache.dart
```diff
@@ -0,0 +1,49 @@
...
+        return true;
+      }
+
+      return false;
+    });
+  }
+
+  /// Removes [account] and its associated value, if present, from the cache.
+  ///
+  ///If present the value associated with `account` is disposed.
```
provokateurin, line 35:
```suggestion
  /// If present the value associated with [account] is disposed.
```

packages/neon/neon/lib/src/models/account_cache.dart
```diff
@@ -0,0 +1,49 @@
...
+    final removed = _cache.remove(account?.id);
+    removed?.dispose();
+  }
+
+  /// The value for the given [account], or `null` if [account] is not in the cache.
+  T? operator [](final Account account) => _cache[account.id];
+
+  /// Associates the [account] with the given [value].
+  ///
+  /// If the account was already in the cache, its associated value is changed.
```
provokateurin, line 46:
```suggestion
  /// If the account was already in the cache, it's associated value is changed.
```

packages/neon/neon/test/account_cache_test.dart
```diff
@@ -0,0 +1,74 @@
...
+      expect(cache[account0], isNull);
+
+      cache[account0] = disposable0;
+      expect(cache[account0], disposable0);
+
+      expect(cache[account0] ??= disposable1, disposable0);
+      expect(cache[account1] ??= disposable1, disposable1);
+    });
+
+    test('map functionality', () {
```
provokateurin, line 34:
```suggestion
    test('update', () {
```

provokateurin (approved):

packages/neon/neon/lib/src/models/account_cache.dart
Leptopoda: This is correct.
Its in the form of "sein/seine"
Not it is like "es ist"

provokateurin (approved):

PR #734, opened 2023-09-07
refactor(neon): Refactor icon sizes

provokateurin:
The `kAvatarSize` constant was way overused. Also doing the 2/3 thing lead to odd sizes and it seems like scaling vector graphics with that leads to bad results.

### DIFF
--- a/packages/neon/neon/lib/src/widgets/unified_search_results.dart
+++ b/packages/neon/neon/lib/src/widgets/unified_search_results.dart
@@ -7,13 +7,13 @@ import 'package:neon/src/bloc/result_builder.dart';
 import 'package:neon/src/blocs/accounts.dart';
 import 'package:neon/src/blocs/unified_search.dart';
 import 'package:neon/src/models/account.dart';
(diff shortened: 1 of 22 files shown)


Leptopoda (commented):
I don't really know whether I like the changes.
Looks like we now have 28 and 40 as two sizes.

The IconThemeData defaults to 24 so we should maybe use that instead of 28.
https://api.flutter.dev/flutter/widgets/IconThemeData/size.html

I also think that we should have the 40 in a unified place

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -77,15 +76,15 @@ class NeonUnifiedSearchResults extends StatelessWidget {
               ListTile(
                 leading: const Icon(
                   Icons.close,
-                  size: kAvatarSize,
+                  size: 40,
```
Leptopoda, line 79: And this should not be 40.
I'm for 24 (the default) but 28 is probably also ok.

Leptopoda (approved):

PR #740, opened 2023-09-08
perf: Stop using unnecessary spread operators

provokateurin:
Closes https://github.com/nextcloud/neon/issues/696

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/client.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/client.dart
@@ -169,14 +169,13 @@ Class buildRootClient(
                       ..toSuper = true
                       ..named = true,
                   ),
-                  if (spec.hasAnySecurity) ...[
+                  if (spec.hasAnySecurity)
                     Parameter(
                       (final b) => b
                         ..name = 'authentications'
                         ..toSuper = true
                         ..named = true,
                     ),
-                  ],
                 ]),
             ),
             Constructor(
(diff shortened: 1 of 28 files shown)


Leptopoda (commented):
I think that everywhere where we had a nested `...[]` should use a different approach. I don't like that we now rely on indentation for readability.

Either use an external list and conditionally add to it or use a `BuiltList` like
```dart
BuiltList((final b) {
  for ( element in collection ){
    b.add(element);
  }
})
```

provokateurin:
But we will need spread operators for those again, right?

Leptopoda:
Ah my bad. I meant BuiltList.build():
```dart
    final body = SettingsList(
      categories: BuiltList<SettingsCategory>.build((final b) {
        for (final category in [...appImplementation.options.categories, null]) {
          final options = appImplementation.options.options.where((final option) => option.category == category);
          final title = category != null ? category.name(context) : AppLocalizations.of(context).optionsCategoryOther;

          if (options.isEmpty) {
            continue;
          }
          final categoryTile = SettingsCategory(
            title: Text(title),
            tiles: BuiltList<SettingsTile>.build((final b) {
              for (final option in options) {
                if (option is ToggleOption) {
                  b.add(
                    CheckBoxSettingsTile(
                      option: option,
                    ),
                  );
                } else if (option is SelectOption) {
                  b.add(
                    DropdownButtonSettingsTile(
                      option: option,
                    ),
                  );
                }
              }
            }).toList(),
          );

          b.add(categoryTile);
        }
      }).toList(),
    );
```

this also allows us to do the cache values like the `appImplementation.options.options.where` wich is also not that cheap.

provokateurin:
Oh that looks neat!

provokateurin:
But how do we handle conditionals? Just put the if inside a BuiltList as well?

Leptopoda:
can you show me an example?

I only want the BuiltList approach where we have nested if/for.

provokateurin:
I feel like using external lists will be easier to comprehend for new people, so I would prefer that approach.

Doing this refactor will be much more work than I anticipated so I will close this PR and maybe comeback later or someone else can do it as well.

PR #742, opened 2023-09-08
Refactor/types

Leptopoda:


provokateurin:
Is there a point in using `Future<void>` over `Future`? I personally don't like it because it adds no useful information.

packages/neon/neon_files/lib/widgets/file_preview.dart
```diff
@@ -72,19 +70,17 @@ class FilePreview extends StatelessWidget {
...
   factory FilePreviewImage({
-    required final Account account,
     required final FileDetails file,
     required final Size size,
   }) {
     final width = size.width.toInt();
     final height = size.height.toInt();
     final path = file.path.join('/');
-    final cacheKey = '${account.id}-preview-$path-$width-$height';
+    final cacheKey = 'preview-$path-$width-$height';
```
provokateurin, line 81: This needs to stay as another account might have the same path, width and height but the image is actually different. Not critical since we don't leak any info as we are already on the client side, but still a problem.

Leptopoda:
`Future` will be interpreted as `Future<dynamic>` and this shows to callers that they don't need to expect a result.

```dart
void main(List<String> args) async{
  final val1 = await test();
  print(val1); // Error: This expression has type 'void' and can't be used.
  
  final val2 = await test2();
  print(val2);
}


Future<void> test()async{
  //
}

Future test2()async{
  //
}
```

provokateurin (changes requested):

packages/neon/neon_files/lib/widgets/file_preview.dart
Leptopoda: NeonApiImage does that for us :)

provokateurin (changes requested):
Please enable `strict-raw-types` in the analysis_options.yaml to enforce `Future<void>`

packages/neon/neon_files/lib/widgets/file_preview.dart
provokateurin: Ah I overlooked that part.

provokateurin (approved):
There is also `strict-inference` which we could enable as well, but I'm not sure if it is that helpful for us.

Leptopoda:
doesn't look lie it changes anything rn.
I'd be curious if this could catch some of the type errors we had with the `RequestManager._wrap`

PR #748, opened 2023-09-08
refactor(dynamite): Use built_value for OpenAPI specification models

provokateurin:
Closes https://github.com/nextcloud/neon/issues/719


packages/dynamite/dynamite/lib/src/models/discriminator.dart
```diff
@@ -1,30 +1,16 @@
...
 
-  @override
-  bool operator ==(final Object other) =>
-      other is Discriminator &&
-      propertyName == other.propertyName &&
-      const MapEquality().equals(mapping, other.mapping);
-
-  @override
-  int get hashCode => propertyName.hashCode + const MapEquality().hash(mapping);
+  Map<String, String>? get mapping;
```
Leptopoda, line 15: I think to gain real collection equality we will need to use built_collection (`BuiltMap`, `BuiltList`, `BuiltSet`) throughout the models.

Also guarantees immutability to we don't mutate our input in dynamite  by accident.

packages/dynamite/dynamite/lib/src/models/serializers.dart
```diff
@@ -0,0 +1,44 @@
...
+  RequestBody,
+  Response,
+  Schema,
+  SecurityScheme,
+  Server,
+  ServerVariable,
+  Tag,
+])
+final Serializers serializers = _$serializers;
+final Serializers jsonSerializers = (serializers.toBuilder()..addPlugin(StandardJsonPlugin())).build();
```
Leptopoda, line 44: I think we can inline this.
I don't see a reason where we ever need the built_value serialization

packages/dynamite/dynamite/lib/src/models/discriminator.dart
```diff
@@ -1,30 +1,16 @@
...
-  const Discriminator({
-    required this.propertyName,
-    this.mapping,
-  });
+abstract class Discriminator implements Built<Discriminator, DiscriminatorBuilder> {
+  factory Discriminator([final void Function(DiscriminatorBuilder) updates]) = _$Discriminator;
 
-  factory Discriminator.fromJson(final Map<String, dynamic> json) => _$DiscriminatorFromJson(json);
-  Map<String, dynamic> toJson() => _$DiscriminatorToJson(this);
+  Discriminator._();
```
Leptopoda, line 9: I usually like making them const :woman_shrugging: 

This is nothing functional and rather a preference of code style

packages/dynamite/dynamite/lib/src/models/serializers.dart
provokateurin: You mean just move it to openapi_builder.dart?

Leptopoda: I meant dropping access to the serializer without the StandardJsonPlugin (Sorry, really bad wording from me):

```suggestion
final Serializers serializers = (_$serializers.toBuilder()..addPlugin(StandardJsonPlugin())).build();
```

packages/dynamite/dynamite/lib/src/models/path_item.dart
```diff
@@ -1,97 +1,61 @@
...
+  put('put'),
+  post('post'),
+  delete('delete'),
+  options('options'),
+  head('head'),
+  patch('patch'),
+  trace('trace');
+
+  const PathItemOperation(this.value);
+  final String value;
```
Leptopoda, line 60: as the enum names are the same as their value we could just drop the extra value and call `PathItemOperation.get.name`.

If you like the current approach for saftey or codestyle reasons I'm fine with it.

Leptopoda (approved):
I'm fine with the current state but if you agree with me on the remaining points please change them.

Thank you for tackling this :)

Leptopoda:
I just noticed that this is going to make it nearly impossible (or at least very hard) for us to support yaml in the future as https://pub.dev/packages/checked_yaml only supports json_serializable.

I don't think this is a problem (as json is more common and both are easily convertible), I just wanted to mention this.

packages/dynamite/dynamite/lib/src/models/path_item.dart
provokateurin: Ah I forgot that this is possible

PR #749, opened 2023-09-09
WIP: support date and date-time formats

Leptopoda:
This is going to be needed for cookbook support.
I had this on a branch from march and thought it's time :)

I also plan to implement the `base64` `byte` format and an option for custom formats.
I plan to add a config option to dynamites build-yaml config (or a separate file) that can specify:
- string with the import directive
- name of the custom serializer
- serialization method
- deserialization method

This would enable us to have custom formats like `unix-time` or iso 8601 duration support (the later also being needed by cookbook).

I do not plan to implement it without tests though as I can see those features not being used all to frequesntly

PR #759, opened 2023-09-11
feat(dynamite): Support decoding booleans

provokateurin:


Leptopoda:
Is this something we need immediately or just a nice to have?

As this isn't really openapi conformant I'd like to have this behind an option or specify custom types (see #749). In that case let's talk about it next week.

provokateurin:
I had this problem with the generated Talk spec, I don't remember exactly in what context this occurred 😅

provokateurin:
I remember what this was used for: Booleans in headers
We now have the problem that this is also present in the core of Nextcloud server, so we need to support this in some way.
As far as I can tell this has to be fully supported since headers are just a normal OpenAPI schema.
I'm not sure how to proceed with this :thinking:

provokateurin:
As far as I can tell this is expected behavior. https://swagger.io/docs/specification/describing-parameters has an example which encodes a boolean as literal `true` and https://github.com/stoplightio/elements (which Nextcloud uses to render the OpenAPI documentation) also does this when sending requests or generating code snippets.
The encoding part doesn't need to be updated because it already works ootb when calling `toString()` on a Dart boolean.

Leptopoda (approved):
LGTM.
It'll probably be changed a bit with the DynamiteRawResponse

Issue #761, opened 2023-09-11
Test against petstore examples


provokateurin:
https://github.com/OAI/OpenAPI-Specification/tree/main/examples

Leptopoda:
I think we should already generate for the petstore.
This should help us further test dynamite.

Resolving the authentication issues and adding proper tests can still be added later on.
What about opening a PR with your current changes? @provokateurin

provokateurin:
I can do that, shall we just patch the petstore spec to remove the broken authentication or how do you want to handle it?

Leptopoda:
IIrc we weren't failing to build it.
I'd just keep it vanilla.

provokateurin:
Building works, but the code won't pass the analyzer because of missing classes.

Leptopoda:
Yeah in that case patching might be the only possible thing.

provokateurin:
Let's just continue here: We have to write unit tests for the example and run them against https://github.com/OpenAPITools/openapi-petstore.

PR #764, opened 2023-09-13
Test/dynamite/petstore

Leptopoda:
resolves: #761 
I've tried setting up the petstore docker container form the petstore repo (https://github.com/swagger-api/swagger-petstore) but I run into java killing itself with:
```
library initialization failed - unable to allocate file descriptor table - out of memory%                                                                                                          
```

According to this [stackoverflow](https://stackoverflow.com/questions/49435109/error-upon-jar-execution-unable-to-allocate-file-descriptor-table#56895801) thread this is because of the number of open files and it should be fixed with java 9.
I don't have time to further investigate this (build the docker container with a newer java version) and would rather keep my high open file count (as I regularly need this). If anyone wants to pick this up feel free :)


packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -0,0 +1,148 @@
...
+                ..name = parameterName
+                ..type = refer(result.nullableName)
+                ..named = true
+                ..required = dartParameterRequired,
+            ),
+          );
+
+          if (dartParameterNullable) {
+            code.write('if ($parameterName != null) {');
+          }
```
provokateurin, line 135: Can we unify this with the code above?

packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -0,0 +1,148 @@
...
+        state,
+        'identifier-$mimeType',
+        mediaType.schema!,
+      );
+
+      if (mimeType == '*/*' || mimeType == 'application/octet-stream' || mimeType.startsWith('image/')) {
+        return ('Uint8List', '_response.bodyBytes', true);
+      } else if (mimeType.startsWith('text/') || mimeType == 'application/javascript') {
+        return ('String', '_response.body', true);
+      } else if (mimeType == 'application/json') {
```
provokateurin, line 37: This could be turned into a switch statement

Leptopoda: It has `mimeType.startsWith()` and I don't think that we could put it in a switch

provokateurin:
I think we should generate the e2e test as the example inside the dynamite package.

Leptopoda:
Closing in favor of a new version @provokateurin  is working on

PR #776, opened 2023-09-16
style(neon): make RelativeTime stateful

Leptopoda:
As discussed this is more performant.

We can try to reproduce this tomorrow but I think it was most aparent with the listview PR

### DIFF
--- a/packages/neon/neon/lib/src/widgets/relative_time.dart
+++ b/packages/neon/neon/lib/src/widgets/relative_time.dart
@@ -1,49 +1,52 @@
-import 'package:flutter/widgets.dart';
+import 'dart:async';
 
-class RelativeTime extends StatelessWidget {
+import 'package:flutter/widgets.dart';
+import 'package:neon/src/utils/relative_time.dart';
+
+/// Shows the time elapsed since a [DateTime] and periodically updates itself.
+class RelativeTime extends StatefulWidget {
+  /// Creates a new relative DateTime widget.
   const RelativeTime({
     required this.date,
     this.style,
     super.key,
   });
 
+  /// The timestamp to be displayed.
   final DateTime date;
+
+  /// The text style of the calculated time.
+  ///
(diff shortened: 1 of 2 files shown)


packages/neon/neon/lib/src/widgets/relative_time.dart
```diff
@@ -1,49 +1,52 @@
+import 'dart:async';
+
 import 'package:flutter/widgets.dart';
+import 'package:neon/src/utils/relative_time.dart';
 
-class RelativeTime extends StatelessWidget {
+/// Shows the time elapsed since a [DateTime] and periodically updates itself.
+class RelativeTime extends StatefulWidget {
+  /// Creates a new relative DateTime widget.
```
provokateurin, line 8: I find these constructor docs useless if they only say that they create a new object of the class. Better to ignore those as they are not helpful imo

packages/neon/neon/lib/src/widgets/relative_time.dart
```diff
@@ -1,49 +1,52 @@
...
   const RelativeTime({
     required this.date,
     this.style,
     super.key,
   });
 
+  /// The timestamp of the UI element.
```
provokateurin, line 15:
```suggestion
  /// The timestamp to be displayed.
```

packages/neon/neon/lib/src/widgets/relative_time.dart
```diff
@@ -1,49 +1,52 @@
...
   final DateTime date;
+
+  /// The style of the calculated time.
```
provokateurin, line 18:
```suggestion
  /// The text style of the calculated time.
```

provokateurin (changes requested):

provokateurin:
Should have been perf or refactor :upside_down_face:

provokateurin (approved):

PR #779, opened 2023-09-17
feat(neon_files): Implement sharing file natively

provokateurin:
Closes https://github.com/nextcloud/neon/issues/42


Leptopoda:
I'd rather implement this after #803 in a cleaner way without the cache potentially growing forever

provokateurin:
@Leptopoda can we please just merge this? The current code that "syncs" a file already uses this logic, so adding this small feature which reuses most of that logic is not too bad imo. I understand that we should do this in a better way, but it's also not great to keep this PR around forever until we have the cleaner way.

Leptopoda (approved):
If you really want to ....

But please open an issue if we don't already have one

PR #832, opened 2023-09-22
feat(tool): Configure renovate groups

provokateurin:
Closes https://github.com/nextcloud/neon/issues/828

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -6,5 +6,37 @@
   "labels": [
     "dependencies"
   ],
-  "rangeStrategy": "bump"
+  "rangeStrategy": "bump",
+  "packageRules": [
+    {
+      "groupName": "built_value",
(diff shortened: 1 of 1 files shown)


Leptopoda (commented):
👍 
Once we publish our packages dynamite => dynamite_runtime would also be a good candidate.

Leptopoda (approved):

provokateurin:
Technically yes, but since we use them via local path they never get any updates

Leptopoda:
But the published nextcloud package will need to depend on the published dynamite versions.
Technically dynamite itself doesn't matter as it's a dev dependency but we should make sure they're both in sync.

Maybe melos also does that for us?

provokateurin:
Ah that part you mean. I guess you are right about configuring a renovate group for them

PR #837, opened 2023-09-23
Fix/dashboard api

provokateurin:
While writing dashboard tests I found that the specification was broken: https://github.com/nextcloud/server/pull/40511 After updating that I wanted to regenerate everything, but then I hit https://github.com/nextcloud/server/pull/40575 which I fixed as well, but then I also hit https://github.com/nextcloud/neon/pull/759 again.


Leptopoda (approved):

provokateurin:
I will wait for https://github.com/nextcloud/server/pull/40575 to be merged, otherwise it gets messy with depending on a commit that is not in the master branch.

PR #843, opened 2023-09-24
fix(nextcloud): Export all serializers in the merged serializers

provokateurin:
Forgot to export all of the serializers. This part should also be generated automatically once we automate the client generation even further.


Leptopoda:
It will be irrelevant with the DynamiteRawResponse as we will no longer need the `serializeWith` helper methods.

Leptopoda (approved):

PR #857, opened 2023-09-27
fix(nextcloud): Fix OCS-APIRequest header

provokateurin:
Some nice cleanups.
All auto-generated as always, so nothing to really review here.


Leptopoda (approved):
indeed some nice cleanups

PR #859, opened 2023-09-27
fix(nextcloud): Fix autocomplete

provokateurin:
Apparently I misunderstood the server code and there is no bug. Just an overcomplicated API that has weird quirks for historical reasons. Depends on https://github.com/nextcloud/server/pull/40660


Leptopoda (approved):
Meh :(
But nothing we can do about it.

PR #868, opened 2023-09-28
Prevent untested Neon APIs

provokateurin:
Closes https://github.com/nextcloud/neon/issues/867


Leptopoda (approved):
I don't fully understand the regex magic but it looks quite nice

PR #869, opened 2023-09-29
Refactor nextcloud package exports

provokateurin:
Splits all the exports into separate files.

Moves the supported check to the app implementations to prevent importing the respective APIs in the core.

Also generates the app IDs. They are still not implemented as an enum since enum can't be used in consts yet (although there has been a fix in the Dart SDK already).

Sadly this doesn't reduce the app size at all (which was my initial goal), but I still think that it improves the structure quite a lot (also because we now won't forget to export something).

Closes https://github.com/nextcloud/neon/issues/796
Partially addresses https://github.com/nextcloud/neon/issues/794


packages/nextcloud/lib/nextcloud.dart
Leptopoda, line 1: Why not having them all in one file again?

packages/nextcloud/bin/generate_exports.dart
```diff
@@ -0,0 +1,37 @@
...
+
+    File('lib/$id.dart').writeAsStringSync('''
+import 'package:nextcloud/src/api/$id.openapi.dart';
+import 'package:nextcloud/src/client.dart';
+
+${exports.join('\n')}
+
+// ignore: public_member_api_docs
+extension ${classPrefix}Extension on NextcloudClient {
+  static final _$variablePrefix = Expando<${classPrefix}Client>();
```
Leptopoda, line 35: why exactly an expando?

packages/neon/neon/lib/src/blocs/apps.dart
```diff
@@ -119,24 +117,26 @@ class AppsBloc extends InteractiveBloc implements AppsBlocEvents, AppsBlocStates
...
-          _ => (true, null),
-        };
-
-        if (!supported) {
-          notSupported.add((id, minVersion));
+        final support = await app.isSupported(_account, capabilities.requireData);
+        if (support == null) {
+          continue;
+        }
+        if (!support.$1) {
```
Leptopoda, line 138: I'm not a fan of the `.$n` syntax for records.

I'd rather have variable names that I can understand while reading the code

packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -33,6 +36,18 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
...
   R get options;
 
+  /// Checks if the app is supported on the server of the [account].
+  ///
+  /// A return value of `null` means that it can not be known if the app is supported.
+  /// This is the case for apps that depend on the server version like files.
+  /// The server support is handled differently.
+  ///
+  /// The first value of the record is the supported status and the second value is the supported minimum version.
+  FutureOr<(bool, String?)?> isSupported(
```
Leptopoda, line 46:
```suggestion
  FutureOr<(bool isSupported, String? minimumVersion)?> isSupported(
```

packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -33,6 +36,18 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
   @mustBeOverridden
   R get options;
 
+  /// Checks if the app is supported on the server of the [account].
+  ///
+  /// A return value of `null` means that it can not be known if the app is supported.
+  /// This is the case for apps that depend on the server version like files.
```
Leptopoda, line 42: Please add information on what we do in this case (treating it as supported).

packages/nextcloud/lib/ids.dart
```diff
@@ -0,0 +1,25 @@
+// ignore_for_file: public_member_api_docs
+
+sealed class AppIDs {
```
Leptopoda, line 3: An enhanced enum is better.

Leptopoda (changes requested):
I love these changes :heart_eyes:

packages/nextcloud/lib/nextcloud.dart
provokateurin: True, I stopped doing this because I though I could reduce the app size this way, but since it doesn't help I can add it back.

packages/nextcloud/bin/generate_exports.dart
provokateurin: Extensions can't introduce new variables. Using a simple map would keep the reference to the object until the whole program exits, so we wouldn't get rid of unreachable objects. Expando does exactly that for us: https://api.dart.dev/stable/3.1.3/dart-core/Expando-class.html

packages/neon/neon/lib/src/blocs/apps.dart
provokateurin: Agreed, but then we would have to introduce a class for this (I would be fine with that).

packages/neon/neon/lib/src/models/app_implementation.dart
provokateurin: Ahhh, stupid me forgot that you can name them :)

packages/nextcloud/bin/generate_exports.dart
Leptopoda: > Using a simple map would keep the reference to the object until the whole program exits, so we wouldn't get rid of unreachable objects.

Can you further explain this/ point me to some documentation

provokateurin: So previously I had something like
```dart
static final _core = <NextcloudClient, CoreClient>{};
```
which worked just fine.
Now this static constant stays alive during the whole program lifecycle.
This means the values inside the map are never cleaned up by GC when the respective NextcloudClient becomes unreachable.
To correctly free the object you can use Expando since it will remove the value once it becomes unreachable.

Leptopoda: Ah now I get it.

Thanks.

provokateurin: I also didn't know about Expando, but it's a pretty cool feature :)

packages/neon/neon/lib/src/blocs/apps.dart
provokateurin: I removed this from the PR and will open a second one since the changes are not directly related to this PR anymore.

packages/nextcloud/lib/ids.dart
provokateurin: As I said, this is not possible due to the need have the IDs as const.

Leptopoda: Makes sense.

Then please make the class final. There is no reason to have it sealed here.

Leptopoda: If it complains about it being a utility class just add a hidden constructor.

provokateurin:
@Leptopoda please review again :)

Leptopoda (approved):

PR #870, opened 2023-09-29
gracefully handle multiple mimeTypes and default responses

Leptopoda:
Needed for #764

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
@@ -15,7 +15,7 @@ TypeResult? resolveMimeTypeDecode(
 ) {
   if (response.content != null) {
     if (response.content!.length > 1) {
-      throw Exception('Can not work with multiple mime types right now');
+      print('Can not work with multiple mime types right now. Using the first supported.');
     }
 
     for (final content in response.content!.entries) {
@@ -35,10 +35,9 @@ TypeResult? resolveMimeTypeDecode(
         return TypeResultBase('String');
       } else if (mimeType == 'application/json') {
         return result;
-      } else {
-        throw Exception('Can not parse mime type "$mimeType"');
       }
     }
+    throw Exception('Can not parse any mime type of Operation: "$identifier"');
   }
   return null;
 }
@@ -52,7 +51,7 @@ Iterable<String> resolveMimeTypeEncode(
 ) sync* {
   if (operation.requestBody != null) {
     if (operation.requestBody!.content!.length > 1) {
-      throw Exception('Can not work with multiple mime types right now');
+      print('Can not work with multiple mime types right now. Using the first supported.');
     }
     for (final content in operation.requestBody!.content!.entries) {
       final mimeType = content.key;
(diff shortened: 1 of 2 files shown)


packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -15,7 +15,7 @@ TypeResult? resolveMimeTypeDecode(
 ) {
   if (response.content != null) {
     if (response.content!.length > 1) {
-      throw Exception('Can not work with multiple mime types right now');
+      print('Can not work with multiple mime types right now. Using the first supported.');
```
provokateurin, line 18: This will break since afterwards it just iterates over all entries. You need to change the entries to be just the first one.

packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -52,7 +51,7 @@ Iterable<String> resolveMimeTypeEncode(
 ) sync* {
   if (operation.requestBody != null) {
     if (operation.requestBody!.content!.length > 1) {
-      throw Exception('Can not work with multiple mime types right now');
+      print('Can not work with multiple mime types right now. Using the first supported.');
```
provokateurin, line 54: Same as above

packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -35,10 +35,9 @@ TypeResult? resolveMimeTypeDecode(
         return TypeResultBase('String');
       } else if (mimeType == 'application/json') {
         return result;
-      } else {
-        throw Exception('Can not parse mime type "$mimeType"');
       }
     }
+    throw Exception('Can not parse any mime type of Operation:"$identifier"');
```
provokateurin, line 40:
```suggestion
    throw Exception('Can not parse any mime type of Operation: "$identifier"');
```

packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
```diff
@@ -121,9 +120,9 @@ Iterable<String> resolveMimeTypeEncode(
             yield '}';
           }
           return;
-        default:
-          throw Exception('Can not parse mime type "$mimeType"');
       }
     }
+
+    throw Exception('Can not parse any mime type of Operation:"$identifier"');
```
provokateurin, line 126:
```suggestion
    throw Exception('Can not parse any mime type of Operation: "$identifier"');
```

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/builder/resolve_mime_type.dart
Leptopoda: Why?

I'll return the first supported entry. If the first one is unsupported (like xml) the for loop will just continue to the next entry.
If no entry is returned from within the loop the loop will finish and the exception below will be thrown,

provokateurin: Ahh I see, I just wasn't aware of the whole control flow. I forgot this now uses early returns and is no longer part of the monster functions :sweat_smile:. So both are actually handled correctly.

provokateurin (approved):

PR #877, opened 2023-09-29
Move app support check to app implementations

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/869 with the review feedback included.
Needed for https://github.com/nextcloud/neon/issues/808


packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -33,6 +36,18 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
...
   R get options;
 
+  /// Checks if the app is supported on the server of the [account].
+  ///
+  /// A `supported` value of `null` means that it can not be known if the app is supported.
+  /// This is the case for apps that depend on the server version like files and we assume that the app is supported.
+  /// The server support is handled differently.
+  ///
+  /// The first value of the record is the supported status and the second value is the supported minimum version.
+  FutureOr<(bool? supported, String? minimumVersion)> isSupported(
```
Leptopoda, line 46: In #869 you had this nullable wich i think was cleaner.
Especially as many plugins won't need this check.

```suggestion
  FutureOr<(bool? supported, String? minimumVersion)>? isSupported(
```

provokateurin: It's more annoying to destructure though, that's why I went with non-nullable.
The problem is you first need to check if the result is null and then convert the record into two variables.

provokateurin: And returning `(null, null)` instead of just `null` isn't much more effort imo.

Leptopoda: Ah I remember. That's just badly made from dart.

In that case please make the records constant `const (null, null)`.

PR #745, opened 2023-09-08
docs(neon): document public api

Leptopoda:
tackles the code side of #692

This is just some initial work I did. I mainly wanted to document the public facing api and whatever we find odd should be opened in an issue. Ill work on this every now and then in the next weeks so we can have a decent documentation at launch.

@provokateurin  If you have the time and want help feel free to add commits to the branch (and fix my typos. I think reviews are just an overhead here). We can later squash everything together into a nice big commit :)

PR #880, opened 2023-09-29
refactor image handling in neon

Leptopoda:
Again pulled out from #745


packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -192,42 +236,128 @@ class _NeonCachedImageState extends State<NeonCachedImage> {
...
+  /// {@macro NeonCachedImage.iconColor}
   final Color? iconColor;
+
+  /// {@macro NeonCachedImage.errorBuilder}
   final ErrorWidgetBuilder? errorBuilder;
 
   @override
   Widget build(final BuildContext context) {
-    final account = NeonProvider.of<AccountsBloc>(context).activeAccount.value!;
+    final account = this.account ?? NeonProvider.of<AccountsBloc>(context).activeAccount.value!;
```
provokateurin, line 297: This feels quite dangerous because we should always load the images with the correct account. Can you mark the account field as non-nullable and required and then explicitly specify it this way everywhere where it is used?

packages/neon/neon/lib/src/widgets/unified_search_results.dart
```diff
@@ -129,18 +127,16 @@ class NeonUnifiedSearchResults extends StatelessWidget {
...
-        size: size,
+        size: Size.square(IconTheme.of(context).size!),
         url: entry.icon,
         account: account,
       );
     }
 
     if (entry.icon.startsWith('icon-')) {
       return NeonServerIcon(
-        size: size,
```
provokateurin, line 143: This should still be passed since we don't want the icon to just fill all the available space

provokateurin (changes requested):

packages/neon/neon/lib/src/widgets/unified_search_results.dart
Leptopoda: Bruh I made it default to the current icon size and forgot to also pull those changes in.

Should I pull them in or leave them out for a separate PR?

provokateurin: I think putting it in here makes sense, so yes

packages/neon/neon/lib/src/widgets/image.dart
Leptopoda: Not really

The old way did use the active one by default and this is needed as we (eventually) don't want to export the `AccountsBloc`.
So how would other clients then be able to use the currently active one.

provokateurin: Either you always use the active account and note that down as the expected behavior or you need to always specify it. This way of having both options is dangerous in my opinion. Better to just hide the account field then

Leptopoda: I totally agree.
Do you think a separate constructor would help? Like `NeonApiImage.custom` that would require the account.

`NeonUserAvatar` already uses this option and while it has access to the `AccountsBloc` other ones like the potential `NeonTalk` client wouldn't

provokateurin: Yes that seems to be the best option :thinking:

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -201,33 +256,127 @@ class NeonApiImage extends StatelessWidget {
     this.isSvgHint = false,
     this.size,
     this.fit,
-    this.svgColor,
+    this.iconColor,
+    this.errorBuilder,
+    super.key,
+  }) : account = null;
+
+  /// Creates a new Neon api image with the active account.
```
provokateurin, line 264: Not the active account

packages/neon/neon/lib/src/widgets/server_icon.dart
```diff
@@ -1,26 +1,68 @@
 import 'package:flutter/material.dart';
 import 'package:vector_graphics/vector_graphics.dart';
 
+/// Nextcloud server icon.
+///
+/// Draws an icon from the nextcloud server icon set.
```
provokateurin, line 6:
```suggestion
/// Draws an icon from the Nextcloud server icon set.
```

packages/neon/neon/lib/src/widgets/server_icon.dart
```diff
@@ -1,26 +1,68 @@
...
+  /// The color (whether specified explicitly here or obtained from the
+  /// [IconTheme]) will be further adjusted by the nearest [IconTheme]'s
+  /// [IconThemeData.opacity].
+  ///
+  /// {@tool snippet}
+  /// Typically, a Material Design color will be used, as follows:
+  ///
+  /// ```dart
+  /// NeonServerIcon(
+  ///   Icons.widgets,
```
provokateurin, line 32:
```suggestion
  ///   'icon-add',
```

provokateurin (changes requested):

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -12,38 +12,62 @@ import 'package:neon/src/widgets/error.dart';
...
+/// The image is cached in the [DefaultCacheManager] to avoid expensive
+/// fetches.
+///
+/// See:
+///  * [NeonApiImage] for an image from the [NextcloudClient]
+///  * [NeonImageWrapper] for a wrapping widget for images
 class NeonCachedImage extends StatefulWidget {
-  const NeonCachedImage({
-    required this.image,
+  /// Pints the data from [image] to the canvas.
```
provokateurin, line 39:
```suggestion
  /// Prints the data from [image] to the canvas.
```

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -63,15 +86,36 @@ class NeonCachedImage extends StatefulWidget {
         ),
         super(key: Key(cacheKey));
 
+  /// The image content.
   final Future<Uint8List> image;
+
+  /// {@template NeonCachedImage.svgHint}
+  /// Hint whether the image is an svg.
```
provokateurin, line 93:
```suggestion
  /// Hint whether the image is an SVG.
```

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -63,15 +86,36 @@ class NeonCachedImage extends StatefulWidget {
...
+  /// {@template NeonCachedImage.fit}
+  /// How to inscribe the image into the space allocated during layout.
+  /// {@endtemplate}
   final BoxFit? fit;
 
-  final Color? svgColor;
+  /// {@template NeonCachedImage.iconColor}
+  /// The color to use when drawing the image.
+  ///
+  /// Applies to svg images only and the loading animation.
```
provokateurin, line 110:
```suggestion
  /// Applies to SVG images only and the loading animation.
```

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -192,7 +236,18 @@ class _NeonCachedImageState extends State<NeonCachedImage> {
       );
 }
 
+/// Nextcloud api image.
```
provokateurin, line 239:
```suggestion
/// Nextcloud API image.
```

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -192,7 +236,18 @@ class _NeonCachedImageState extends State<NeonCachedImage> {
...
+///
+/// Extension for [NeonCachedImage] providing a [NextcloudClient] to the caller
+/// to retrieve the image.
+///
+/// See:
+///  * [NeonCachedImage] for a customized image
+///  * [NeonImageWrapper] for a wrapping widget for images
 class NeonApiImage extends StatelessWidget {
+  /// Creates a new Neon api image with the active account.
```
provokateurin, line 248:
```suggestion
  /// Creates a new Neon API image with the active account.
```

packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -201,33 +256,127 @@ class NeonApiImage extends StatelessWidget {
     this.isSvgHint = false,
     this.size,
     this.fit,
-    this.svgColor,
+    this.iconColor,
+    this.errorBuilder,
+    super.key,
+  }) : account = null;
+
+  /// Creates a new Neon api image for a given account.
```
provokateurin, line 264:
```suggestion
  /// Creates a new Neon API image for a given account.
```

packages/neon/neon/lib/src/widgets/server_icon.dart
```diff
@@ -1,26 +1,68 @@
...
+  /// The color (whether specified explicitly here or obtained from the
+  /// [IconTheme]) will be further adjusted by the nearest [IconTheme]'s
+  /// [IconThemeData.opacity].
+  ///
+  /// {@tool snippet}
+  /// Typically, a Material Design color will be used, as follows:
+  ///
+  /// ```dart
+  /// NeonServerIcon(
+  ///   'icon-add',
```
provokateurin, line 32:
```suggestion
  ///   icon: 'icon-add',
```

provokateurin (changes requested):

provokateurin (approved):

PR #883, opened 2023-09-30
fix(neon_files): Fix files navigator

provokateurin:


Leptopoda (approved):
Code looks good.

Why exactly is the commit marked as a `fix`?

provokateurin:
The button with the directory name was cut off due to the height constraint

PR #899, opened 2023-10-01
Fix request manager

provokateurin:


Leptopoda (commented):
1. why check the runtime type and not `T` directly like `T iIs Never`?
2. why not add this assertion to the constructors so we can also get compile time warnings for the `const Result.loading()` usecase.

provokateurin:
With the assertion in the constructor the constructors can't be const anymore because `Never` is no const, but I think that trade-off is probably worth it. What do you think?

provokateurin:
I would like to leave the assertion in the copyWith method. It is only relevant for that piece of code and we don't get any compile time checks by moving the assertion into the constructor since the constructor then can't be const anymore (see previous message). This will now lead to random failures because we might not instantly run into all the assertions and fix them.

Leptopoda:
I don't know what you are running into but 
```dart
  const Result.error(this.error)
      : data = null,
        isLoading = false,
        isCached = false,
        assert(T is! Never, '');
```

Is doing the trick on my side.

provokateurin:
You can not use `T is! Never` since you want to compare types and not if the type of `T` (which is `Type`) is not from type `Never`. Therefore you need to use `T != Never` and in this case it complains that `Never` is not constant.

provokateurin:
But I just got an idea how to make it work. Let me try it before we continue to discuss this and potentially waste time :)

Leptopoda:
Yeah I also noticed that.
I'm currently investigating the issue further as `const Result.loading()` in the unit tests is correctly inferred to `Result<dynamic>.loading()`.
This means the unit test is passing when removing the assertion.

Another thing is that:
```dart
    const  loading =  Result.loading();
    final value = subject.valueOrNull?.asLoading() ?? loading;
    subject.add(value);
```

both produces the lint `inference_failure_on_instance_creation` for line (1)
and infers the type of value to `Result<dynamic>` producing an error in line (3)

Leptopoda:
> But I just got an idea how to make it work. Let me try it before we continue to discuss this and potentially waste time :)

Too late for me :P

provokateurin:
I think T is dynamic when it's not a const and Never when it's a const. A bit annoying since it makes testing harder

Leptopoda:
See my comment above where the type is inferred to `dynamic` with a const invocation and making `T` exdend either  `Object` or `Object?` does also not work at all.

I feel like the type interference is wrongly inferring `Never` instead of `dynamic`

Leptopoda:
I'm currently trying to fully understand what Never is used for. Maybe you find [the documentation](https://dart.dev/null-safety/understanding-null-safety#never-for-unreachable-code)helpful.

provokateurin:
My idea doesn't work :woman_shrugging:

Leptopoda:
The following correctly infers the type to `A<dynamic>`

```dart
void main(List<String> args) {
  final b = B<A>();
  
  final val = b.valueOrNull?.copyWith() ?? const A();

}


class A<T> {
  const A();
  
  A<T> copyWith() => A();
  
}


class B<T>{
 
  T? valueOrNull;
}
```

provokateurin:
Ah I see, so Dart can infer the type correctly. That also works when calling `BehaviorSubject.add`

provokateurin:

Leptopoda (commented):
I think we should now drop the assertion.

If needed only a `T != Never` should be added to the constructors

provokateurin:
I think `T == Never` can only happen if any of the the constructors would be const.
The assertion that `T != dynamic` should stay because it is still possible to trigger it if you either ignore the analysis warnings or explicitly call `Result<dynamic>` (see the tests).

Leptopoda:
But `copyWith` is correctly working with `Result<dynamic>` so why block it?

Just remove the immutable annotation and it should be fine.

Leptopoda (approved):

Leptopoda:
I`ve opened an issue on the dart GH.
Let's see what will come out of this: https://github.com/dart-lang/sdk/issues/53668

PR #918, opened 2023-10-03
feat(tool,ci): Only run tests of changed packages and their dependents

provokateurin:
Closes https://github.com/nextcloud/neon/issues/494

The script is a bit more complicated because melos does not care about the order of arguments.
First we need to find the changed packages and the dependents.
Then all tests of those packages need to run.
Putting both conditions into a single command does not work.

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -20,7 +20,7 @@ ide:
 
 scripts:
   format: dart format --fix --line-length 120 .
-  format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
+  format:check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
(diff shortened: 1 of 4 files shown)


melos.yaml
```diff
@@ -20,8 +20,11 @@ ide:
 
 scripts:
   format: dart format --fix --line-length 120 .
-  format-check: dart format --output=none --set-exit-if-changed --line-length 120 .
+  format:check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
   test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
+  test:changed: |
+    packages_glob="{$(echo $(melos list --parsable --diff=origin/main --include-dependents | xargs -r -n1 basename) | sed "s/ /,/g")}"
+    melos exec --scope="$packages_glob" --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
```
Leptopoda, line 28: Does it make sense to move this into it's own shell script with proper variables and methods?

It's kinda hard to understand

provokateurin: Yeah, I'm struggling with this right now as it gets quite bad...

provokateurin:
Let me know if I should format the commands in a better way

PR #927, opened 2023-10-04
refactor: Cleanup ignore_for_file's

provokateurin:
ignore_for_file should be avoided as much as possible and instead the problems should be fixed where possible.


Leptopoda (approved):
I generally agree with this but I think that for unit tests `missing_override_of_must_be_overridden` and `avoid_implementing_value_types` could be handled on a file level.
I also don't think `inference_failure_on_collection_literal` is really helping here and makes code less readable. I find `Future<void>.delayed` quite ugly.

Nevertheless it's still a valid change so I leave it up to you if you also share my opinion.

PR #940, opened 2023-10-07
Add dashboard app

provokateurin:
Closes https://github.com/nextcloud/neon/issues/354

I wanted to add tests for the bloc as well, but I'm not sure how to best do that.
Something like https://pub.dev/packages/rx_bloc_test would be pretty neat, but in this case for example the problem is that the refresh method gets called in the constructor so the test is unable to control it from the outside.


provokateurin:
The weather status is not implemented in this PR because it's a different API, but it could be added in the future.

packages/neon/neon_dashboard/lib/src/widgets/widget_item.dart
```diff
@@ -0,0 +1,66 @@
...
+          leading,
+          SizedBox.square(
+            dimension: largeIconSize,
+            child: Align(
+              alignment: Alignment.bottomRight,
+              child: SizedBox.square(
+                dimension: smallIconSize,
+                child: NeonCachedImage.url(
+                  url: item.overlayIconUrl,
+                  account: NeonProvider.of<AccountsBloc>(context).activeAccount.value,
```
Leptopoda, line 44: Usage of the AccountsBloc is discouraged.

Do you think we can maybe extend `NeonApiImage` to fit here?

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
```diff
@@ -0,0 +1,113 @@
...
+                        ),
+                      );
+                    }
+
+                    return Align(
+                      child: FilledButton(
+                        onPressed: onPressed,
+                        child: label,
+                      ),
+                    );
```
Leptopoda, line 92: Separate widget or function please

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
```diff
@@ -0,0 +1,113 @@
...
+              ),
+              const SizedBox(
+                height: 20,
+              ),
+              if (halfEmptyContentMessage.isNotEmpty) _renderMessage(halfEmptyContentMessage),
+              if (emptyContentMessage.isNotEmpty) _renderMessage(emptyContentMessage),
+              if (halfEmptyContentMessage.isEmpty && emptyContentMessage.isEmpty && (items?.items.isEmpty ?? true))
+                _renderMessage(AppLocalizations.of(context).noEntries)
+              else
+                for (final item in items?.items ?? <dashboard.WidgetItem>[])
```
Leptopoda, line 58: the `?? []` is a bad performing way to avoid null ability.

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
```diff
@@ -0,0 +1,113 @@
...
+    required this.items,
+    super.key,
+  });
+
+  final dashboard.Widget widget;
+  final dashboard.WidgetItems? items;
+
+  @override
+  Widget build(final BuildContext context) {
+    final halfEmptyContentMessage = items?.halfEmptyContentMessage ?? '';
```
Leptopoda, line 21: another bad pattern to avoid dart null ability.

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
```diff
@@ -0,0 +1,113 @@
...
+    final emptyContentMessage = items?.emptyContentMessage ?? '';
+
+    return SizedBox(
+      width: 320,
+      height: 560,
+      child: Card(
+        child: InkWell(
+          onTap: widget.widgetUrl != null && widget.widgetUrl!.isNotEmpty ? () => context.go(widget.widgetUrl!) : null,
+          borderRadius: const BorderRadius.all(Radius.circular(12)),
+          child: ListView(
```
Leptopoda, line 31: Maybe just write a generator function where you can cleanly add multiple checks into a condition.
Or make this a `ListView.builder` that takes a builder function by default.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      if (v1Widgets.isNotEmpty) {
+        debugPrint('Loading v1 widgets: ${v1Widgets.join(', ')}');
+
+        final response = await _account.client.dashboard.dashboardApi.getWidgetItems(widgets: v1Widgets);
+        for (final entry in response.body.ocs.data.entries) {
+          widgets[widgets.keys.singleWhere((final widget) => widget.id == entry.key)] = dashboard.WidgetItems(
+            (final b) => b
+              ..items = entry.value.toBuilder()
```
Leptopoda, line 65:
```suggestion
              ..items.replace(entry.value) 
```

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
```diff
@@ -0,0 +1,90 @@
...
+          );
+        }
+      }
+
+      if (v2Widgets.isNotEmpty) {
+        debugPrint('Loading v2 widgets: ${v2Widgets.join(', ')}');
+
+        final response = await _account.client.dashboard.dashboardApi.getWidgetItemsV2(widgets: v2Widgets);
+        for (final entry in response.body.ocs.data.entries) {
+          widgets[widgets.keys.singleWhere((final widget) => widget.id == entry.key)] = entry.value;
```
Leptopoda, line 77: This is really ugly.
I'd rather have the id as the key of the map.

packages/neon/neon_dashboard/lib/src/pages/main.dart
```diff
@@ -0,0 +1,60 @@
...
+    super.initState();
+
+    bloc = NeonProvider.of<DashboardBloc>(context);
+  }
+
+  @override
+  Widget build(final BuildContext context) => ResultBuilder.behaviorSubject(
+        stream: bloc.widgets,
+        builder: (final context, final widgets) => Center(
+          child: ListView(
```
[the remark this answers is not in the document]
Leptopoda: EDIT: looks like the `ListView` and `Wrap` are used to achieve a `GridView`.
Iirc `NeonListView.custom` should also accept `SliverGrid`s

Leptopoda (commented):
I started the review last week so sorry if it is a bit messy.

packages/neon/neon_dashboard/lib/src/pages/main.dart
```diff
@@ -0,0 +1,60 @@
...
+              ),
+              NeonLinearProgressIndicator(
+                visible: widgets.isLoading,
+              ),
+              if (widgets.hasData)
+                Wrap(
+                  alignment: WrapAlignment.center,
+                  spacing: 8,
+                  runSpacing: 8,
+                  children: widgets.requireData.entries
```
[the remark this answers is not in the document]
provokateurin: Renamed to `snapshot`, but since `snapshot.requireData` only ever used once I won't add a variable for it.

packages/neon/neon_dashboard/lib/src/widgets/widget_item.dart
provokateurin: Didn't you refactor this so passing the active account isn't necessary anymore? I wanted to use that, but couldn't find it :thinking:

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
provokateurin: This is on purpose because the halfEmptyContentMessage is an empty string if is "empty". This way we only need to check if the string is empty or not instead of first checking if the items are null and then checking if the string is empty.

packages/neon/neon_dashboard/lib/src/widgets/widget_item.dart
Leptopoda: We have the mentioned NeinApiImage but that can only use api endpoints (DynamiteRawResponses).

But maybe we can sorta extend it.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
provokateurin: But we need to have it as a map of widgets and their respective items. I improved it by introducing an extension for finding the widgets by id.

packages/neon/neon_dashboard/lib/src/pages/main.dart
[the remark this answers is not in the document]
provokateurin: Well it should be a Wrap so we don't need to calculate how many items need to go in a row, but it would be possible to calculate this ourselves if you prefer that.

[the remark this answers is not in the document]
Leptopoda: There is a GridView fixed extend widget that forces a fixed width for the children and automatically calculates the fitting number of children.

You probably only thought of the fixed child count sliver builder.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
Leptopoda: Maybe a map with the id as the key and a record of widget and entry as value?
Or just have two normal maps and merge them before returning.

In the worst case the current code takes `n²` while having two maps and combining them later is in `2n`. I don't expect us to notice any performance benefit as we the number of widgets will be relatively small but lets make it cleaner from the beginning.

packages/neon/neon_dashboard/lib/src/pages/main.dart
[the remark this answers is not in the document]
Leptopoda: I think the following should help
https://api.flutter.dev/flutter/widgets/SliverGrid/SliverGrid.extent.html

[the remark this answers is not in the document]
provokateurin: I'm not sure if I like it. The effect looks very ugly when resizing and everything is already on the same row. Also I don't really want it to change the size of the cards. They should just always have the same size.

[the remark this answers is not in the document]
provokateurin: Please try out the current state. I'm not sure how to improve it so that is not this ugly to watch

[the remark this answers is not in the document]
Leptopoda: I can see; In this case flow is probably the right widget. Sorry for the confusion.
Nevertheless please use a `NeonListView` with a `SliverToBoxAdapter` like:

```dart
    return ResultBuilder.behaviorSubject(
      stream: bloc.widgets,
      builder: (final context, final snapshot) {
        Widget? child;
        if (snapshot.hasData) {
          child = Wrap(
            alignment: WrapAlignment.center,
            spacing: 8,
            runSpacing: 8,
            children: snapshot.requireData.entries
                .map(
                  (final widget) => DashboardWidget(
                    widget: widget.key,
                    items: widget.value,
                  ),
                )
                .toList(),
          );
        }

        return Center(
          child: NeonListView.custom(
            isLoading: snapshot.isLoading,
            error: snapshot.error,
            onRefresh: bloc.refresh,
            sliver: SliverToBoxAdapter(
              child: child,
            ),
          ),
        );
      },
    );
```

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
Leptopoda: Before we go back and for even more here is what I envisioned:
IMO this is more readable than the current approach and better than the old version.

```dart
import 'package:go_router/go_router.dart';
import 'package:neon/theme.dart';
import 'package:neon/widgets.dart';
import 'package:neon_dashboard/l10n/localizations.dart';
import 'package:neon_dashboard/src/widgets/widget_button.dart';
import 'package:neon_dashboard/src/widgets/widget_item.dart';
import 'package:nextcloud/dashboard.dart' as dashboard;

class DashboardWidget extends StatelessWidget {
  const DashboardWidget({
    required this.widget,
    required this.items,
    super.key,
  });

  final dashboard.Widget widget;
  final dashboard.WidgetItems? items;

  @override
    final halfEmptyContentMessage = _renderMessage(items?.halfEmptyContentMessage);
    final emptyContentMessage = _renderMessage(items?.emptyContentMessage);

    return Card(
      child: InkWell(
        onTap: widget.widgetUrl != null && widget.widgetUrl!.isNotEmpty ? () => context.go(widget.widgetUrl!) : null,
        borderRadius: const BorderRadius.all(Radius.circular(12)),
        child: ListView(
          padding: const EdgeInsets.all(8),
          children: [
            ListTile(
              title: Text(
                widget.title,
                style: const TextStyle(
                  fontWeight: FontWeight.bold,
                ),
              ),
              leading: SizedBox.square(
                dimension: largeIconSize,
                child: NeonCachedImage.url(
                  url: widget.iconUrl,
                  svgColorFilter: ColorFilter.mode(Theme.of(context).colorScheme.primary, BlendMode.srcIn),
                  size: const Size.square(largeIconSize),
                ),
              ),
            ),
            const SizedBox(
              height: 20,
            ),
            if (halfEmptyContentMessage != null) halfEmptyContentMessage,
            if (emptyContentMessage != null) emptyContentMessage,
            if (halfEmptyContentMessage == null && emptyContentMessage == null)
              _renderMessage(AppLocalizations.of(context).noEntries)!,
            ...?items?.items.map(
              (final item) => DashboardWidgetItem(
                item: item,
                roundIcon: widget.itemIconsRound,
              ),
            ),
            const SizedBox(
              height: 20,
            ),
            ...?widget.buttons?.map(
              (final button) => DashboardWidgetButton(
                button: button,
              ),
            ),
          ],
        ),
      ),
    );
  }

  Widget? _renderMessage(final String? message) {
    if (message == null || message.isEmpty) {
      return null;
    }

    return Center(
      child: Column(
        children: [
          const Icon(
            Icons.check,
            size: largeIconSize,
          ),
          Text(message),
        ],
      ),
    );
  }
}

```

packages/neon/neon_dashboard/lib/src/widgets/widget.dart
provokateurin: I would keep it how it is now. There is no benefit in using a builder because all items will always be displayed at once and there won't be many at once anyway (it's limited to 7 by default).

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
provokateurin: Can you explain a little more how you want to use two maps and then merge them later? I'm not sure I understand what you want (although I understand why).

Leptopoda: I must have mixed up something in an earlier review as I also can't see the change I wanted anymore.

provokateurin: I changed it a bit already

packages/neon/neon_dashboard/lib/src/widgets/widget_item.dart
provokateurin: Solved in https://github.com/nextcloud/neon/pull/969

provokateurin:
Depends on https://github.com/nextcloud/neon/pull/969

Leptopoda:
@provokateurin is this up for re review or should I wait until #969 is merged?

provokateurin:
Now it is, but please ignore the two commits from https://github.com/nextcloud/neon/pull/969.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
```diff
@@ -0,0 +1,101 @@
...
+      widgets.add(Result.error(e));
+      return;
+    }
+  }
+}
+
+extension WidgetFind on Iterable<dashboard.Widget> {
+  dashboard.Widget? tryFind(final String? id) => firstWhereOrNull((final widget) => widget.id == id);
+  dashboard.Widget find(final String id) => firstWhere((final widget) => widget.id == id);
+}
```
Leptopoda, line 101: Please move this to `utils` add unit tests.
Should be fairly easy to just copy the ones we already have for this.

packages/neon/neon_dashboard/lib/src/pages/main.dart
```diff
@@ -0,0 +1,51 @@
...
+                    items: widget.value,
+                  ),
+                )
+                .toList(),
+          );
+        }
+
+        return Center(
+          child: NeonListView.custom(
+            shrinkWrap: true,
```
Leptopoda, line 39: I still don't understand why we need shrinkwrap.
The children (`DashboardWidget` in this case) have a fixed size.

Or are there use cases that I just don't see?

packages/neon/neon/lib/src/widgets/list_view.dart
Leptopoda, line 1: I don't feel like we should support shrinkwrap.

It will only be needed when a NeonListView is places inside another scroll able widget and that just shouldn't happen.

packages/neon/neon_dashboard/analysis_options.yaml
Leptopoda, line 1: Sorry that I didn't see it earlier but we should use `include: package:neon_lints/src/base.yaml` to also force us to document everything.

This makes it much easier for other devs to understand what's going on.

Leptopoda (changes requested):

packages/neon/neon_dashboard/lib/src/pages/main.dart
provokateurin: The whole list should be centered on the page. You can't center the items of a list, so shrinkWrap and Center is needed to center the whole list.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
provokateurin: I was thinking of creating a class Findable which every thing that we `find` can implement. Then we don't need to duplicate these extensions and can just have one for all.

Leptopoda: I agree that a common interface is better but the class in question is provided by the nextcloud package so it wouldn't help here.

packages/neon/neon_dashboard/lib/src/pages/main.dart
Leptopoda: In this case
```dart
        return NeonListView.custom(
          isLoading: snapshot.isLoading,
          error: snapshot.error,
          onRefresh: bloc.refresh,
          sliver: SliverFillRemaining(
            hasScrollBody: false,
            child: Center(
              child: child,
            ),
          ),
        );
```

Should also do the trick.

provokateurin: Oh I see, I will try it out. I should definitely get more into slivers.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
provokateurin: Ah true

packages/neon/neon_dashboard/analysis_options.yaml
provokateurin: https://github.com/nextcloud/neon/blob/main/docs/workflows.md doesn't say what linting rules should be used and how to actually add them. Can you document this there (in your docs PR)?

Leptopoda: I think we should rather just drop the rules that disable the documentation.
The only package that doesn't need documentation is `app` and even there it is helpful for other devs to take a look.

packages/neon/neon_dashboard/lib/src/pages/main.dart
provokateurin: This works like a charm! Super nice :heart:

packages/neon/neon_dashboard/lib/src/app.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:neon_dashboard/src/blocs/dashboard.dart';
+import 'package:neon_dashboard/src/options.dart';
+import 'package:neon_dashboard/src/pages/main.dart';
+import 'package:neon_dashboard/src/routes.dart';
+import 'package:nextcloud/core.dart' as core;
+import 'package:nextcloud/nextcloud.dart';
+
+/// Implementation of the server `dashboard` app.
+class DashboardApp extends AppImplementation<DashboardBloc, DashboardAppSpecificOptions> {
+  /// Creates a new Dashboard app implementation instance
```
Leptopoda, line 14:
```suggestion
  /// Creates a new Dashboard app implementation instance.
```

Let's get the low hanging fruit :)

Leptopoda (approved):
Other than that LGTM

PR #943, opened 2023-10-07
feat(neon_framework): Support custom backgrounds

provokateurin:
Closes https://github.com/nextcloud/neon/issues/939

I'm pretty sure this whole feature but also the way I implemented it is going to be quite controversial, but please go ahead reviewing :)

### DIFF
--- a/packages/neon_framework/lib/src/widgets/custom_background.dart
+++ b/packages/neon_framework/lib/src/widgets/custom_background.dart
@@ -1,0 +1,69 @@
+import 'package:flutter/material.dart';
+import 'package:neon_framework/src/theme/server.dart';
+import 'package:neon_framework/src/utils/global_options.dart';
+import 'package:neon_framework/src/utils/hex_color.dart';
+import 'package:neon_framework/src/utils/provider.dart';
+import 'package:neon_framework/src/widgets/image.dart';
+import 'package:neon_framework/src/widgets/options_collection_builder.dart';
+
+/// Displays the custom background if the user has it enabled.
+///
+/// The background will be loaded from the server.
+/// It might be a single color or a full image.
+class NeonCustomBackground extends StatelessWidget {
+  const NeonCustomBackground({
+    required this.child,
+    super.key,
+  });
+
+  /// Widget displayed on top of the custom background.
+  final Widget child;
+
+  @override
+  Widget build(final BuildContext context) {
+    final globalOptions = NeonProvider.of<GlobalOptions>(context);
+
+    return OptionsCollectionBuilder(
+      valueListenable: globalOptions,
+      child: child,
+      builder: (final context, final options, final child) {
+        if (!options.themeUseNextcloudTheme.value || !options.themeCustomBackground.value) {
+          return child!;
+        }
+
+        final theme = Theme.of(context).extension<ServerTheme>()!.nextcloudTheme;
+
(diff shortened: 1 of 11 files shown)


Leptopoda (changes requested):
I haven't looked much at the code but it looks fine at a first glance.
But the design is just awe full and unacceptable. 

I really struggle reading any text displayed so that's also a hard no go regarding #300.

Leptopoda:
I think we should not have a background behind the main app content but i quite like how the nextcloud web interface has the image behind the drawer (with sufficient contrast!!!).
The contrast is somewhat acceptable when both enabling dark mode and the oled override.
As the PR isn't rebased yet so I couldn't have a look at the dashboard app yet.

I think if the option is enabled we should only show the image behind the drawer. The dashboard app could then make its scaffold background transparent.
Each client developer should decide this on their own without a global rule.

provokateurin:
To implement this it would be great if there is a way to get cropped part of the background image. We don't want the image to repeat itself so all widgets would need to take the part of the image of their position as the background. I don't know if there is an easy way to achieve it, but using transparency like right now is definitely not good because it will break easily.

provokateurin:
The transparency approach is hard to implement and adds unnecessary code. "Punching a hole" would be way cleaner and easier to use everywhere.

provokateurin:
For now I changed it so that it is a widget that can be used independently. Only used in the dashboard right now, but we can expand it later.

packages/neon_framework/lib/src/widgets/custom_background.dart
```diff
@@ -0,0 +1,81 @@
...
+  final Widget child;
+
+  @override
+  Widget build(final BuildContext context) {
+    final globalOptions = NeonProvider.of<GlobalOptions>(context);
+    final accountsBloc = NeonProvider.of<AccountsBloc>(context);
+
+    return OptionsCollectionBuilder(
+      valueListenable: globalOptions,
+      builder: (final context, final options, final _) {
```
Leptopoda, line 30: Please provide the child widget to the child parameter and use it in the builder.
This avoids rebuilding the entire subtree if only the option changes.

packages/neon_framework/lib/src/widgets/custom_background.dart
```diff
@@ -0,0 +1,81 @@
...
+
+        return StreamBuilder(
+          stream: accountsBloc.activeAccount,
+          builder: (final context, final activeAccountSnapshot) =>
+              ResultBuilder<core.OcsGetCapabilitiesResponseApplicationJson_Ocs_Data?>.behaviorSubject(
+            subject: activeAccountSnapshot.hasData
+                ? accountsBloc.getCapabilitiesBlocFor(activeAccountSnapshot.data!).capabilities
+                : null,
+            builder: (final context, final capabilitiesSnapshot) {
+              final theme = capabilitiesSnapshot.data?.capabilities.themingPublicCapabilities?.theming;
```
Leptopoda, line 43: The server theme is already queried on startup.
I think we should insert it as a theme extension in the [AppTheme](https://github.com/nextcloud/neon/blob/119dfe912c16fd0cd82dcafb790d9547c8efd258/packages/neon_framework/lib/src/theme/theme.dart).

Leptopoda (approved):

PR #886, opened 2023-09-30
Refactor/neon

Leptopoda:
Various other refactors from #745

PR #959, opened 2023-10-12
refactor(neon,nextcloud): cleanup PushNotification decryption

Leptopoda:
extracted from #886

I really dislike this approach. I think we should just have a second method like `PushNotification.fromEncrypted` similar to the fromJson one. We could also do something similar with `DecryptedSubject`.

What do you think.

### DIFF
--- a/packages/neon/neon/lib/src/models/push_notification.dart
+++ b/packages/neon/neon/lib/src/models/push_notification.dart
@@ -1,13 +1,26 @@
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
-import 'package:nextcloud/notifications.dart' show DecryptedSubject;
+import 'package:nextcloud/notifications.dart' as notifications;
 
 part 'push_notification.g.dart';
 
+/// The json key for [PushNotification.accountID].
+const String _accountIDKey = 'accountID';
+
+/// The json key for [PushNotification.priority].
+const String _priorityKey = 'priority';
+
+/// The json key for [PushNotification.type].
+const String _typeKey = 'type';
+
+/// The json key for [PushNotification.subject].
+const String _subjectKey = 'subject';
+
+/// Data for a push notification.
 @JsonSerializable()
-@immutable
 @internal
 class PushNotification {
+  /// Creates a new push notification.
   const PushNotification({
     required this.accountID,
     required this.priority,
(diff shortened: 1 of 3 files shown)


provokateurin:
Is this supposed to be a draft?

Leptopoda:
It's up for review and as I'm not sure if I lile this approach I made it a draft.

provokateurin:
Well the review is there to discuss the approach :)

packages/neon/neon/lib/src/models/push_notification.dart
```diff
@@ -1,28 +1,119 @@
...
     required this.subject,
   });
 
-  factory PushNotification.fromJson(final Map<String, dynamic> json) => _$PushNotificationFromJson(json);
-  Map<String, dynamic> toJson() => _$PushNotificationToJson(this);
+  /// Creates a new PushNotification object from the given [json] data.
+  factory PushNotificationEncrypted.fromJson(
+    final Map<String, dynamic> json,
+  ) {
+    final privateKey = PushUtils.loadRSAKeypair();
```
provokateurin, line 90: If you want to do it this way please pass in the private key instead of calling this logic in the model.

provokateurin (changes requested):
I also don't like this approach. Let's go with your `PushNotification.fromEncrypted` idea, that sounds much better. Please don't call the logic for loading the private key inside that method and instead just pass the private key to it.

packages/neon/neon/lib/src/models/push_notification.dart
```diff
@@ -1,28 +1,72 @@
...
+const String _typeKey = 'type';
+
+/// The json key for [PushNotification.subject].
+const String _subjectKey = 'subject';
+
+/// Data for a push notification.
 @JsonSerializable()
-@immutable
+@internal
 @internal
```
provokateurin, line 22: Duplicate internal

packages/neon/neon/lib/src/models/push_notification.dart
```diff
@@ -1,28 +1,72 @@
...
+const String _accountIDKey = 'accountID';
+
+/// The json key for [PushNotification.priority].
+const String _priorityKey = 'priority';
+
+/// The json key for [PushNotification.type].
+const String _typeKey = 'type';
+
+/// The json key for [PushNotification.subject].
+const String _subjectKey = 'subject';
```
provokateurin, line 17: This seems useless now that there is only a single class with those fields.

Leptopoda: No it isn't.
I want to ensure the same key is used for both the encrypted and decrypted serializaiton.
Therefore it's still used in the ´fromEncrypted` method.

provokateurin: Ahhh, I oversaw that.

provokateurin (approved):

PR #974, opened 2023-10-16
fix(nextcloud): fix appid sorting

Leptopoda:


packages/nextcloud/bin/generate_exports.dart
```diff
@@ -5,18 +5,24 @@ import 'package:dynamite/src/helpers/dart_helpers.dart';
...
 
   for (final file in files) {
     final basename = p.basename(file.path);
     final id = basename.substring(0, basename.length - 13);
     final variablePrefix = toDartName(id);
     final classPrefix = toDartName(id, uppercaseFirstCharacter: true);
 
-    idStatements.add("  static const $variablePrefix = '$id';");
+    idStatements
+      ..writeln('  /// ID for the $variablePrefix app.')
```
provokateurin, line 24:
```suggestion
      ..writeln('  /// ID for the $id app.')
```

This should be the ID because the variablePrefix is just a converted version that is not the real name.

provokateurin (changes requested):

provokateurin (approved):

PR #977, opened 2023-10-16
docs(neon): unify and document Result and ResultBuilder

Leptopoda:


packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -40,18 +71,43 @@ class Result<T> {
...
+  /// If this is non-null, [hasData] will be true.
+  ///
+  /// If the data fetching computation has never returned a value, this may be
+  /// set to an initial data value specified by the relevant widget. See
+  /// [ResultBuilder.initialData].
   final T? data;
+
+  /// The latest error object received by the data fetching computation.
+  ///
+  /// If [data] is not null, this will be null.
```
provokateurin, line 85: This is wrong. It can have both data and error not null at the same time.

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -40,18 +71,43 @@ class Result<T> {
...
   final Object? error;
+
+  /// The loading state of the data fetching.
```
provokateurin, line 88:
```suggestion
  /// Whether new [data] is being fetched right now.
```

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -40,18 +71,43 @@ class Result<T> {
...
+  ///
+  /// A loading result may have cached data.
   final bool isLoading;
+
+  /// Whether the data is [data] is fetched from cache.
```
provokateurin, line 93:
```suggestion
  /// Whether the [data] is fetched from cache.
```

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -40,18 +71,43 @@ class Result<T> {
...
+  Result<R> transform<R>(final R? Function(T data) callback) => Result(
+        hasData ? callback(data as T) : null,
         error,
         isLoading: isLoading,
         isCached: isCached,
       );
 
+  /// Copies this Result in a loading state.
+  ///
+  /// Marks the data as cached if present.
```
provokateurin, line 110: No?

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -67,12 +123,29 @@ class Result<T> {
         isCached: isCached ?? this.isCached,
       );
 
+  /// Returns whether this snapshot contains a non-null [error] value.
+  ///
+  /// This is always true if the asynchronous computation's last result was
+  /// failure.
```
provokateurin, line 129: I don't think this is helpful and somewhat misleading because there are also other reasons why it might be true.

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -67,12 +123,29 @@ class Result<T> {
...
+  /// A result may both have an error and data.
   bool get hasError => error != null;
 
+  /// Returns whether this snapshot contains a non-null [data] value.
+  ///
+  /// This can be false even when the asynchronous computation has completed
+  /// successfully, if the computation did not return a non-null value. For
+  /// example, a [Future<void>] will complete with the null value even if it
+  /// completes successfully.
```
provokateurin, line 138: I feel like this is rather obvious and implied. If T is a nullable type, then this is totally expected.

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -67,12 +123,29 @@ class Result<T> {
...
+  /// A result may both have an error and data.
   bool get hasData => data != null;
 
+  /// Returns whether this snapshot [hasData] and [isCached] is not true.
   bool get hasUncachedData => hasData && !isCached;
 
+  /// Returns latest data received, failing if there is no data.
+  ///
+  /// Throws [error], if [hasError]. Throws [StateError], if neither [hasData]
+  /// nor [hasError].
```
provokateurin, line 148: First part is not true. Only StateError part is true

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -88,3 +161,94 @@ class Result<T> {
   @override
   int get hashCode => Object.hash(data, error, isLoading, isCached);
 }
+
+/// Signature for strategies that build widgets based on asynchronous
+/// [Result]s.
```
provokateurin, line 166: Single line

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -88,3 +161,94 @@ class Result<T> {
...
+
+  /// Builder function called with the current [Result] value.
+  ///
+  /// This builder must only return a widget and should not have any side
+  /// effects as it may be called multiple times.
+  final ResultWidgetBuilder<T> builder;
+
+  /// The data that will be used to create the initial result.
+  ///
+  /// Providing this value (presumably obtained synchronously somehow when the
```
provokateurin, line 217:
```suggestion
  /// Providing this value (presumably obtained synchronously when the
```

provokateurin (changes requested):

packages/neon/neon/lib/src/bloc/result.dart
```diff
@@ -40,18 +71,43 @@ class Result<T> {
           'Result.error() called without specifying the data type. Call Result<T>.error() instead',
         );
 
+  /// The latest data received by the data fetching.
+  ///
+  /// If this is non-null, [hasData] will be true.
+  ///
+  /// If the data fetching computation has never returned a value, this may be
+  /// set to an initial data value specified by the relevant widget. See
```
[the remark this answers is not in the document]
Leptopoda: It's perfectly fine to use the `Result` outside the `ResultBuilder` so I'd rather keep it generalized.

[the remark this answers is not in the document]
provokateurin: Yes, but you can also use it outside of widgets. It sounds like it would be required to use it with widgets.

packages/neon/neon/lib/src/bloc/result.dart
Leptopoda: I agree that this might be expected but I don't think it's obvious without looking at the code.

If we publish neon on pub.dev the documentation will be hosted with it. Someone browsing the api reference without looking at the code might find this information helpful.

provokateurin (approved):

PR #980, opened 2023-10-16
feat(nextcloud): Add spreed

provokateurin:
Towards https://github.com/nextcloud/neon/pull/366


provokateurin:
The last patch is applied from https://github.com/nextcloud/spreed/pull/10804 and this PR thus depends on that PR.

provokateurin:
I assume https://github.com/nextcloud/spreed/pull/10805 will be merged without objections as well. Just an easy rename of an endpoint to make it easier to know what the endpoint does, so we should include it too

tool/generate-specs.sh
```diff
@@ -50,6 +50,14 @@ done
   generate_spec "." "notifications"
 )
 
+for spec in packages/nextcloud/lib/src/api/*.openapi.json; do
+  name="$(basename "$spec" | cut -d "." -f 1)"
+  if [ -f "tool/patches/$name.json" ]; then
```
Leptopoda, line 59: I'd change the format to `tool/patches/$appName/$patchName.json` this way we can support multiple patches per spec.

This way we can keep the patches themselves smaller and easier to understand.

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,145 @@
...
+  inCall(1),
+  providesAudio(2),
+  providesVideo(4),
+  usesSIPDialIn(8);
+
+  const InCallFlag(this.code);
+  final int code;
+}
+
+class InCallFlags {
```
Leptopoda, line 116: Same as above.

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,145 @@
...
+  startCall(2),
+  joinCall(4),
+  canIgnoreLobby(8),
+  canPublishAudio(16),
+  canPublishVideo(32),
+  canScreenShare(64),
+  canSendMessageAndShareAndReact(128);
+
+  const ParticipantPermission(this.code);
+  final int code;
```
Leptopoda, line 71: Please again use bit operations here.
Just make it a getter that shifts `1` by the index of the element.

```dart
int get value => 1 << index; 
```

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,145 @@
...
+  stub,
+  owner,
+  moderator,
+  user,
+  guest,
+  userFollowingPublicLink,
+  guestWithModeratorPermissions,
+}
+
+enum ParticipantPermission {
```
Leptopoda, line 59: If we decide to use the proper way with bit operators this should be documented.
I think this is way easier to understand.

Leptopoda (changes requested):
Please also document everything in `packages/nextcloud/lib/src/helpers/spreed.dart`

CONTRIBUTING.md
```diff
@@ -13,6 +13,7 @@ Note that you need to have Dart installed and [`~/.pub-cache/bin/` needs to be i
 
 You will need to have the following packages installed:
 - [yq](https://github.com/kislyuk/yq)
+- [jsonpatch](https://pypi.org/project/jsonpatch)
```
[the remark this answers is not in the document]
provokateurin: True, let's add a separate section

tool/generate-specs.sh
provokateurin: I wanted to keep it simple for the beginning, but it makes sense to already have this.

tool/generate-specs.sh
```diff
@@ -14,7 +14,7 @@ function generate_spec() {
 for dir in external/nextcloud-server external/nextcloud-notifications; do
   (
     cd "$dir"
-    composer update
+    composer install
     composer install --no-dev
```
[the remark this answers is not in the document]
provokateurin: Technically only the first on is needed, but it will leave behind some files if you have a different php or composer version. The second command basically cleans up those files. Otherwise we would always have diffs on the submodules which is annoying. The composer update was just wrong because it updated the lockfile which we don't want to do because it could pull in versions that we aren't supposed to use.

provokateurin:
The helpers code is now much improved (bitwise operations) and updated (missing enum values). To be fair it was just quickly thrown together when I wrote the Talk prototype, so it was not very good to begin with (which I should have fixed before creating the PR but yeah).

provokateurin:
The patch with the bug fix is still there because the PR wasn't merged yet, but the one with the small improvement is already in.

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  /// Integer representation of the [ParticipantType].
+  int get value => index + 1;
+}
+
+/// Attendee permissions.
+///
+/// Use [value] to get the integer representation that is used in the API.
+/// See https://github.com/nextcloud/spreed/blob/master/lib/Model/Attendee.php.
+enum ParticipantPermission {
```
Leptopoda, line 144: Please mention and explain both the `fromValue` method and the `ParticipantPermissionsValue` extension,how and why they are used.

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,199 @@
...
+      permissions.add(ParticipantPermission.$default);
+    }
+
+    return permissions;
+  }
+}
+
+/// Extension for the integer representation of multiple [ParticipantPermission]s.
+extension ParticipantPermissionsValue on Set<ParticipantPermission> {
+  /// Integer representation of multiple [ParticipantPermission]s.
```
Leptopoda, line 197:
```suggestion
  /// Gets the integer representation of multiple [ParticipantPermission]s.
```

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,199 @@
...
+  /// Note to self.
+  noteToSelf;
+
+  /// Integer representation of the [ParticipantType].
+  int get value => index + 1;
+}
+
+/// Types of chat messages.
+///
+/// Use [name] to get the string representation that is used in the API.
```
Leptopoda, line 51: Not implemented that way.

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,199 @@
...
+  reaction,
+
+  /// Deleted emoji reaction.
+  // ignore: constant_identifier_names
+  reaction_deleted;
+}
+
+/// Actor types of chat messages.
+///
+/// Use [name] to get the string representation that is used in the API.
```
Leptopoda, line 81: again wrong

Leptopoda (changes requested):

packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -0,0 +1,199 @@
...
+
+  /// Former "One to one" (When a user is deleted from the server or removed from all their conversations, `1` "One to one" rooms are converted to this type).
+  oneToOneFormer,
+
+  /// Note to self.
+  noteToSelf;
+
+  /// Integer representation of the [ParticipantType].
+  int get value => index + 1;
+}
```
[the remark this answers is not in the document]
provokateurin: The doc comments for the enum values are copied straight from the Talk docs.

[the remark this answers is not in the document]
provokateurin: But yeah I can write better ones if you want to have them.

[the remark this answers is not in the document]
Leptopoda: Yes please :)

packages/nextcloud/lib/src/helpers/spreed.dart
provokateurin: No this is right? See the unit tests where this is used.

packages/nextcloud/lib/src/helpers/spreed.dart
provokateurin: Same as above

packages/nextcloud/lib/src/helpers/spreed.dart
Leptopoda: My bad. This time I forgot about the `name` getters on enums :sweat_smile:

Leptopoda (approved):
:tada:

provokateurin:
Noice, now we just need to wait for https://github.com/nextcloud/spreed/pull/10804 which will probably be merged tomorrow

packages/nextcloud/lib/src/api/spreed.openapi.json
```diff
@@ -0,0 +1,16719 @@
...
+                        "format": "int64"
+                    },
+                    "lastMessage": {
+                        "oneOf": [
+                            {
+                                "$ref": "#/components/schemas/ChatMessage"
+                            },
+                            {}
+                        ]
+                    },
```
Leptopoda, line 954: The spec is broken as a ChatMessage also matches an empty object

packages/nextcloud/lib/src/api/spreed.openapi.json
```diff
@@ -0,0 +1,16719 @@
...
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "type": "object",
+                                    "required": [
+                                        "ocs"
+                                    ],
+                                    "properties": {
+                                        "ocs": {
+
```
Leptopoda, line 12864: same here.

Leptopoda (changes requested):
I just tested #1001 on this branch and it fails.

packages/nextcloud/lib/src/api/spreed.openapi.json
provokateurin: The empty array problem strikes again...

provokateurin:
Will be fixed with https://github.com/nextcloud/spreed/pull/10810. Then we have some ugly empty arrays again (only those appearing in oneOf which was the problematic part), but there is no other way to do this properly.

provokateurin:
@Leptopoda we have another problem now. BuiltList happily validates to the raw value stored in `data`. So the same problem is still there, but it's now on our side because this shouldn't happen. Do you have an idea how we can work around this? I pushed the current state so you can try it with the correctly generated specs.

provokateurin:
Talk PRs are merged, now we only have the problem with oneOf left.

provokateurin:
To workaround the bugs we will patch the problematic parts of the specs for now (see https://github.com/nextcloud/neon/pull/1057#issuecomment-1785120763 for discussion).

Leptopoda (approved):

provokateurin:
🎉

PR #981, opened 2023-10-16
Fix/dynamite/numbers

Leptopoda:
This is needed for cases where we can't make the default parameters default like the upcoming `Int64` migration.


Leptopoda:
Closing this one as the `Int64` migration won't happen because of limitations in the dart provided JsonConverter.

I'll open separate PRs for the refactors I made in dynamite.

PR #984, opened 2023-10-17
fix(neon_files): empty browser view

Leptopoda:
How didn't we realize that the files view was broken :thinking: 
I think this was introduced with the NeonListView rework.

### DIFF
--- a/packages/neon/neon_files/lib/widgets/browser_view.dart
+++ b/packages/neon/neon_files/lib/widgets/browser_view.dart
@@ -123,6 +123,7 @@ class _FilesBrowserViewState extends State<FilesBrowserView> {
                               bloc: widget.filesBloc,
                               browserBloc: widget.bloc,
                               details: details,
+                              mode: widget.mode,
                             );
                           },
                           isLoading: filesSnapshot.isLoading,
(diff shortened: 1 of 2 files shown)


Leptopoda:
Ok there is still a bug with moving a file

Leptopoda:
Ok should be working now.
@provokateurin please take the time to test file moving, folder selection, and both in conjunction with the hidden option.

I hope I didn't miss anything this time.

packages/neon/neon_files/lib/widgets/browser_view.dart
```diff
@@ -47,9 +47,25 @@ class _FilesBrowserViewState extends State<FilesBrowserView> {
...
-                ? const SizedBox()
-                : BackButtonListener(
+            builder: (final context, final tasksSnapshot) {
+              if (!pathSnapshot.hasData || !tasksSnapshot.hasData) {
+                return const SizedBox();
+              }
+              return ValueListenableBuilder(
+                valueListenable: widget.bloc.options.showHiddenFilesOption,
+                builder: (final context, final showHiddenFiles, final _) {
+                  files.data?.removeWhere((final file) {
```
provokateurin, line 57: This doesn't work correctly when you enable the option again because the hidden files were removed from the data.

provokateurin (changes requested):

packages/neon/neon_files/lib/widgets/browser_view.dart
```diff
@@ -47,9 +47,25 @@ class _FilesBrowserViewState extends State<FilesBrowserView> {
...
+                    if (widget.mode == FilesBrowserMode.selectDirectory) {
+                      return !file.isDirectory;
+                    }
```
provokateurin, line 60: I think this should respect the hidden option as well. This is how it normally works in any file manager I know.

provokateurin (changes requested):

Leptopoda:
While being at it I also fixed two other things :)

provokateurin (approved):

Issue #535, opened 2023-08-07
Cleanup and unify dialogs


provokateurin:
We use lots of different dialogs and most don't respect the "new" constraints. Cleanup all of them and make them respect the constraints.

Leptopoda:
We should also use the new adaptive dialog from flutter 3.13 where possible.

PR #985, opened 2023-10-18
refactor l10n handling

Leptopoda:
In order to have common strings that can be reused and exported by the framework we need to have unique L10n classes.

As discussed at the conf.
The reuse of strings is needed for the work I'm doing on #535

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,0 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
-melos run format:check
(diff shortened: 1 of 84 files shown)


provokateurin (changes requested):
The classes should be named `Localizations` instead of `L10n`.
Other than that everything LGTM!

Leptopoda:
Why not L10n?
I quite like the shorter name.

Should we now also drop the format check commit hook? It just makes commits slower and with autoformat turned on in the IDE this shouldn't happen as freequent.

provokateurin:
For me L10n looks just ugly, but I also don't see a reason why we shouldn't just use the full name

Leptopoda:
> Yeah I also don't like the format-check hook, but you wanted to have it because you often forgot to format before pushing.

Ok then I'll remove it as the auto generation of the l10n won't happen anymore (it just broke with the changes and we have the melos command for that)

PR #998, opened 2023-10-19
Refactor/neon/dialogs

Leptopoda:
Fixes: #995 
Fixes: #535 

I moved the folder dialogs (create, rename, delete) into the framework as this can be reused in many clients.


provokateurin (changes requested):
I do not agree with the move to put app implementation specific dialogs into the framework. If we want to support this then it should be done through the interface abstractions.
I'm sorry, but I think you will need to revert large parts of the PR.

Leptopoda:
I guess you mean the folder dialogs?
I agree that the framework shouldn't hold client specific code but I also argue that the concept of folders (or labels/categories as they are called on notes) is a concept many apps need. I think that they should be reused.
I can pick it apart again. And we can discuss this in a separate issue. 

Can you please provide feedback to the design (as I think it changed for every dialog).
Similarly a rough review of `packages/neon/neon/lib/src/widgets/dialog.dart` would be great to know wether the widgets are good enough.

packages/neon/neon/lib/src/utils/dialog.dart
```diff
@@ -0,0 +1,113 @@
...
+/// Returns a future with the new name of name.
+Future<String?> showFolderRenameDialog({
+  required final BuildContext context,
+  required final String folderName,
+}) async =>
+    showRenameDialog(
+      context: context,
+      title: NeonLocalizations.of(context).folderRename,
+      initialValue: folderName,
+    );
```
provokateurin, line 113: IMO these do not belong here, all the other ones can stay.

provokateurin (commented):
I like the design a lot :heart:

I found some places that need to be improved:
image
image

I haven't tested all the dialogs yet, but most of them look very good!

packages/neon/neon/lib/src/utils/dialog.dart
provokateurin: Yeah I get that and I agree with you, but where do you draw the line? If it is the exact same code snippet?

Leptopoda:
@provokateurin ping :)

Leptopoda:
just a rebase for now.

packages/neon_framework/lib/src/widgets/dialog.dart
provokateurin, line 1: I know this is quite some work, but could you please add unit and golden tests for the dialogs? Confirmation, rename and error dialogs would be enough for now

Leptopoda: I refuse adding golden tests as they are less flexible and normal widget tests cover the same.

packages/neon_framework/test/dialog_test.dart
```diff
@@ -0,0 +1,171 @@
...
+
+      testWidgets('NeonConfirmationDialog decline', (final widgetTester) async {
+        const title = 'My Title';
+        const dialog = NeonConfirmationDialog(title: title);
+        await widgetTester.pumpWidget(wrapDialog(dialog));
+
+        await widgetTester.tap(find.byType(ElevatedButton));
+        verify<void>(() => mockedObserver.didPop(any(), any())).called(1);
+      });
+    });
```
provokateurin, line 99: These should test that the expected values are popped and not just any

packages/neon_framework/test/dialog_test.dart
```diff
@@ -0,0 +1,171 @@
...
+      });
+
+      testWidgets('NeonRenameDialog actions', (final widgetTester) async {
+        const title = 'My Title';
+        const value = 'My value';
+        const dialog = NeonRenameDialog(title: title, value: value);
+
+        // Equal value should not submit
+        await widgetTester.pumpWidget(wrapDialog(dialog));
+        await widgetTester.enterText(find.byType(TextFormField), '');
```
provokateurin, line 120: Not needed to type an empty string

packages/neon_framework/test/dialog_test.dart
```diff
@@ -0,0 +1,171 @@
...
+        await widgetTester.tap(find.byType(NeonDialogAction));
+        verifyNever<void>(() => mockedObserver.didPop(any(), any()));
+
+        // Different value should submit
+        await widgetTester.pumpWidget(wrapDialog(dialog));
+        await widgetTester.enterText(find.byType(TextFormField), 'My new value');
+        await widgetTester.tap(find.byType(NeonDialogAction));
+        verify<void>(() => mockedObserver.didPop(any(), any())).called(1);
```
provokateurin, line 128: Also check that the new value is popped

packages/neon_framework/test/dialog_test.dart
```diff
@@ -0,0 +1,171 @@
...
+
+        await widgetTester.tap(find.byType(NeonDialogAction));
+        verify<void>(() => mockedObserver.didPop(any(), any())).called(1);
+      });
+    });
+
+    group('NeonDialog', () {
+      testWidgets('NeonDialog', (final widgetTester) async {
+        const dialog = NeonDialog();
+        await widgetTester.pumpWidget(wrapDialog(dialog, TargetPlatform.macOS));
```
provokateurin, line 163: Why macOS?

provokateurin (changes requested):

packages/neon_framework/test/dialog_test.dart
Leptopoda: On cupertino platforms a cancel button is needed as dialogs can not be dismissed by tapping outside of them.
This tests handles this.

I chose macOS to hit both iOS and macOS in the switch of `NeonDialogAction`.

packages/neon_framework/test/dialog_test.dart
Leptopoda: ah wrong description. It's to explicitly check that "Empty values should not submit"

packages/neon_framework/test/dialog_test.dart
provokateurin: I see, then please name the test accordingly

packages/neon_framework/test/dialog_test.dart
Leptopoda: Please show me how.
I already did quite some research and was not able to find a solution to this.
The any only matches the routes and not the attached values.

provokateurin: Looking at https://github.com/flutter/flutter/blob/b0b0e423d6ec5f115e0580f9e21ffed9607dd302/packages/flutter/test/material/dialog_test.dart#L363 it seems like you can do this by not pumping the dialog but showing it the normal way using showDialog (and then checking the awaited result).

provokateurin:
Tests LGTM, I'll give it another round of tests to check that everything works correctly.

provokateurin:
From my testing:
- Moving a file or folder has the file or folder set as the initial path. For folders this makes no sense and for files even less. The parent dir should be the initial path.
- Instead of greying out the choose button when trying to move a directory into itself the whole entry should just be hidden in the listing. This completely avoids the problem.
- Copying a folder into itself should still be possible (even though it is a weird use-case).

Everything else worked flawlessly :heart:

provokateurin (changes requested):
From my testing:
- Moving a file or folder has the file or folder set as the initial path. For folders this makes no sense and for files even less. The parent dir should be the initial path.
- Instead of greying out the choose button when trying to move a directory into itself the whole entry should just be hidden in the listing. This completely avoids the problem.
- Copying a folder into itself should still be possible (even though it is a weird use-case).

Everything else worked flawlessly :heart:

Leptopoda:
> Copying a folder into itself should still be possible (even though it is a weird use-case).

Why? This sounds like it should just break

provokateurin:
It is not a problem because the original folder still exists. You are only creating new files, not moving which is deleting and re-adding them which requires the folder to still exist while it doesn't

Leptopoda:
Trying to copy a folder into itself returns in a 409 and an error message being shown
[13] [Fri Dec 29 18:23:03 2023] 172.17.0.1:49316 [409]: COPY /remote.php/webdav/Notizen
[13] [Fri Dec 29 18:23:03 2023] 172.17.0.1:49316 Closing

provokateurin:
Ok, then not special handling :+1:

provokateurin (approved):
LGTM, I think some of the logic for showing modals and bottom sheets could be moved from files into the framework so it can be reused, but that can be done later when needed.

PR #1000, opened 2023-10-19
perf(tool): avoid rebuilding the specs if nothing changed

Leptopoda:
I quickly tested it with subsequent runs so only the `dart fix` part is really invoked but changing something in the dynamite package still correctly rebuilds the specs.

Please also test this on your side. As said I only quickly tested it for now


provokateurin (approved):
Nice, this makes it so much faster if nothing in dynamite was changed.

PR #999, opened 2023-10-19
Fix/dynamite/numbers

Leptopoda:
Needed for some future dynamite `someOfs` cleanups.
Depends on #997 

Currently the provisioning_api spec looks broken so not yet mergable.

PR #1001, opened 2023-10-19
make oneOf and anyOf throw at build time rather then on deserialization

Leptopoda:
Also correctly requires oneOf to only have one matching member.
Depends on #999 thus also blocked on the spec being corrected

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
@@ -101,29 +101,63 @@ TypeResult resolveOfs(
       fields[result.name] = toFieldName(dartName, result.name);
     }
 
-    state.output.addAll([
-      buildInterface(
-        identifier,
-        methods: BuiltList.from(
-          results.map(
-            (final result) => Method(
-              (final b) {
-                final s = schema.ofs![results.indexOf(result)];
-                b
-                  ..name = fields[result.name]
-                  ..returns = refer(result.nullableName)
-                  ..type = MethodType.getter
-                  ..docs.addAll(s.formattedDescription);
-              },
-            ),
+    final interface = buildInterface(
+      identifier,
+      methods: BuiltList.from(
+        results.map(
+          (final result) => Method(
+            (final b) {
+              final s = schema.ofs![results.indexOf(result)];
+              b
+                ..name = fields[result.name]
+                ..returns = refer(result.nullableName)
+                ..type = MethodType.getter
+                ..docs.addAll(s.formattedDescription);
+            },
           ),
         ),
       ),
-      buildBuiltClass(
-        identifier,
-        customSerializer: true,
-        methods: BuiltList.build(
-          (final b) => b.add(
+    );
+
+    final hook = Method((final b) {
+      b
+        ..name = '_validate'
+        ..returns = refer('void')
+        ..annotations.add(
+          refer('BuiltValueHook').call([], {'finalizeBuilder': literalTrue}),
+        )
+        ..static = true
+        ..requiredParameters.add(
+          Parameter(
+            (final b) => b
+              ..name = 'b'
+              ..type = refer('${identifier}Builder'),
+          ),
+        );
+
+      final buffer = StringBuffer()
+        ..writeln('// When this is rebuild from another builder')
+        ..writeln('if (b._data == null) { return;}')
+        ..writeln()
+        ..writeln('final match = ')
+        ..writeln('[${fields.values.map((final e) => 'b._$e').join(',')}]')
+        ..writeln(schema.oneOf != null ? '.singleWhereOrNull' : '.firstWhereOrNull')
+        ..writeln('((final x) => x != null);')
+        ..writeln('if (match == null) {')
+        ..writeln(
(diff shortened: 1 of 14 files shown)


packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
```diff
@@ -101,29 +101,63 @@ TypeResult resolveOfs(
...
+
+      final buffer = StringBuffer()
+        ..writeln('// When this is rebuild from another builder')
+        ..writeln('if (b._data == null) { return;}')
+        ..writeln()
+        ..writeln('final match = ')
+        ..writeln('[${fields.values.map((final e) => 'b._$e').join(',')}]')
+        ..writeln(schema.oneOf != null ? '.singleWhereOrNull' : '.firstWhereOrNull')
+        ..writeln('((final x) => x != null);')
+        ..writeln('if ( match == null) {')
```
provokateurin, line 146:
```suggestion
        ..writeln('if (match == null) {')
```

packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
```diff
@@ -101,29 +101,63 @@ TypeResult resolveOfs(
...
+        ..writeln(
+          'throw StateError("Need ${schema.oneOf != null ? 'exactly' : 'at least'} one of ${fields.values.map((final e) => "'$e'").join(',')} for \${b._data}");',
```
provokateurin, line 148:
```suggestion
          'throw StateError("Need ${schema.oneOf != null ? 'exactly' : 'at least'} one of ${fields.values.map((final e) => "'$e'").join(', ')} for \${b._data}");',
```

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.json
provokateurin, line 1: Was it necessary to make everything required or what's the reason you did this change?

packages/dynamite/dynamite_end_to_end_test/test/nested_ofs_test.dart
```diff
@@ -37,48 +38,31 @@ void main() {
     final object = BaseNestedOneOf(
       (final b) => b
         ..data = JsonObject(['attribute-oneOf', 'baseOneOfAttributeOneOfValue'])
-        ..baseAllOf.update((final b) {})
         ..baseOneOf.update(
           (final b) => b
             ..data = JsonObject(['attribute-oneOf', 'baseOneOfAttributeOneOfValue'])
-            ..base.update((final b) {})
+            //..base.update((final b) {})
```
provokateurin, line 44: Commented code

provokateurin (changes requested):

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.json
Leptopoda: At least for the oneOf case a `null` value would match all of the fields so it was just a wrong schema from the beginning.

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.json
```diff
@@ -8,7 +8,13 @@
         "schemas": {
             "Base": {
                 "type": "object",
+                "required": [
+                    "attribute"
+                ],
                 "properties": {
+                    "required": [
+                        "attribute"
+                    ],
```
provokateurin, line 17: I think this is here by accident?

provokateurin (changes requested):
I think you didn't correctly run dart fix in the end. Other than that and the small nitpick LGTM

Leptopoda:
These weren't accidental but just the builder failing due to #1024 and I just didn't notice it.

provokateurin:
I think that particular one is by accident since you created a new property instead of setting the property to be required.

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.json
provokateurin: I'm quite sure this one is not intended. It is a new property in the schema and not part of the schema itself. Please take a look at the full diff to see what I mean.

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.dart
```diff
@@ -33,7 +34,8 @@ class Client extends DynamiteClient {
 
 @BuiltValue(instantiable: false)
 abstract interface class BaseInterface {
-  String? get attribute;
+  JsonObject? get required;
```
provokateurin, line 37: Here in the code it adds the new property and I don't see a reason why this would be needed for the test

packages/dynamite/dynamite_end_to_end_test/lib/nested_ofs.openapi.json
Leptopoda: I understand but I thought I already fixed it. :thinking:

provokateurin (approved):
Rebase, then it will pass

PR #997, opened 2023-10-19
Refactor/dynamite

Leptopoda:

PR #1005, opened 2023-10-22
refactor(dynamite,nextcloud): add default parameter documentation

Leptopoda:
continuation of #997


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -430,6 +397,46 @@ return rawResponse.future;
   }
 }
 
+Iterable<String> buildSerialization(
```
provokateurin, line 400:
```suggestion
Iterable<String> buildParameterSerialization(
```

provokateurin (approved):
Only on small name change

Issue #849, opened 2023-09-25
Refactor settings


Leptopoda:
The settings code both has a bad api and even worse performance.
When rewriting we should also take into account a possible iOS design.

To clean up the code we could add a boolean field to `Option` controlling whether it is supported. The settings list could have a generator function that would dismiss unsupported options and empty option categories (related to #696 ).
An unsupported option could also be optimized by adding early returns to the setters potentially improving performance.

The entire settings layout with Rows, and intrinsic sizes should be refactored for better performance.

PR #1012, opened 2023-10-23
Refactor/neon/settings

Leptopoda:
handles the design part of #849
Changes to existing platforms:
- the visual density of the Settings will be higher on Linux than before
- `SelecOption`s will now be handled in popup dialogs as it is common on m3
- SettingsCategories will no longer be separated by a divider to match the m3 design 

The m3 docs themselves do not mention the modifications I made but I got inspired by the system settings on stock android


Leptopoda:
I think that the entire idea of a separate `SettingsListTile` is bad.
Can we just remove it and provide it and provide normal widgets and helpers to style custom ones?

Leptopoda (commented):
https://github.com/nextcloud/neon/pull/1030#discussion_r1372870104

So we don't forget those refactors

.cspell/dart_flutter.txt
```diff
@@ -1,4 +1,5 @@
 autofocus
+Cupertino
```
provokateurin, line 2: All words in the dictionaries are lowercase

packages/neon/neon/lib/src/settings/widgets/settings_category.dart
```diff
@@ -1,41 +1,78 @@
...
-  final List<SettingsTile> tiles;
+  final List<Widget> tiles;
+  final Widget? footer;
+  final bool hasLeading;
 
   @override
   Widget build(final BuildContext context) {
+    final theme = Theme.of(context);
+
+    switch (theme.platform) {
```
provokateurin, line 24: Should use isCupertino()

provokateurin (changes requested):
Select options are a bit broken because the currently selected value is not shown in the tile. Initially I thought you forgot to add it (I checked how it works in the settings on my phone) but it looks like you added it and it's just not working.

Is there an easy way to test the cupertino layout?

provokateurin (changes requested):
packages/app/pubspec.lock is not updated

Leptopoda:
> Is there an easy way to test the cupertino layout?

Just specify the platform in the theme. There are still bigger issues with the scaffold background color mismatching but as iOS is still a WIP I will address them separately.
```dart
    return ThemeData(
        platform: TargetPlatform.iOS,
        [...]
      );
```

Leptopoda:
Also because of some license issues with the apple font you will not be able to see the `+` icon in the add account tile unless you run it on iOS itself.

Leptopoda:
Also because of some license issues with the apple font you will not be able to see the `+` icon in the add account tile unless you run it on iOS itself.

provokateurin:
> Also because of some license issues with the apple font you will not be able to see the + icon in the add account tile unless you run it on iOS itself.

lol what. fucking apple...

provokateurin (approved):

Issue #555, opened 2023-08-13
Add CI workflow to check that all generated things are up-to-date


provokateurin:

provokateurin:
Similarly to https://github.com/nextcloud/neon/issues/494 this should only be done for changed packages.
It would be possible to also check if the docs, assets or specs changed, but that is more complex than letting melos figure that out, so I would propose to just check those every time because they should be quite fast anyway.

provokateurin:
@Leptopoda Besides the code generation which is implemented now we should also add CI to check that our specs are up-to-date.

PR #1014, opened 2023-10-23
feat(nextcloud): test validity of the generated clients

Leptopoda:
implements:  #555

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -46,4 +46,5 @@ jobs:
           if [ -z "$packages" ]; then exit 0; fi
           packages_glob="$(printf ",%s" "${packages[@]}")"
           packages_glob="${packages_glob:1}"
-          melos exec --scope="$packages_glob" --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
+          melos exec --scope="$packages_glob" --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all) -x source_verification"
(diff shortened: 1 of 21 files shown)


Leptopoda:
depends on #1024
Looks like the CI liked my changes.

Leptopoda:
This is based on the work we did in #1035

provokateurin (commented):
This looks mostly good, but I still don't like that it gets so complex.
I have an idea how we can do this better:

Move the ensure_validity_test.dart files into a source_verification_test/ dir inside the packages.
Then adjust the melos commands to run with concurrency 1 for packages with that dir and point the test command only to that dir.
Of course this makes it a bit more work to run those tests locally for a single package because you now need to run `fvm dart test source_verification_test/`, but I would argue that you rarely want to run those tests locally if not never.
They are really just for the CI to ensure everything committed is up-to-date.

Does this make sense? Maybe I'm still missing something, but in my head this is much better than the tricks with the tags and checking for the specific test file name.

provokateurin:
The other option that gets completely rid of these problems and the build_verify way is to just setup a workflow that runs the generate scripts and checks for diffs with git diff. That would be a more classic approach which might be better because we just avoid this whole mess.

Leptopoda:
> The other option that gets completely rid of these problems and the build_verify way is to just setup a workflow that runs the generate scripts and checks for diffs with git diff. That would be a more classic approach which might be better because we just avoid this whole mess.

Feel free to implement this. I'm not a script wizard as you are :)
I think this approach is completely fine and once we support more platforms we might want to expand the tags even further (thinking about web tests).

Maybe we can just skip the tag entirely in packages that don't use it but I'm not sure this would work.

provokateurin:
Ok I'll do that and open a new PR

PR #1015, opened 2023-10-23
refactor(dynamite,nextcloud): change path parameter construction.

Leptopoda:
in continuation of #997

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/client.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/client.dart
@@ -219,7 +219,6 @@ Iterable<Method> buildTags(
           .join(',');
 
       code.writeln('''
-  var _path = '${pathEntry.key}';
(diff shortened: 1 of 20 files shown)


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -275,8 +274,8 @@ Iterable<Method> buildTags(
           ),
         );
 
-        buildPatternCheck(result, parameter).forEach(code.writeln);
         buildParameterSerialization(result, parameter).forEach(code.writeln);
+        buildPatternCheck(result, parameter).forEach(code.writeln);
```
provokateurin, line 278: Why is the order switched? I think we should first check if the parameter values match the patterns and then start to use them to construct the path.

provokateurin (approved):

PR #1016, opened 2023-10-23
[Feature][#921] dependency_validator added

### LINKED ISSUE #921: Setup CI for checking unused dependencies
https://pub.dev/packages/dependency_validator

NishchayShakya1:
# Description
A dependency validator is added to automatically check unused dependencies.

closes #921 
`hacktoberfest`

### DIFF
--- a/packages/app/pubspec.lock
+++ b/packages/app/pubspec.lock
@@ -194,7 +194,7 @@ packages:
     source: hosted
     version: "0.7.8"
   dependency_validator:
-    dependency: transitive
+    dependency: "direct dev"
     description:
       name: dependency_validator
       sha256: f727a5627aa405965fab4aef4f468e50a9b632ba0737fd2f98c932fec6d712b9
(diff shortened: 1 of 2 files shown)


provokateurin:
@Leptopoda should we add it to our global workspace package or rather to each package individually? I think both would work for us

NishchayShakya1:
> @Leptopoda should we add it to our global workspace package or rather to each package individually? I think both would work for us

Yes, both will work

provokateurin:
It was more of a architecture question than a technical one. Of course we can add it to each package, but the question is if we want to do it that way.

NishchayShakya1:
> It was more of a architecture question than a technical one. Of course we can add it to each package, but the question is if we want to do it that way.

I think adding to each package will be better.

provokateurin:
@Leptopoda Agreed with your points, I thought the same.

provokateurin:
@NishchayShakya1 Would it be possible for you to incorporate the feedback? We can help you if you get stuck or don't know what to do.

Leptopoda:
@NishchayShakya1 Are you still planning to implement this feature?
If you need any help feel free to ask.

Otherwise I'm going to close this one at the end of next week.

PR #1020, opened 2023-10-24
feat(dynamite): Support content in parameters

provokateurin:
Needed for the upcoming change with https://github.com/nextcloud/openapi-extractor/pull/22.
Content parameters are basically just content strings, but already available in OpenAPI 3.0.
I'm not sure if the way I implemented it is a good idea, but I felt it would be annoying to always check for both schema and content and then do the conversion manually all the time.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
+++ b/packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
@@ -4,6 +4,7 @@ import 'package:built_value/serializer.dart';
 import 'package:dynamite/src/helpers/dart_helpers.dart';
 import 'package:dynamite/src/helpers/docs.dart';
 import 'package:dynamite/src/models/exceptions.dart';
+import 'package:dynamite/src/models/openapi/media_type.dart';
 import 'package:dynamite/src/models/openapi/schema.dart';
 
 part 'parameter.g.dart';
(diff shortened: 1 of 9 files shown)


packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -34,7 +61,7 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
       throw OpenAPISpecError('Path parameters must be required but ${b.name} is not.');
     }
```
Leptopoda, line 63: Please throw an `OpenAPISpecError` if both content and schema are specified.
According to the docs they are mutually exclusive.

packages/dynamite/dynamite_end_to_end_test/pubspec.yaml
```diff
@@ -4,14 +4,18 @@ description: Tests for dynamite. Not meant for publishing.
 version: 1.0.0
 
 environment:
-  sdk: '>=3.1.1 <4.0.0'
+  sdk: '>=3.1.0 <4.0.0'
 
 dependencies:
-  built_value: ^8.6.2
+  built_collection: ^5.0.0
+  built_value: ^8.0.0
```
Leptopoda, line 11: Not related to this PR but we should probably pin the dependencies of non publishable packages to the latest version or use `any` with a up to date lockfile.

This way we could test the code dynamite generates against the latest version.
I noticed something similar in the `neon_lints` package as I don't think renovate would handle the lint_maker update (#1021).

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
provokateurin: Wanted to add it - forgot to do so in the end :)

packages/dynamite/dynamite_end_to_end_test/pubspec.yaml
provokateurin: I thought about this as well. I think there is no way to build rules for renovate to do this automatically, but we could make a separate rule to enable non-major bumps for some packages specifically.

provokateurin: But for your lint_maker changes it is fine since dev_dependencies are always bumped to the latest version.

Leptopoda (approved):

PR #1021, opened 2023-10-24
feat(neon_lints): use published lint_maker package

Leptopoda:
I needed to manually remove `avoid_unstable_final_fields` for now as it is not available in dart `3.1.4` yet but marked as available since `3.1`.

### DIFF
--- a/packages/neon_lints/lint_maker.yaml
+++ b/packages/neon_lints/lint_maker.yaml
@@ -1,20 +1,5 @@
-dart:
-  commitHash: 0a567d0
+base:
   output: 'lib/src/base.yaml'
-  disabledRules:
-    - always_specify_types
-    - avoid_annotating_with_dynamic
-    - avoid_as
-    - avoid_catches_without_on_clauses
-    - avoid_final_parameters
-    - avoid_print
-    - diagnostic_describe_all_properties
-    - lines_longer_than_80_chars
-    - no_default_cases
-    - one_member_abstracts
-    - prefer_double_quotes
-    - prefer_relative_imports
-    - unnecessary_final
   preset:
     analyzer:
       language:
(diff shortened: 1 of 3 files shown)


packages/neon_lints/lint_maker.yaml
```diff
@@ -26,4 +11,18 @@ dart:
...
+        - avoid_catches_without_on_clauses
+        - avoid_final_parameters
+        - avoid_print
+        - diagnostic_describe_all_properties
+        - lines_longer_than_80_chars
+        - no_default_cases
+        - one_member_abstracts
+        - prefer_double_quotes
+        - prefer_relative_imports
+        - unnecessary_final
```
provokateurin, line 28: Shouldn't those be set to `: false`?

Leptopoda: It can be debated wether this makes sense but all rules listed in the lint_maker.yaml file are considered to not be wanted thus disabled.
I wrote it that way because that's the only thing that makes sense in the context of enabling all available rules by default.

The syntax with `rule_name: false` is also possible. Should I change it over here?

provokateurin: I mean it is totally unintuitive because the key is called rules and not disabled_rules or something like that.

Leptopoda: I initially wanted to keep the syntax 1:1 compatible to the analysis_options file to leverage a parser from the analyzer package which I didn't end up doing.

Do you think renaming the map to `disabled_rules` as you mentioned is better or should I just use the analysis_options syntax and interpret list elements as enabled?

Leptopoda:
I reported the issue with the wrongly released rule upstream.
https://github.com/dart-lang/sdk/issues/53852

provokateurin (approved):

PR #1023, opened 2023-10-25
supply a full version for the minimum supported openapi version

Leptopoda:


provokateurin (approved):
I think we should also limit the upper version. That would be 3.1 right now

Leptopoda:
Why not limit it to the next major one (4.0.0 in this case)

provokateurin:
Because then we are expected to also support 3.2 and other versions, but since don't claim full compatibility anyway it doesn't matter. Please pin to <4.0.0

packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -45,7 +48,8 @@ class OpenAPIBuilder implements Builder {
         _ => throw StateError('Openapi specs can only be yaml or json.'),
       };
 
-      if (Version.parse(spec.version).major != 3) {
+      final version = Version.parse(spec.version);
+      if (version < minSupportedVersion || version > minSupportedVersion.incrementMajor()) {
         throw Exception('Only OpenAPI 3.0.0 and later are supported');
```
provokateurin, line 53: Error message should also include the upper version

provokateurin (changes requested):

provokateurin (approved):

PR #1024, opened 2023-10-25
Fix/dynamite/header type resolver

Leptopoda:
This was a funny bug :)

1. We use generator functions where possible in dynamite for performance reasons. The problem is that they only get invoked lazily (when iterating). This means that they where only executed in the `addAll` code segment. Because `generateImports` was called before adding the code generated by `generateClients` we where not taking the resolved types from these methods into account.

2. an empty operationId caused our code to fail (fb01f7eaf1f714235774bf59d25c013857af4164)

3. We didn't hit it but we where calling `output.add` for the TypeDefs before we added the imports to the output meaning that we could potentially have code before the imports.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
@@ -3,7 +3,6 @@ import 'package:dynamite/src/builder/resolve_type.dart';
 import 'package:dynamite/src/builder/state.dart';
 import 'package:dynamite/src/helpers/dart_helpers.dart';
 import 'package:dynamite/src/models/openapi.dart' as openapi;
-import 'package:dynamite/src/models/type_result.dart';
 
 Iterable<Spec> generateSchemas(
   final openapi.OpenAPI spec,
@@ -12,27 +11,13 @@ Iterable<Spec> generateSchemas(
   if (spec.components?.schemas != null) {
     for (final schema in spec.components!.schemas!.entries) {
       final identifier = toDartName(schema.key, uppercaseFirstCharacter: true);
-      if (schema.value.type == null && schema.value.ref == null && schema.value.ofs == null) {
-        yield TypeDef(
-          (final b) => b
-            ..name = identifier
-            ..definition = refer('dynamic'),
-        );
-      } else {
-        final result = resolveType(
-          spec,
-          state,
-          identifier,
-          schema.value,
-        );
-        if (result is TypeResultBase) {
-          yield TypeDef(
-            (final b) => b
-              ..name = identifier
-              ..definition = refer(result.name),
-          );
-        }
-      }
+
+      resolveType(
+        spec,
+        state,
+        identifier,
+        schema.value,
+      );
     }
   }
 
--- a/packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
@@ -1,0 +1,40 @@
+import 'package:code_builder/code_builder.dart';
+import 'package:dynamite/src/builder/resolve_type.dart';
+import 'package:dynamite/src/builder/state.dart';
+import 'package:dynamite/src/helpers/dart_helpers.dart';
+import 'package:dynamite/src/models/openapi.dart' as openapi;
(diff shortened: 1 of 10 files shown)


packages/dynamite/dynamite/lib/src/builder/imports.dart
```diff
@@ -24,7 +25,11 @@ List<Spec> generateImports(final AssetId outputId, final State state) => [
       Directive.import('package:meta/meta.dart'),
       Directive.import('package:universal_io/io.dart'),
       const Code(''),
-      if (state.resolvedTypes.isNotEmpty) ...[
+      if (state.resolvedTypes
+          .where(
+            (final type) => type is TypeResultEnum || (type is TypeResultObject && type.className != 'ContentString'),
```
provokateurin, line 30: Please move this into a separate method for readability (and maybe reusability as well).

packages/dynamite/dynamite_end_to_end_test/lib/headers.openapi.json
provokateurin, line 1: Please add another test with header and content as well

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/builder/imports.dart
Leptopoda: What about adding it as a method to the state class?

provokateurin: Sounds good

Leptopoda:
I hope that we can sort the generated classes in the future to avoid such diffs.
I couldn't really figure out why this happens.

provokateurin (approved):

provokateurin:
Maybe we should then first figure out how to prevent this diff?

Leptopoda:
@provokateurin can I get a froce merge?

The generation check for dynamite_e2e timed out and I already fix this in #1014

packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
```diff
@@ -0,0 +1,40 @@
...
+
+Iterable<Spec> generateSchemas(
+  final openapi.OpenAPI spec,
+  final State state,
+) sync* {
+  if (spec.components?.schemas != null) {
+    for (final schema in spec.components!.schemas!.entries) {
+      final identifier = toDartName(schema.key, uppercaseFirstCharacter: true);
+      if (schema.value.type == null && schema.value.ref == null && schema.value.ofs == null) {
+        yield TypeDef(
```
Leptopoda, line 16: @provokateurin can you please provide a spec that hits those two `TypeDef`s?
I don't entirely understand what they try to accomplish and they aren't covered by any of the nc specs.

As mentioned above these are also broken without this PR.

provokateurin: I don't know where these originate from and I think I never had them used. They also don't make any sense imo, so just remove them

packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
```diff
@@ -0,0 +1,40 @@
...
+            (final b) => b
+              ..name = identifier
+              ..definition = refer(result.name),
+          );
+        }
+      }
+    }
+  }
+
+  yield* state.output;
```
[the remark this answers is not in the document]
provokateurin: Let's merge this and figure this out later. Worst case we just make another PR that has a big diff, but then it's clean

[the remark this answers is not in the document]
provokateurin: Can you explain this a little bit? I'm not sure I understand why this is needed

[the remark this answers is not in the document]
Leptopoda: We still need to add the `state.output` to the listbuilder in the main `openapi_builder.dart`.
We could have done it there but then we wouldn't yield anything from this method thus we'd need to both call this method and separately add the out add the generated code to the output. I think this is error prone.

Sadly registering everything to the state directly resulted in some error.
Nevertheless the dynamite code is coming into shape and the separation we have is very nice.

PR #1036, opened 2023-10-26
fix(neon_dashboard): Fix localizations class name

provokateurin:
This slipped through the cracks because the dashboard functionality was built while the localization class names were changed.


provokateurin:
Honestly the docs for adding a new package are super incomplete and quite useless for actually adding a new package. Also it's generic to all packages and not the client implementations.

Leptopoda (approved):

Issue #808, opened 2023-09-18
Meta: Plugin system


Leptopoda:
Clients not strictly in the drawer should be moved into a new type similar to AppImplementations called Plugin.
A plugin differs from a client as they don't require to be tied to an app and are usually more entangled into the neon framework itself. Clients would have their Bloc interfaces inside the framework.
An app dev would be able to both specify a list of Clients and plugins improving the ability to configure them and also more tightly handle native dependencies or permissions.

Potential new plugins:
- [ ] User provisioning
- [ ] Unified search (with configuration options to disable global search)
- [ ] Add a sync plugin
- [ ] push_notifications (with the native dependencies)
- [ ] #303 
- [ ] Hide the bloc of the clients from other apps (making them protected)
- [ ] Potentially interfaces to be called by other  clients themselves (files plugin that can be used without bundling the entire files app
- [ ] login plugins supporting custom authentication methods like pre provisioning or the ability to leave QR login out to avoid relying on the camera permission
- [ ] Rename AppImplementation to Client to avoid confusion

provokateurin:
I think we need to put in some more thought on how different clients/plugins can know about and interact with each other. Maybe we can starting by drafting something in PlantUML again which we can later add to the docs as well? Just to have a nice overview of the architecture and a plan for the implementation.

provokateurin:
In my head every server app will be implemented by either a client or a plugin. For the core we will split it even more to separate features like unified search, login methods etc. I think it is quite important to have good guidelines on how each feature should be implemented and where we draw the lines between clients and plugins.

provokateurin:
I think we should introduce some more abstractions for the plugins so we can try to keep most stuff generic in a way and don't customize everything in the framework.
For example there can be a class called `FilesActionsPlugin` and all the plugins that extend this class can modify the files browser in a way, like adding a sharing menu for example.
The same thing should work for other plugins that are in the app bar (like unified search or notifications).

This needs some more thoughts though and a proper UML diagram should make this much clearer.

Leptopoda:
For now we do not see any benefit of such a major refactor.
Once we have a need for it we'll revisit the idea.

Leptopoda:
@provokateurin should we close this as we already discussed our general direction with a repository layer?

provokateurin:
I think we should still keep it open, this is not only about those issues as we also need the new app registration stuff and so on.

PR #1041, opened 2023-10-27
fix(neon): Fix SVG code compilation for web

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/372
This wasn't compiling before because it was relying on dart:io code internally.

### DIFF
--- a/packages/neon/neon/lib/src/utils/push_utils.dart
+++ b/packages/neon/neon/lib/src/utils/push_utils.dart
@@ -7,7 +7,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
-import 'package:flutter_svg/flutter_svg.dart' show SvgFileLoader, vg;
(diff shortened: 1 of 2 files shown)


Leptopoda (changes requested):
From the SvgLoader documentation:
> /// A [BytesLoader] that parses a SVG data in an isolate and creates a
/// vector_graphics binary representation.

This means loading the file could happen in a separate isolate thus not blocking the main one.
I'd rather make a custom `UniversalSvgFileLoader` as each loader is only around 30 lines of code.

provokateurin:
The annoying thing is that this code isn't even executed on web since we have no push notification support there. Maybe we should instead use a stub on web to fix the compilation (conditional import if you know what I mean).

Leptopoda:
I don't know what makes most sense rn but this shouldn't even be a problem with #808.
Do you think it's worth stubbing?

provokateurin:
I think I will go with the stub for now and once we have the plugin system we can easily drop it.

provokateurin:
I went with your suggestion because creating the loader was very easy indeed. Just had to copy the SvgFileLoader and replace dart:io with universal_io.

Leptopoda (approved):

PR #1044, opened 2023-10-28
Feat/named records

Leptopoda:
How didn't we use named records earlier?

Our dialogs are still untouched but I'd rather add it to #998

### DIFF
--- a/packages/neon/neon/lib/src/models/app_implementation.dart
+++ b/packages/neon/neon/lib/src/models/app_implementation.dart
@@ -15,6 +15,7 @@ import 'package:neon/src/settings/models/storage.dart';
 import 'package:neon/src/utils/provider.dart';
 import 'package:neon/src/widgets/drawer_destination.dart';
 import 'package:nextcloud/core.dart' as core;
+import 'package:nextcloud/nextcloud.dart' show VersionSupported;
 import 'package:provider/provider.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:vector_graphics/vector_graphics.dart';
@@ -36,15 +37,14 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions> 
 
   /// Checks if the app is supported on the server of the [account].
   ///
-  /// A `supported` value of `null` means that it can not be known if the app is supported.
+  /// A value of `null` means that it can not be known if the app is supported.
   /// This is the case for apps that depend on the server version like files and we assume that the app is supported.
   /// The server support is handled differently.
-  ///
-  /// The first value of the record is the supported status and the second value is the supported minimum version.
-  FutureOr<(bool? supported, String? minimumVersion)> isSupported(
+  FutureOr<VersionSupported<String>?> isSupported(
     final Account account,
     final core.OcsGetCapabilitiesResponseApplicationJson_Ocs_Data capabilities,
-  );
+  ) =>
+      null;
 
   final blocsCache = AccountCache<T>();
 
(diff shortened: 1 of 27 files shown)


packages/nextcloud/lib/src/helpers/common.dart
```diff
@@ -0,0 +1,2 @@
+/// The result of a version check.
+typedef VersionSupported<T> = ({bool supported, T minimumVersion});
```
provokateurin, line 2:
```suggestion
typedef VersionSupported<T> = ({bool isSupported, T minimumVersion});
```

packages/neon/neon_notifications/lib/neon_notifications.dart
```diff
@@ -56,9 +56,9 @@ class NotificationsApp extends AppImplementation<NotificationsBloc, Notification
   BehaviorSubject<int> getUnreadCounter(final NotificationsBloc bloc) => bloc.unreadCounter;
 
   @override
-  (bool? supported, String? minimumVersion) isSupported(
+  VersionSupported<String?> isSupported(
     final Account account,
     final core.OcsGetCapabilitiesResponseApplicationJson_Ocs_Data capabilities,
   ) =>
-      const (null, null);
+      const (supported: true, minimumVersion: null);
```
provokateurin, line 63: I feel like this should still be null. We don't know if the app is actually supported because that depends on the core. Thus saying it is, is not actually true. Null means we don't know and I think we should keep it that way

packages/sort_box/lib/sort_box.dart
```diff
@@ -17,7 +17,7 @@ class SortBox<T extends Enum, R> {
   /// A mapping of values [T] to their *Boxes*.
   ///
   /// The Boxes are applied if two elements are considered equal regarding their property [T].
-  final Map<T, Set<(T property, SortBoxOrder order)>> _boxes;
+  final Map<T, Set<({T property, SortBoxOrder order})>> _boxes;
```
provokateurin, line 20: Please make the record type a separate typedef as it is used multiple times

provokateurin (changes requested):

packages/neon/neon_notifications/lib/neon_notifications.dart
Leptopoda: Then I'd rather make the entire check return null. When a check is provided it shouldn't be null able.

provokateurin: Yep that's better. Without named records this was annoying with deconstructing the records if the whole record was null, but that isn't a problem anymore

Leptopoda: Ah I remember :+1:

packages/neon/neon/lib/src/models/app_implementation.dart
```diff
@@ -41,10 +42,11 @@ abstract class AppImplementation<T extends Bloc, R extends NextcloudAppOptions>
   /// The server support is handled differently.
   ///
   /// The first value of the record is the supported status and the second value is the supported minimum version.
-  FutureOr<(bool? supported, String? minimumVersion)> isSupported(
+  FutureOr<VersionSupported<String?>?> isSupported(
```
provokateurin, line 45:
```suggestion
  FutureOr<VersionSupported<String>?> isSupported(
```

provokateurin (changes requested):

Leptopoda:
I forgot to adjust the doc comment sorry.

provokateurin (approved):

PR #1048, opened 2023-10-29
feat(ci): Validate code generation

provokateurin:
Closes https://github.com/nextcloud/neon/issues/555
Replaces https://github.com/nextcloud/neon/pull/1014


.github/workflows/generate.yml
Leptopoda, line 1: Can we please use `yaml` to keep it consistent with other files.

Leptopoda, line 1: Looks like these tests will now run on every PR instead of only when a package or dependency changes.

Leptopoda (commented):
You can also remove the build_verify dependency in dynamite_e2e

.github/workflows/generate.yml
provokateurin: This is quite hard to solve because the code generation doesn't just depend on changed packages. It also depends on changes in external/ or tool/

provokateurin: But we should only run the tests when (probably) relevant.
I guess we should filter on changes in `external/`, `tool/`, `packages/dynamite`, `packages/nextcloud`.
I wouldn't try to filter beyond that as we can not be sure which changes actually affect us.

Leptopoda: sounds good :+1:

provokateurin:
With skipping instead of filtering we can set the workflow to be required which is good to avoid accidental merges with red CI.

provokateurin:
@Leptopoda Ready for review

Leptopoda (approved):

PR #1025, opened 2023-10-25
feat(dynamite): introduce dynamite config to set ignore directives

Leptopoda:
We should probably first wait for #1024

I hope that we can leverage the config more in the future.  I disabled coverage ignores in the dynamite_e2e  tests as we should get the real data once coverage reports are set up.

PR #1051, opened 2023-10-29
feat(dynamite): Support default responses

provokateurin:
Closes https://github.com/nextcloud/neon/issues/871

I had this laying around from the petstore example. Might as well merge it already now that we have tests for dynamite.

### DIFF
--- a/packages/dynamite/dynamite_end_to_end_test/lib/responses.openapi.dart
+++ b/packages/dynamite/dynamite_end_to_end_test/lib/responses.openapi.dart
@@ -1,0 +1,245 @@
+// ignore_for_file: camel_case_types
+// ignore_for_file: discarded_futures
+// ignore_for_file: public_member_api_docs
+// ignore_for_file: unreachable_switch_case
+import 'dart:typed_data';
+
+import 'package:built_value/serializer.dart';
+import 'package:built_value/standard_json_plugin.dart';
+import 'package:dynamite_runtime/built_value.dart';
+import 'package:dynamite_runtime/http_client.dart';
+import 'package:meta/meta.dart';
+import 'package:universal_io/io.dart';
+
+class Client extends DynamiteClient {
+  Client(
+    super.baseURL, {
+    super.baseHeaders,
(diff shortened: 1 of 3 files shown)


Leptopoda (commented):
How does this handle having both a status code and a default response specified?

I think we really have to make the status code validation taka a function that uses a switch case. This would also allow us to handle status code ranges.

provokateurin:
It is handled the same way as we currently do for multiple status codes: As long as they have the same schema they will be handled the same. If we have a default response then we can ignore if we have other status codes with the same schema because they will be handled by the default already.
I don't think we need your function for this yet (although we certainly need it in the future for supporting multiple schemas). I would like to keep everything as simple as possible until we actually need something more complex.

Leptopoda:
The please also add a test that handles this (having both a default and a status code)

packages/dynamite/dynamite_end_to_end_test/lib/responses.openapi.dart
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  /// This method and the response it returns is experimental. The API might change without a major version bump.
+  ///
+  /// Returns a [Future] containing a [DynamiteRawResponse] with the raw [HttpClientResponse] and serialization helpers.
+  /// Throws a [DynamiteApiException] if the API call does not return an expected status code.
+  ///
+  /// Status codes:
+  ///   * 200
+  ///   * 201
```
Leptopoda, line 103: This is really bad as the doc comments don't reflect the generated code :grimacing:

Leptopoda (approved):
LGTM

Leptopoda:
You'll need to rebase as I just merged #1025.

packages/dynamite/dynamite_end_to_end_test/lib/responses.openapi.dart
provokateurin: Hm true, but it is also good to have this information to handle the other status codes that are not handled by us.

Issue #1049, opened 2023-10-29
Use records for oneOf or anyOf.


Leptopoda:
This would let us reduce the number of generated classes and improve the general api around them.

Leptopoda:
We should think about using sealed stub classes so `oneOf`s with only objects can use an exhaustive switch to match all available types.
A convenience getter could be inttroduces that returns the value that's not null.

The getter does not make sense with `anyOf`s as it's not clear which of the potentially many values should be used.

Leptopoda:
@provokateurin can I get some initial feedback. I think I now got the dart side figured out.
```dart
typedef OneOfIntDoubleOther = ({int? $int, double? $double, String? string});

extension OneOfIntDoubleOtherExtension on OneOfIntDoubleOther {
  void validate() {
    final match = [$int, $double, string].singleWhereOrNull((final x) => x != null);
    if (match == null) {
      throw StateError("Need exactly one of 'int', 'double', 'string'");
    }
  }
}

class _$OneOfIntDoubleOtherSerializer implements PrimitiveSerializer<OneOfIntDoubleOther> {
  @override
  final Iterable<Type> types = const [OneOfIntDoubleOther];

  @override
  final String wireName = 'OneOfIntDoubleOther';

  @override
  Object serialize(
    final Serializers serializers,
    final OneOfIntDoubleOther object, {
    final FullType specifiedType = FullType.unspecified,
  }) {
    dynamic value = object.$int;
    if (value != null) {
      return _jsonSerializers.serialize(value, specifiedType: const FullType(int))!;
    }

    value = object.$double;
    if (value != null) {
      return _jsonSerializers.serialize(value, specifiedType: const FullType(double))!;
    }

    value = object.string;
    if (value != null) {
      return _jsonSerializers.serialize(value, specifiedType: const FullType(String))!;
    }

    throw Error();
  }

  @override
  OneOfIntDoubleOther deserialize(
    final Serializers serializers,
    final Object data, {
    final FullType specifiedType = FullType.unspecified,
  }) {
    int? $int;
    double? $double;
    String? string;

    try {
      final value = _jsonSerializers.deserialize(data, specifiedType: const FullType(int))! as int;
      $int = value;
    } catch (_) {}
    try {
      final value = _jsonSerializers.deserialize(data, specifiedType: const FullType(double))! as double;
      $double = value;
    } catch (_) {}
    try {
      final value = _jsonSerializers.deserialize(data, specifiedType: const FullType(String))! as String;
[...]

PR #1057, opened 2023-10-30
circumvent errors with dynamic or string arrays in oneOfs

Leptopoda:
required for #980 
This is a hacks that just uses the first match in a oneOf to avoid failing when multiple values match.

In theory the normal built_value serializer does include the original type before in the serialized type but this is stripped when a `FullType` is specified.
I don't know yet how we can get access to this information with the `StandardJsonPlugin` enabled.
I plan to investigate other methods and take these issues into consideration when working on #1049.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/ofs_builder.dart
@@ -290,6 +290,7 @@ TypeResult resolveOfs(
                       'result.${fields[result.name]!} = value;'
                     else
                       'result.${fields[result.name]!}.replace(value);',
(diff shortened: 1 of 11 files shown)


provokateurin:
This might backfire because the problematic type might not be the last one but before the one that is actually correct. So we would need to treat oneOf as anyOf for now as the problem can still happen with this change.

Leptopoda:
You are right. I didn't think about this.

What about not trying to fix this issue in dynamite and just patch the spreed spec to be `anyOf`?
I think for now this is the cleanest and easiest we can do.

provokateurin:
Hm yeah that sounds like a good idea. The same thing will also happen in server, but there it only affects two places which we don't even use right now.

PR #1061, opened 2023-10-30
Fix/dynamite/revert typedef removal

Leptopoda:
Fixes and more importantly tests top level schemas that aren't a plain object.

This reverts the previous removal of the typedef generation.


provokateurin (changes requested):
Please just make one commit for the dynamite fixes and one for the new test. Using an actual revert commit is ugly IMO (and the second commit mixes dynamite and e2e changes).

provokateurin (approved):

PR #1062, opened 2023-10-30
chore: test for dart doc errors

Leptopoda:
Adds a very strict dartdoc configuration.
It also already adjusts some errors found in the neon package.

The configuration is taken straight from the [flutter repo](https://github.com/flutter/flutter/blob/master/dartdoc_options.yaml) as I couldn't find any complete list of dartdoc options that can easily be parsed.

Having it in the project root allows all packages to use it.

### DIFF
--- a/dartdoc_options.yaml
+++ b/dartdoc_options.yaml
@@ -1,0 +1,35 @@
+include: package:neon_lints/dartdoc_options.yaml
+
+dartdoc:
+  nodoc: ['lib/l10n/*.dart']
+  showUndocumentedCategories: true
+  ignore:
+    ## Errors that are ignored:
+    ## Warnings that are ignored:
+    - reexported-private-api-across-packages
+    ## Default ignores for dartdoc:
+  errors:
+    ## Default errors of dartdoc:
+    - duplicate-file
+    - invalid-parameter
+    - tool-error
+    - unresolved-export
+    ## Warnings that are elevated to errors:
+    - ambiguous-doc-reference
+    - ambiguous-reexport
+    - broken-link
+    - category-order-gives-missing-package-name
+    - deprecated
+    - ignored-canonical-for
+    - missing-from-search-index
+    - no-canonical-found
+    - no-documentable-libraries
+    - no-library-level-docs
+    - orphaned-file
+    - unknown-file
+    - unknown-macro
+    - unresolved-doc-reference
+    ## Ignores that are elevated to errors:
+    # - type-as-html # broken, https://github.com/dart-lang/dartdoc/issues/3545
+    - missing-constant-constructor
+    - missing-code-block-language
--- a/.cspell/tools.txt
+++ b/.cspell/tools.txt
@@ -10,6 +10,7 @@ classpath
 cloc
 commitlint
 dapplication
+dartdoc
 dists
 dockerenv
 endforeach
@@ -43,6 +44,7 @@ libindicator
 libsqlite
 mipmap
 ndebug
+nodoc
 nproc
 openrelay
 openrelayprojectsecret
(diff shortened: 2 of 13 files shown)


provokateurin:
https://github.com/dart-lang/dartdoc/blob/main/lib/src/warnings.dart seems to have many of those options? I wouldn't automate it, but maybe you can check if there are some more to enable.

provokateurin:
Please separate the docs fixes and the new configuration

Leptopoda:
> Please separate the docs fixes and the new configuration into two commits

Sure. I just pushed it as a wip. I'll also need to check other packages for any errors and test compatibility with the two currently disabled rules.

dartdoc_options.yaml
provokateurin, line 1: Is it enough to just put the file in the root?

Leptopoda: > Having it in the project root allows all packages to use it.

Yep

provokateurin: Awesome

Leptopoda:
Yes

provokateurin:
I put the step after the tests because the tests are more likely to fail than the docs. We can also put it in front if you like that better. It's really annoying that dart doc is so slow, otherwise this wouldn't really matter.

Leptopoda:
I initially thought of doing this in a completely separate job so they can run in parallel.

Dart doc took ages on my device and I expect it to be worse in CI

provokateurin:
Yeah that is also a possibility, then we could also move the formatting and linting steps into separate jobs. I didn't do it yet because each of those jobs will setup the whole project and that seems like a lot of wasted energy.
How about leaving it as is for now and if it gets too annoying we move it into a separate job?

Leptopoda:
> I didn't do it yet because each of those jobs will setup the whole project and that seems like a lot of wasted energy.

I agree and don't think linting needs to be done separately from the tests. But I think that the dartdoc stuff could also just run on a plain dart image without any special setup.

>  How about leaving it as is for now and if it gets too annoying we move it into a separate job?
It already takes very long but lets see how it behaves in the real world.

Leptopoda:
We most likely need to exclude the `neon_lints` package though as it doesn't have anything to document but I'd still want to have `no-documentable-libraries` enabled.

provokateurin:
Let me check what happens when I run dart doc on all packages

provokateurin:
nextcloud package fails because of the references to the enum name getters which dart doc doesn't know about and neon_lints fails because it has no library as you said (so we should exclude it).

provokateurin:
How do we best exclude the neon_lints package? Just in the CI using a filter or is there an option we can use?

Leptopoda:
I think we'll need to add a default dart doc config to the package where `no-documentable-libraries` is disabled.
We might need to add custom configs anyways if we want to include custom markdown files like our architecture overview.

provokateurin:
Can you add that? Then this PR should be ready to merge

provokateurin:
Hm neon_lints doesn't need any dart docs, so we might just disable it completely for that package (if that is possible?)

Leptopoda:
> if that is possible?

We can just add a `--ignore="neon_lints"` to the melos command

provokateurin:
Done

provokateurin (approved):

provokateurin:
Please also "review"

Leptopoda:
LGTM

provokateurin:
Yeah it took 30 mins to complete which is 66% of the total duration of the workflow. Not acceptable IMO (especially if your job doesn't start directly and is stuck in the queue so it takes even longer)

provokateurin:
@Leptopoda how about this: We drop my commit and just merge the fixes and config and I will add the CI for it later?

Leptopoda:
Ok lets do that. Should I or are you doing it?
And we should open an issue for this.

provokateurin:
I'll do both

PR #1064, opened 2023-10-31
fix(neon): Separate Nextcloud and device theme to fix contrast problems

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/975

I'm not sure if there is a better name for the option since even without it the Nextcloud theme is used, but I think this is the easiest to understand for end-users.

I noticed that buttons are now way more readable... We should have really done this earlier.

### DIFF
--- a/packages/neon/neon/lib/src/app.dart
+++ b/packages/neon/neon/lib/src/app.dart
@@ -297,7 +297,7 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, tray.Tra
               builder: (final context, final capabilitiesSnapshot) {
                 final appTheme = AppTheme(
                   capabilitiesSnapshot.data?.capabilities.themingPublicCapabilities?.theming,
-                  keepOriginalAccentColor: options.themeKeepOriginalAccentColor.value,
+                  useNextcloudTheme: options.themeUseNextcloudTheme.value,
                   oledAsDark: options.themeOLEDAsDark.value,
                   appThemes: _appImplementations.map((final a) => a.theme).whereNotNull(),
                   neonTheme: widget.neonTheme,
--- a/packages/neon/neon/lib/src/app.dart
+++ b/packages/neon/neon/lib/src/app.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:collection/collection.dart';
+import 'package:dynamic_color/dynamic_color.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_native_splash/flutter_native_splash.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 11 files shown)


packages/neon/neon/lib/src/theme/theme.dart
```diff
@@ -34,17 +34,26 @@ class AppTheme {
...
-      primary: keepOriginalAccentColorOverride,
-      secondary: keepOriginalAccentColorOverride,
+      primary: primaryColorOverride,
+      onPrimary: onPrimaryColorOverride,
+      secondary: primaryColorOverride,
+      onSecondary: onPrimaryColorOverride,
+      tertiary: primaryColorOverride,
+      onTertiary: onPrimaryColorOverride,
     );
   }
```
Leptopoda, line 58: I think something similar to the follwoing is way more readable.
Having this set up we can also drop the useNextcloudTheme null check with useNextcloudTheme in the constructor.

```dart
  ColorScheme _buildColorScheme(final Brightness brightness) {
    final nextcloudTheme = this.nextcloudTheme;

    var colorScheme = ColorScheme.fromSeed(
      seedColor: neonTheme.colorScheme.primary,
      brightness: brightness,
    );

    if (nextcloudTheme != null) {
      final primaryColor = HexColor(nextcloudTheme.color);
      colorScheme = ColorScheme.fromSeed(
        seedColor: primaryColor,
        brightness: brightness,
      );

      if (useNextcloudTheme) {
        final onPrimaryColor = HexColor(nextcloudTheme.colorText);

        colorScheme = colorScheme.copyWith(
          primary: primaryColor,
          onPrimary: onPrimaryColor,
          secondary: primaryColor,
          onSecondary: onPrimaryColor,
          tertiary: primaryColor,
          onTertiary: onPrimaryColor,
        );
      }
    }

    if (oledAsDark && brightness == Brightness.dark) {
      colorScheme = colorScheme.copyWith(
        background: NcColors.oledBackground,
      );
    }

    return colorScheme;
  }
```

Leptopoda (changes requested):

Leptopoda:
> I'm not sure if there is a better name for the option since even without it the Nextcloud theme is used, but I think this is the easiest to understand for end-users.

What about removing the `Material Nextcloud` theme and offer a system theme (using the material you on android, or the system accent color on mac/linux/windows).

This way it would be an option between the Nextcloud or system theme.
https://pub.dev/packages/dynamic_color

provokateurin:
I don't understand what you mean :thinking:

Leptopoda:
We currently have the nextcloud color scheme and a material one using the nextcloud accent color.
My idea is to use the device accent color scheme instead of the material one with an nc accent color.

This would be independent of this PR but would help avoid confusion with the naming of the option.

provokateurin:
Ah I see. So completely separate the material (based on device) and nextcloud (based on server)  themes, right?

Leptopoda:
Exactly.
We could still offer the mix but with the contrast issues it has I wouldn't recommend to.

provokateurin:
I wouldn't limit the two themes to a platform, I still like to have the Nextcloud theme on my phone :)

provokateurin:
I set the option to be enabled by default now because the official clients also have it by default (I think there isn't even a way to turn it off). Now we have the option to disable it if the users don't like it.

packages/neon/neon/lib/src/app.dart
```diff
@@ -284,42 +285,53 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, tray.Tra
...
-                    ...NeonLocalizations.localizationsDelegates,
-                  ],
-                  supportedLocales: {
-                    ..._appImplementations.map((final app) => app.supportedLocales).expand((final element) => element),
-                    ...NeonLocalizations.supportedLocales,
+  Widget build(final BuildContext context) => DynamicColorBuilder(
+        builder: (final deviceThemeLight, final deviceThemeDark) {
+          if (deviceThemeLight == null || deviceThemeDark == null) {
+            return const SizedBox.shrink();
+          }
```
Leptopoda, line 292: Having null is a perfectly fine value:
>  /// Builds the child widget of this widget, providing a light and dark [ColorScheme].
  ///
  /// The [ColorScheme]s will be null if dynamic color is not supported on the
  /// platform, or if the OS has yet to respond.
  
  The corresponding plugin calls (`DynamicColorPlugin.getCorePalette` and `DynamicColorPlugin.getAccentColor`) return null on unsupported devices. This would mean we'd never display the app for them.
  As we construct our own theme and colorscheme anyways we don't need separate dark and light colorschemes. I'd rather call these methods ourselves in a future builder where we'd also get the connection state and build our colorscheme from the provided one/provided accent color.

packages/neon/neon/lib/src/theme/theme.dart
```diff
@@ -34,18 +42,33 @@ class AppTheme {
...
-
-    return ColorScheme.fromSeed(
-      seedColor: primary,
-      brightness: brightness,
-    ).copyWith(
-      background: oledBackgroundOverride,
-      primary: keepOriginalAccentColorOverride,
-      secondary: keepOriginalAccentColorOverride,
-    );
+    late ColorScheme? colorScheme;
```
Leptopoda, line 45:
```suggestion
    late ColorScheme colorScheme;
```

packages/neon/neon/pubspec.yaml
```diff
@@ -66,7 +67,7 @@ dev_dependencies:
     git:
       url: https://github.com/nextcloud/neon
       path: packages/neon_lints
-  test: ^1.24.9
+  test: ^1.24.3
```
Leptopoda, line 70: Was this intended?

Leptopoda (changes requested):

packages/neon/neon/lib/src/app.dart
provokateurin: Ah I thought null would only happen while the device theme is still loading.
The problem is getting the device theme is async so we need some kind builder widget to handle it.
I don't understand how you would drop one of the color schemes since we need both to build the theme?
It would be better for us if not the color scheme but the raw color palette was exposed.

packages/neon/neon/pubspec.yaml
provokateurin: Yes, the dynamic_color package uses flutter_test in lib and has it as a normal dependency. This breaks some constraints and I had to downgrade this. The same is true for neon_dashboard because flutter_test from sdk and the latest test release are not compatible.

packages/neon/neon/lib/src/app.dart
Leptopoda: >  It would be better for us if not the color scheme but the raw color palette was exposed.

Abd that's exactly what the raw methods do. I encurage you to look at the implementation of the builder and you'll find it obvious:
https://github.com/material-foundation/flutter-packages/blob/main/packages/dynamic_color/lib/src/dynamic_color_builder.dart

provokateurin: I fixed the null thing, but is there something that needs to be changed about the two color schemes?

provokateurin: Well I looked at the implementation of the builder and decided it is easier to use it instead of copying the logic into our own widget.

Leptopoda: Not directly.

provokateurin: And not using it would still require to variables because we need both the colorpalette and the accent color.

provokateurin: So can we just stick to this?

Leptopoda: Let's go with this for now.

Leptopoda (approved):

PR #1067, opened 2023-11-01
Refactor dev container

provokateurin:
- Improves our Talk tests performance because we don't need to use slow apache anymore
- Dev server doesn't crash anymore when resizing terminal
- Static files (news RSS feeds) are served using the dev server for increased performance and less clutter in the tests
- Tests don't try to build the docker image anymore for decreased start up time


packages/nextcloud/test/helper.dart
```diff
@@ -156,8 +156,25 @@ Future<TestNextcloudClient> getTestClient(
   return client;
 }
 
-Future<DockerContainer> getDockerContainer(final DockerImage image, {final bool useApache = false}) async {
+Future<DockerContainer> getDockerContainer() async {
+  const dockerImageName = 'ghcr.io/nextcloud/neon/dev';
+
   late ProcessResult result;
+  result = await runExecutableArguments(
```
Leptopoda, line 163:
```suggestion
  var result = await runExecutableArguments(
```

Leptopoda (commented):
Does this mean we have to manually rebuild the dev container every once in a while?
If so I dislike this approach.

provokateurin:
It is automatically rebuilt when you run `./tool/dev.sh`. The problem with building it inside the tests is that all suites try to do it at the same time and then you end up building the image 8 times in parallel. It also doesn't utilize the GHCR cache, so it's even slower. Relying on the image already being available is much better in my opinion (and we have an error message if it is missing).

Leptopoda (approved):

PR #1068, opened 2023-11-01
fix(neon,neon_files,neon_news,neon_notes): Fix notch clipping content

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/294
I tested with both the normal drawer and the always visible one and both work with this in portrait and landscape mode.

### DIFF
--- a/packages/neon/neon/lib/src/pages/account_settings.dart
+++ b/packages/neon/neon/lib/src/pages/account_settings.dart
@@ -144,10 +144,12 @@ class AccountSettingsPage extends StatelessWidget {
     return Scaffold(
       resizeToAvoidBottomInset: false,
       appBar: appBar,
-      body: Center(
-        child: ConstrainedBox(
-          constraints: NeonDialogTheme.of(context).constraints,
-          child: body,
+      body: SafeArea(
+        child: Center(
+          child: ConstrainedBox(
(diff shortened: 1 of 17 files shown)


Leptopoda:
This does not fix the issue for other pages like the settings or the file info where it's fairly noticeable

provokateurin:
Ah I only fixed it for the main page because of the screenshot in the issue. I'll look into the other pages as well.

Leptopoda:
I'm not sure it'll work but maybe we can just wrap the entire material app with a safe area :thinking:

provokateurin:
The problem is that then the status bar will be just black. We can do it, but apps usually don't do it.

Leptopoda:
Thinking about it some clients might want to display something full screen.
Like the video feed of talk in a call.

Maybe we can create a custom go_route like `SafeAreaRoute` that wraps the given page in a safe area?

provokateurin:
For other pages than those displayed in home we need to individually add the SafeArea. In Talk for example the call is on a new page that isn't rendered inside the home page, so there it wouldn't matter. I think we can just add the SafeArea to all the pages that need it and then we are good.

provokateurin:
@Leptopoda Please review again

Leptopoda (approved):
We should've done this earlier.
Such an easy fix xD

PR #1073, opened 2023-11-01
feat(neon): add ability to display NeonError as a ListTile

Leptopoda:
extracted from #1012

### DIFF
--- a/packages/neon/neon/lib/src/widgets/error.dart
+++ b/packages/neon/neon/lib/src/widgets/error.dart
@@ -11,15 +11,32 @@ import 'package:neon/src/utils/provider.dart';
 import 'package:nextcloud/nextcloud.dart';
 import 'package:universal_io/io.dart';
 
+/// The display mode of the [NeonError] widget.
+enum NeonErrorType {
+  /// Only shows the error icon.
+  ///
+  /// This mode creates an IconButton that is usually used in size constrained
+  /// areas.
+  iconOnly,
+
+  /// Shows a column with the error message and a retry button.
+  column,
+
+  /// Shows a [ListTile] with the error.
+  listTile,
+}
+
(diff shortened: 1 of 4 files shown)


packages/neon/neon/lib/src/widgets/error.dart
```diff
@@ -46,6 +58,11 @@ class NeonError extends StatelessWidget {
   /// Defaults to the nearest [IconTheme]'s [ColorScheme.error].
   final Color? color;
 
+  /// The display mode of this widget.
+  ///
+  /// Defaults to `NeonErrorType.column`.
```
provokateurin, line 63: Shouldn't this be in the constructor because the default value is set in there? Then you can also see this information when looking at the widget in your code directly

packages/neon/neon/lib/src/widgets/error.dart
```diff
@@ -78,56 +95,65 @@ class NeonError extends StatelessWidget {
       color: color,
     );
 
-    final message =
+    final message = Text(
+      details.getText(context),
```
provokateurin, line 99: Since the text is used multiple times it should be put into a variable to avoid recomputation.

packages/neon/neon/lib/src/widgets/error.dart
```diff
@@ -78,56 +95,65 @@ class NeonError extends StatelessWidget {
       color: color,
     );
 
-    final message =
+    final message = Text(
```
provokateurin, line 98: This widget is only used in 2 of the types. Wouldn't it be better to inline it since it is not very complex?

provokateurin (changes requested):

provokateurin (approved):

provokateurin (approved):

PR #1079, opened 2023-11-02
feat(neon): Reset SelectOption if value is not in values

provokateurin:
Towards https://github.com/nextcloud/neon/issues/892, but it doesn't fix it because the AppsBloc is still making problems.

### DIFF
--- a/packages/neon_framework/lib/src/settings/models/option.dart
+++ b/packages/neon_framework/lib/src/settings/models/option.dart
@@ -71,6 +71,7 @@ sealed class Option<T> extends ChangeNotifier implements ValueListenable<T>, Dis
   /// listeners.
   @override
   T get value => _value;
+
   @mustCallSuper
   set value(final T newValue) {
     if (_value == newValue) {
@@ -88,6 +89,7 @@ sealed class Option<T> extends ChangeNotifier implements ValueListenable<T>, Dis
   /// value as evaluated by the equality operator ==, this class notifies its
   /// listeners.
   bool get enabled => _enabled;
+
   @mustCallSuper
   set enabled(final bool newValue) {
     if (_enabled == newValue) {
@@ -108,7 +110,9 @@ sealed class Option<T> extends ChangeNotifier implements ValueListenable<T>, Dis
     final value = deserialize(data);
 
     if (value != null) {
-      this.value = value;
+      // Do not trigger the validation to avoid resetting when the values haven't been loaded yet.
+      _value = value;
+      notifyListeners();
     }
   }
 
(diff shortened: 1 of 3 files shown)


Leptopoda (commented):
Have you made sure that the push distributor option still works?
Some options (initialApp, initialAccount and pushNotificationsDistributor) don't have initial `values`. Therefore the reset might conflict with loading the initial `value`.
I currently could only test the first two this sorry :sweat_smile: 

It might not be a problem as the `forceload` parameter isn't used anywhere (we could maybe remove it in a followup).

Leptopoda (changes requested):
Please squash the first two commits

provokateurin:
>  Some options (initialApp, initialAccount and pushNotificationsDistributor) don't have initial values. Therefore the reset might conflict with loading the initial value.

You were right, I changed it so it only sets `_value`. Once the `values` are set they will validate that the loaded `value` is correct.

I'll do some testing on all these options.

provokateurin:
I tested the 3 critical options and they all still work as expected.

packages/neon/neon/lib/src/utils/global_options.dart
```diff
@@ -101,10 +101,6 @@ class GlobalOptions extends OptionsCollection {
         (final account) => MapEntry(account.id, (final context) => account.humanReadableID),
       ),
     );
-
-    if (!initialAccount.values.containsKey(initialAccount.value)) {
-      initialAccount.reset();
-    }
```
Leptopoda, line 107: This is problematic.
https://github.com/nextcloud/neon/blob/730d5791385478cc26d97e2cdcc2c1cd82626e4d/packages/neon/neon/lib/src/utils/global_options.dart#L30-L37

Sets the value to the first account if the option `rememberLastUsedAccount` is disabled.
If I have at least three accounts logged in and one of them set as the default one.
Then removing this default account leads to the value resetting causing `rememberLastUsedAccount` to be false and the value of `initialAccount` being null. 
I couldn't find any issue with this state but it's not the best to be in.

Leptopoda (changes requested):

packages/neon/neon/lib/src/utils/global_options.dart
Leptopoda: Thinking about this more this issue should also be reproducible on main as we don't do anything different there.
If this is the case we should open a new issue.

provokateurin: I tested it and it works fine, but as you said the state is not great. Instead of recovering from it the next time it should be recovered from immediately.

provokateurin:
The problem of the disappearing initial app is fixed in https://github.com/nextcloud/neon/pull/1310

Leptopoda (approved):

PR #1080, opened 2023-11-02
fix(ci): Fix renovate fvm exclusion

provokateurin:
fvm is currently ignored because it uses the same data source. Restricting this rule to the pub manager makes it work.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -13,6 +13,7 @@
   },
   "packageRules": [
     {
+      "matchManagers": ["pub"],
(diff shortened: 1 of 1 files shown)


Leptopoda:
Out of curiosity does this make it work with the old config (version+channel).

Leptopoda (approved):

provokateurin:
Nope, previously there was an error because it the config file was "invalid". Then it was excluded because of the data source (major bumps would have still worked).

Issue #1099, opened 2023-11-03
Naming collision in allOf


Leptopoda:
The following schema results in a naiming collision:

```json
            "OneObjectAllOf": {
                "description": "All of with one object value.",
                "allOf": [
                    {
                        "type": "object",
                        "required": [
                            "attribute-allOf"
                        ],
                        "properties": {
                            "attribute-allOf": {
                                "type": "string"
                            }
                        }
                    }
                ]
            },
```

```dart
@BuiltValue(instantiable: false)
abstract interface class OneObjectAllOfInterface {
  @BuiltValueField(wireName: 'attribute-allOf')
  String get attributeAllOf;
}

@BuiltValue(instantiable: false)
abstract interface class OneObjectAllOfInterface implements OneObjectAllOfInterface {}

abstract class OneObjectAllOf implements OneObjectAllOfInterface, Built<OneObjectAllOf, OneObjectAllOfBuilder> {
  factory OneObjectAllOf([final void Function(OneObjectAllOfBuilder)? b]) = _$OneObjectAllOf;

  const OneObjectAllOf._();

  factory OneObjectAllOf.fromJson(final Map<String, dynamic> json) =>
      _jsonSerializers.deserializeWith(serializer, json)!;

  Map<String, dynamic> toJson() => _jsonSerializers.serializeWith(serializer, this)! as Map<String, dynamic>;

  static Serializer<OneObjectAllOf> get serializer => _$oneObjectAllOfSerializer;
}
```

provokateurin:
There are many other ways to trigger name collisions :/ We should look for this on a broader scope

Leptopoda:
We should also add e2e test for various problematic attribute names to test the escaping.

Leptopoda:
https://github.com/nextcloud/neon/blob/4643f1f394f6b40b52f7586aaa2882059ee2ca22/packages/dynamite/dynamite_end_to_end_test/lib/type_defs.openapi.dart#L110-L113

Here we already don't escape the `$int` variable name in the throwm error message (it should also have been the type name so `int`).

Issue #1098, opened 2023-11-03
Allof with a primitive schema fails to build


Leptopoda:
While writing tests for the coming someOf changes I noticed that allOf(string) fails to build:
```json
            "PrimitiveAllOf": {
                "description": "All of with an primitive values.",
                "allOf": [
                    {
                        "type": "integer"
                    },
                    {
                        "type": "string"
                    }
                ]
            },
            "MixedAllOf": {
                "description": "All of with object and primitive value.",
                "allOf": [
                    {
                        "type": "string"
                    },
                    {
                        "type": "object",
                        "required": [
                            "attribute-allOf"
                        ],
                        "properties": {
                            "attribute-allOf": {
                                "type": "string"
                            }
                        }
                    }
                ]
            }, 
            "OneValueAllOf": {
                "description": "All of with one primitive value.",
                "allOf": [
                    {
                        "type": "string"
                    }
                ]
            }
```

Leptopoda:
My bad:
> The OpenAPI Specification allows combining and extending model definitions using the allOf property of JSON Schema, in effect offering model composition. allOf takes an array of object definitions that are validated independently but together compose a single object.

allOf only works on objects :woman_facepalming:

provokateurin:
Ah true, I should have noticed that as well

Leptopoda:
We might need to consider support this:
https://github.com/OAI/OpenAPI-Specification/issues/2551

I think the `object` here means a json object and not an openapi object type.
This also exlains why the openapi-generator happily validates the above spec.

Issue #1082, opened 2023-11-02
Merge oneOf double|int to num


provokateurin:
num has the usecase of representing double and int and we can perfectly use it here.

Leptopoda:
If we don't fix this in time for https://github.com/nextcloud/neon/milestone/4 we can just patch the specs and introduce this later.

I can image this not being that trivial.

Leptopoda:
Just adding that this should be optional and configurable within the dynamite config.

Issue #1058, opened 2023-10-30
Fix deserialization of arrays in oneOf


Leptopoda:
arrays of type object or String currently match the intermediate serialized object used by the built_value serializer.
This results in wrong de serialization and potentially an error being thrown.

The following spec is affected by this:
```json
{
  "openapi": "3.1.0",
  "info": {
    "title": "some ofs test",
    "version": "0.0.1"
  },
  "components": {
    "schemas": {
      "Base": {
        "type": "object",
        "required": [
          "attribute"
        ],
        "properties": {
          "attribute": {
            "type": "string"
          }
        }
      },
      "OneOfUnspecifiedArray": {
        "oneOf": [
          {
            "$ref": "#/components/schemas/Base"
          },
          {
            "type": "array"
          }
        ]
      },
      "OneOfStringArray": {
        "oneOf": [
          {
            "$ref": "#/components/schemas/Base"
          },
          {
            "type": "array",
            "items": {
              "type": "string"
            }
          }
        ]
      }
    }
  },
  "paths": {},
  "tags": []
}
```

Leptopoda:
I will give it a try in the coming days.

provokateurin:
This should probably also be removed from the milestone?

Leptopoda:
Let's keep it.
If time allows this should definitely be in the next release after 0.1.0.

Or do you prefer opening a new one for that?

provokateurin:
If we want to continue using milestones we should create a new one instead of always using the same ones.

PR #1100, opened 2023-11-03
test(dynamite_end_to_end_test): test problematic ofs

Leptopoda:
Adds basic tests to the ofs schemas.
I added all scenarios I consider problematic and where I want to work on in the coming weeks (namely #1082 and #1049) minus the allOf issues I found (#1098 and #1099).
Arrays are also still untested and left for #1056 and #1058.

I also still left out unit tests for the generated code as I will first want to implement the changes.
After this we can also unit test the generated code to help us keep a stable api within the generated code.


provokateurin (approved):
Good idea to add tests before trying to change stuff!

PR #1103, opened 2023-11-03
refactor(nextcloud,neon_files): Introduce PathUri for WebDAV path handling

provokateurin:
Based on https://github.com/nextcloud/neon/pull/1096

I still need to test everything and I'm pretty sure I broke some stuff.

### DIFF
--- a/packages/neon/neon_files/lib/blocs/browser.dart
+++ b/packages/neon/neon_files/lib/blocs/browser.dart
@@ -1,25 +1,25 @@
 part of '../neon_files.dart';
 
 abstract interface class FilesBrowserBlocEvents {
-  void setPath(final List<String> path);
+  void setPath(final PathUri uri);
 
-  void createFolder(final List<String> path);
+  void createFolder(final PathUri uri);
 }
 
 abstract interface class FilesBrowserBlocStates {
   BehaviorSubject<Result<List<WebDavFile>>> get files;
 
-  BehaviorSubject<List<String>> get path;
+  BehaviorSubject<PathUri> get uri;
 }
 
 class FilesBrowserBloc extends InteractiveBloc implements FilesBrowserBlocEvents, FilesBrowserBlocStates {
   FilesBrowserBloc(
     this.options,
     this.account, {
-    final List<String>? initialPath,
+    final PathUri? initialPath,
   }) {
     if (initialPath != null) {
-      path.add(initialPath);
+      uri.add(initialPath);
     }
 
     unawaited(refresh());
@@ -31,7 +31,7 @@ class FilesBrowserBloc extends InteractiveBloc implements FilesBrowserBlocEvents
   @override
   void dispose() {
     unawaited(files.close());
-    unawaited(path.close());
+    unawaited(uri.close());
     super.dispose();
   }
 
@@ -39,16 +39,16 @@ class FilesBrowserBloc extends InteractiveBloc implements FilesBrowserBlocEvents
   BehaviorSubject<Result<List<WebDavFile>>> files = BehaviorSubject<Result<List<WebDavFile>>>();
 
   @override
-  BehaviorSubject<List<String>> path = BehaviorSubject<List<String>>.seeded([]);
+  BehaviorSubject<PathUri> uri = BehaviorSubject.seeded(PathUri.cwd());
 
   @override
   Future<void> refresh() async {
     await RequestManager.instance.wrapWebDav<List<WebDavFile>>(
       account.id,
-      'files-${path.value.join('/')}',
+      'files-${uri.value.path}',
       files,
       () => account.client.webdav.propfind(
-        Uri(pathSegments: path.value),
+        uri.value,
         prop: WebDavPropWithoutValues.fromBools(
           davgetcontenttype: true,
           davgetetag: true,
@@ -65,13 +65,13 @@ class FilesBrowserBloc extends InteractiveBloc implements FilesBrowserBlocEvents
   }
 
   @override
-  void setPath(final List<String> p) {
-    path.add(p);
+  void setPath(final PathUri uri) {
+    this.uri.add(uri);
     unawaited(refresh());
   }
 
   @override
-  void createFolder(final List<String> path) {
-    wrapAction(() async => account.client.webdav.mkcol(Uri(pathSegments: path)));
+  void createFolder(final PathUri uri) {
+    wrapAction(() async => account.client.webdav.mkcol(uri));
   }
 }
--- a/packages/neon/neon_files/lib/blocs/files.dart
+++ b/packages/neon/neon_files/lib/blocs/files.dart
@@ -1,25 +1,25 @@
 part of '../neon_files.dart';
 
 abstract interface class FilesBlocEvents {
-  void uploadFile(final List<String> path, final String localPath);
+  void uploadFile(final PathUri uri, final String localPath);
 
-  void syncFile(final List<String> path);
+  void syncFile(final PathUri uri);
 
-  void openFile(final List<String> path, final String etag, final String? mimeType);
+  void openFile(final PathUri uri, final String etag, final String? mimeType);
 
-  void shareFileNative(final List<String> path, final String etag);
+  void shareFileNative(final PathUri uri, final String etag);
 
-  void delete(final List<String> path);
+  void delete(final PathUri uri);
 
-  void rename(final List<String> path, final String name);
+  void rename(final PathUri uri, final String name);
 
-  void move(final List<String> path, final List<String> destination);
+  void move(final PathUri uri, final PathUri destination);
 
-  void copy(final List<String> path, final List<String> destination);
+  void copy(final PathUri uri, final PathUri destination);
 
-  void addFavorite(final List<String> path);
+  void addFavorite(final PathUri uri);
 
-  void removeFavorite(final List<String> path);
+  void removeFavorite(final PathUri uri);
 }
 
 abstract interface class FilesBlocStates {
@@ -58,35 +58,35 @@ class FilesBloc extends InteractiveBloc implements FilesBlocEvents, FilesBlocSta
   BehaviorSubject<List<FilesTask>> tasks = BehaviorSubject<List<FilesTask>>.seeded([]);
 
   @override
-  void addFavorite(final List<String> path) {
+  void addFavorite(final PathUri uri) {
     wrapAction(
       () async => account.client.webdav.proppatch(
-        Uri(pathSegments: path),
+        uri,
         set: WebDavProp(ocfavorite: 1),
       ),
     );
   }
 
   @override
-  void copy(final List<String> path, final List<String> destination) {
-    wrapAction(() async => account.client.webdav.copy(Uri(pathSegments: path), Uri(pathSegments: destination)));
+  void copy(final PathUri uri, final PathUri destination) {
+    wrapAction(() async => account.client.webdav.copy(uri, destination));
   }
 
   @override
-  void delete(final List<String> path) {
-    wrapAction(() async => account.client.webdav.delete(Uri(pathSegments: path)));
+  void delete(final PathUri uri) {
+    wrapAction(() async => account.client.webdav.delete(uri));
   }
 
   @override
-  void move(final List<String> path, final List<String> destination) {
-    wrapAction(() async => account.client.webdav.move(Uri(pathSegments: path), Uri(pathSegments: destination)));
+  void move(final PathUri uri, final PathUri destination) {
+    wrapAction(() async => account.client.webdav.move(uri, destination));
   }
 
   @override
-  void openFile(final List<String> path, final String etag, final String? mimeType) {
+  void openFile(final PathUri uri, final String etag, final String? mimeType) {
     wrapAction(
       () async {
-        final file = await _cacheFile(path, etag);
+        final file = await _cacheFile(uri, etag);
 
         final result = await OpenFile.open(file.path, type: mimeType);
         if (result.type != ResultType.done) {
@@ -98,10 +98,10 @@ class FilesBloc extends InteractiveBloc implements FilesBlocEvents, FilesBlocSta
   }
 
   @override
-  void shareFileNative(final List<String> path, final String etag) {
+  void shareFileNative(final PathUri uri, final String etag) {
     wrapAction(
       () async {
-        final file = await _cacheFile(path, etag);
+        final file = await _cacheFile(uri, etag);
 
         await Share.shareXFiles([XFile(file.path)]);
       },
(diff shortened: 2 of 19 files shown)


packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,76 @@
...
+      : assert(
+          (path == null) != (pathSegments == null) || (path == null && pathSegments == null),
+          'Only one of path and pathSegments must be given',
+        ),
+        _uri = Uri(
+          path: path,
+          pathSegments: pathSegments,
+        ).normalizePath();
+
+  final Uri _uri;
```
Leptopoda, line 20: I don't like having an actual `_Uri` backing ours. Both default implementations of the `Uri` interface `_Uri` and `_SimpleUri` only store the path in a String and compute the segments on the fly.

As we most often use the segments we should use a List (or BuiltList) to back our implementation. This should improve the performance.

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,76 @@
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+/// A [Uri] that only exposes the path handling.
+@immutable
+class PathUri {
```
Leptopoda, line 6: Is there a reason why this does not implement the `Uri` interface?

packages/nextcloud/test/webdav_test.dart
```diff
@@ -19,17 +19,78 @@ void main() {
...
+        expect(
+          WebDavClient.constructUri(baseURL, PathUri(path: '/test/')).toString(),
+          '$sanitizedBaseURL$webdavBase/test',
+        );
       });
     }
   });
 
+  group('PathUri', () {
+    test('path', () {
```
Leptopoda, line 52: I think that the path of an empty `PathUri()` should be `/`.

packages/neon/neon_files/lib/widgets/actions.dart
```diff
@@ -64,15 +65,23 @@ class FileActions extends StatelessWidget {
...
         if (result != null) {
-          bloc.move(details.path, result..add(details.name));
+          bloc.move(
+            details.path,
+            PathUri(
+              pathSegments: [
+                ...result.pathSegments,
+                details.path.basename,
+              ],
+            ),
```
Leptopoda, line 75: We quite often join a path that is a directory with a file name.
I think this could be a helper method in the PathUri that also asserts that the original path is a directory.

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,76 @@
...
+  /// Returns whether the path is a directory.
+  ///
+  /// If the whole path is empty it is considered to be a directory.
+  bool get isDirectory => pathSegments.lastOrNull?.isEmpty ?? true;
+
+  /// Returns the basename of the path.
+  String get basename => pathSegments.where((final s) => s.isNotEmpty).lastOrNull ?? '';
+
+  /// Returns the dirname of the path.
+  PathUri get dirname {
```
Leptopoda, line 41: The name if this getter is confusing.
I'd expect a dirname to be s String and not a Uri.

I suggest to rename the getter `parentDirectory` and add a separate getter `direname` that calls `parentDirectory.basename`

packages/neon/neon_files/lib/models/file_details.dart
```diff
@@ -64,9 +63,7 @@ class FileDetails {
     }
   }
 
-  String get name => path.last;
```
Leptopoda, line 67: I think we should keep this getter.
It made a lot simpler would also keep the diff smaller

Leptopoda (changes requested):

packages/nextcloud/lib/src/webdav/path_uri.dart
provokateurin: Then we would need to expose everything else as well. I thought one of the reasons for this was that we want to hide parts of Uri that we don't want to be exposed?

packages/neon/neon_files/lib/widgets/actions.dart
provokateurin: Yeah I thought about this and didn't find a good solution yet

packages/nextcloud/lib/src/webdav/path_uri.dart
provokateurin: This was based off the `basename` and `dirname` commands, but I agree that `parent` would make more sense.

packages/nextcloud/test/webdav_test.dart
provokateurin: I think `.` which is an empty string here is more intuitive.

Leptopoda: Then please document this behavior in the `PathUri`.

provokateurin:
I will do the testing once you are satisfied with the changes in the nextcloud package. Otherwise I might need to re-do the tedious testing if there is a behavioral change.

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,100 @@
...
+  PathUri? get parent {
+    final nonEmptySegments = _nonEmptySegments;
+    if (nonEmptySegments.length > 1) {
+      return PathUri(
+        pathSegments: [
+          // Potential leading slash
+          if (_segments.first.isEmpty) '',
+          ...nonEmptySegments.sublist(0, nonEmptySegments.length - 1),
+          // Accessing the parent will always lead to a directory so we always have a trailing slash
+          '',
```
Leptopoda, line 71: I have not compleately wrapped my head arround it but I don't like the empty strings in here and somehow don't really understand why we have them.

If this is solely to get a a leading `/` then there are better ways to do this. Maybe the dart `path` library?

packages/neon/neon_files/lib/widgets/browser_view.dart
```diff
@@ -87,12 +87,13 @@ class _FilesBrowserViewState extends State<FilesBrowserView> {
...
                         return NeonListView(
-                          scrollKey: 'files-${pathSnapshot.requireData.join('/')}',
+                          scrollKey: 'files-${pathSnapshot.requireData.path}',
                           itemCount: sorted.length,
                           itemBuilder: (final context, final index) {
                             final file = sorted[index];
-                            final matchingTask = tasksSnapshot.requireData
-                                .firstWhereOrNull((final task) => _pathMatchesFile(task.path, file.name));
+                            final matchingTask = tasksSnapshot.requireData.firstWhereOrNull(
+                              (final task) => widget.bloc.path.value.join([file.name]) == task.path,
```
Leptopoda, line 95: It might make more sense to compare the path and file name separtely.
```suggestion
                              (final task) => file.name == task.path.name && widget.bloc.path.value == task.path.parent,
```
No need to join anything and early aborting if the file name mismatch (the most common)

packages/neon/neon_files/lib/blocs/files.dart
```diff
@@ -115,36 +115,36 @@ class FilesBloc extends InteractiveBloc implements FilesBlocEvents, FilesBlocSta
...
-  void syncFile(final List<String> path) {
+  void syncFile(final PathUri path) {
     wrapAction(
       () async {
         final file = File(
-          p.join(
+          p.joinAll([
             await NeonPlatform.instance.userAccessibleAppDataPath,
             account.humanReadableID,
             'files',
```
Leptopoda, line 145: Not directly related so we can open an issue but we should have a convenience method in the `NeonPlatform` handling this for us.

Just give it the current AppID and get a path to use. This will become especially handy with the sync work.

Leptopoda (changes requested):
I hope I didn't miss anything obvious.
I mainly reviewed the logic and haven't looked at the docs yet. We might also want to rename some parts of the interface as `path.path` is just garbage.

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,100 @@
...
+    return segments
+        .whereIndexed(
+          (final i, final s) => i == 0 || i == segments.length - 1 || s.isNotEmpty,
+        )
+        .toList();
+  }
+
+  final List<String> _segments;
+
+  List<String> get _nonEmptySegments => _segments.where((final s) => s.isNotEmpty).toList();
```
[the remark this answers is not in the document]
provokateurin: Empty segments appear if there is a leading or trailing slash or two slashes next to each other. This is used to find only the parts that matter for path resolving e.g. to figure out the name or parent.

packages/nextcloud/lib/src/webdav/path_uri.dart
provokateurin: We could store this in new fields `isAbsolute` and `isDirectory`, but then it becomes more annoying and error prone to modify and merge paths. It would be a lot cleaner though so we should probably do it. As long as we set both to be required in the constructor we shouldn't make to many mistakes.

provokateurin:
I think it is much cleaner now. I reviewed the changes in neon_files and tested most of it and found no problems (but please also test).

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,154 @@
...
+      if (i == 0 && part.isEmpty) {
+        isAbsolute = true;
+      }
+      if (i == parts.length - 1 && part.isEmpty) {
+        isDirectory = true;
+      }
+      if (part.isNotEmpty) {
+        pathSegments.add(part);
+      }
+    }
```
Leptopoda, line 42:
```suggestion
  factory PathUri.parse(final String path) {
    final parts = path.split('/');
    if (parts.length == 1 && parts.single.isEmpty) {
      return PathUri(
        isAbsolute: false,
        isDirectory: true,
        pathSegments: BuiltList(),
      );
    }

    return PathUri(
      isAbsolute: parts.first.isEmpty,
      isDirectory: parts.last.isEmpty,
      pathSegments: BuiltList(parts.where((final element) => element.isNotEmpty)),
    );
  }
```

packages/nextcloud/test/webdav_test.dart
```diff
@@ -19,17 +19,110 @@ void main() {
...
+      expect(PathUri.parse('/test/abc').parent, PathUri.parse('/test/'));
+      expect(PathUri.parse('/test/abc/').parent, PathUri.parse('/test/'));
+    });
+
+    test('join', () {
+      expect(PathUri.parse('').join(PathUri.parse('test')), PathUri.parse('test'));
+      expect(PathUri.parse('test/').join(PathUri.parse('abc')), PathUri.parse('test/abc'));
+      expect(PathUri.parse('test/').join(PathUri.parse('abc/123')), PathUri.parse('test/abc/123'));
+      expect(() => PathUri.parse('test').join(PathUri.parse('abc')), throwsA(isA<StateError>()));
+    });
```
Leptopoda, line 117: Let's test some funny leading and trailing separators.

```suggestion
      expect(PathUri.parse('/').join(PathUri.parse('test')), PathUri.parse('/test'));
      expect(PathUri.parse('/test/').join(PathUri.parse('abc')), PathUri.parse('/test/abc'));
      expect(PathUri.parse('/test/').join(PathUri.parse('/abc')), PathUri.parse('/test/abc'));
      expect(PathUri.parse('/test/').join(PathUri.parse('/abc/')), PathUri.parse('/test/abc/'));
    });
```

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,154 @@
...
+      isDirectory: other.isDirectory,
+      pathSegments: BuiltList([
+        ...pathSegments,
+        ...other.pathSegments,
+      ]),
+    );
+  }
+
+  /// Renames the last path segment and returns a new path URI.
+  PathUri rename(final String name) => PathUri(
```
Leptopoda, line 130: I think that this is unsafe.
What if the new name is not sanitized and contains slashes?

We should probably throw in that case (needs better handling in the UI maybe with #998)

packages/nextcloud/test/webdav_test.dart
```diff
@@ -19,17 +19,110 @@ void main() {
...
+
+    test('rename', () {
+      expect(PathUri.parse('').rename('test'), PathUri.parse(''));
+      expect(PathUri.parse('test').rename('abc'), PathUri.parse('abc'));
+      expect(PathUri.parse('test/abc').rename('123'), PathUri.parse('test/123'));
+    });
```
Leptopoda, line 123: Please test renaming to paths with slashed in them.
```
      expect(PathUri.parse('test').rename('abc/123'), [...] );
      expect(PathUri.parse('test').rename('abc/'), [...] );
      expect(PathUri.parse('test').rename('/abc'), [...] );
      expect(PathUri.parse('test/').rename('abc/'), [...] );
```

Leptopoda (changes requested):

Leptopoda:
I didn't find any noticeable performance regressions

packages/nextcloud/lib/src/webdav/path_uri.dart
provokateurin: This could also happen when creating a PathUri directly. Should we just remove all slashes and then remove all empty elements?

packages/nextcloud/test/webdav_test.dart
```diff
@@ -19,17 +19,117 @@ void main() {
...
+
+    test('rename', () {
+      expect(PathUri.parse('').rename('test'), PathUri.parse(''));
+      expect(PathUri.parse('test').rename('abc'), PathUri.parse('abc'));
+      expect(PathUri.parse('test/abc').rename('123'), PathUri.parse('test/123'));
+      expect(() => PathUri.parse('test').rename('abc/'), throwsA(isA<Exception>()));
+      expect(() => PathUri.parse('test').rename('/abc'), throwsA(isA<Exception>()));
+      expect(() => PathUri.parse('test').rename('abc/123'), throwsA(isA<Exception>()));
+    });
```
Leptopoda, line 130: This still misses a test for renaming a directory

packages/neon/neon_files/lib/models/file_details.dart
```diff
@@ -64,11 +59,11 @@ class FileDetails {
...
+  String get name => path.name;
 
-  final List<String> path;
+  bool get isDirectory => path.isDirectory;
 
-  final bool isDirectory;
+  final PathUri path;
```
Leptopoda, line 66: I think we should rename this to `uri` so we can avoid `path.path` in the code

packages/neon/neon_files/lib/blocs/browser.dart
```diff
@@ -1,22 +1,22 @@
...
 
-  void createFolder(final List<String> path);
+  void createFolder(final PathUri path);
 }
 
 abstract interface class FilesBrowserBlocStates {
   BehaviorSubject<Result<List<WebDavFile>>> get files;
 
-  BehaviorSubject<List<String>> get path;
+  BehaviorSubject<PathUri> get path;
```
Leptopoda, line 12: This should also be renamed to uri

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,138 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+/// A [Uri] that only exposes the path handling.
```
Leptopoda, line 5:
```suggestion
/// A `Uri` like object that is specialized in file path handling.
```

packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -0,0 +1,138 @@
...
+@immutable
+class PathUri {
+  /// Creates a new path URI.
+  const PathUri({
+    required this.isAbsolute,
+    required this.isDirectory,
+    required this.pathSegments,
+  });
+
+  /// Parses a path URI from a given [path].
```
Leptopoda, line 15:
```suggestion
  /// Creates a new `PathUri` object by parsing a [path] string.
```

Leptopoda (changes requested):
some final changes.

Leptopoda (approved):

provokateurin:
:tada:

PR #1105, opened 2023-11-04
feat(nextcloud): Expose custom HttpClient

provokateurin:
In widget tests we need to be able to mock the HttpClient.


Leptopoda (commented):
I think we should rather mock the entire NextcloudClient.

There is no reason to also test it in the mentioned widget tests.

provokateurin:
Hm true, I would still like to expose the HttpClient in the nextcloud package. The dynamite client has this option so why not expose it as well.

Leptopoda (approved):

PR #1106, opened 2023-11-04
fix(neon_dashboard): Fix matching goldens

provokateurin:


packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -12,6 +12,8 @@ import 'package:neon/src/widgets/error.dart';
 import 'package:neon/src/widgets/linear_progress_indicator.dart';
 import 'package:nextcloud/nextcloud.dart';
 
+export 'package:flutter_cache_manager/flutter_cache_manager.dart' show DefaultCacheManager;
```
Leptopoda, line 15: I don't like exporting it.
I'd rather just add it as a dev dependency where needed.

Leptopoda (changes requested):

Leptopoda (approved):

PR #1107, opened 2023-11-04
fix(neon_dashboard): Remove scrolling inside dashboard widgets

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1075


packages/neon/neon_dashboard/lib/src/pages/main.dart
```diff
@@ -53,4 +83,111 @@ class DashboardMainPage extends StatelessWidget {
...
+  static Iterable<SizedBox> buildWidgetItems({
+    required final BuildContext context,
+    required final dashboard.Widget widget,
+    required final dashboard.WidgetItems? items,
+  }) =>
+      _buildWidgetItems(
+        context: context,
+        widget: widget,
+        items: items,
+      );
```
Leptopoda, line 98: Why not expose the generator function?

Leptopoda (changes requested):
I don't know if I like this approach.
I haven't had a look at the performance but I don't think it will look good.

It looks like the web interface does only render 7 elements and at most one button (counting as an element).
Maybe we should also go this route.

packages/neon/neon_dashboard/lib/src/pages/main.dart
provokateurin: It's not really needed other than internally and for testing, but I guess to keep the code cleaner it is a good idea :+1:

provokateurin:
I understand your concern, but I found this to be the easiest way. Otherwise we have to calculate how many buttons or messages there are and then hide some items accordingly. It gets even harder because the messages and the buttons have different heights.

I'm not sure why you think the performance would be bad (except for calling the generator function twice). After building the widgets like you would normally do we just loop over them once more and read the size. Should be linear complexity so not really concerning?

Leptopoda:
https://api.flutter.dev/flutter/widgets/IntrinsicHeight-class.html

> This class is relatively expensive, because it adds a speculative layout pass before the final layout phase. Avoid using it where possible. In the worst case, this widget can result in a layout that is O(N²) in the depth of the tree.

And this on top of the performance impact you already mentioned. I also dislike that this could potentially result in one list being extremely long forcing others to add more whitespace. This is not pleasing to look at.

Leptopoda:
But if you want I can test the performance impact we can move forward for now.

provokateurin:
> And this on top of the performance impact you already mentioned

Ah I forgot about this :/

> I also dislike that this could potentially result in one list being extremely long forcing others to add more whitespace. This is not pleasing to look at.

I think the chance is quite low. Usually there is 7 items, 1 button and 1 message at max. Technically there could be more than that but I don't think any app actually does that.

> But if you want I can test the performance impact we can move forward for now.

If you don't mind please do.

Leptopoda:
>  Usually there is 7 items, 1 button and 1 message at max.

I currently have two buttons for notes ("Neue Notiz erstellen" and "Weitere Notizen").
I hope we can programatically limit the number of children.

> If you don't mind please do.

Sure. I'll update you in a bit (maybe tomorrow).

provokateurin:
@Leptopoda ping :)

Leptopoda (approved):
Let's just get it done.
Due to the dexing issues I could not verify the impact on a low end phone. There was no measurable impact on my x86 machine.

PR #1113, opened 2023-11-05
refactor(nextcloud): Cleanup spreed helpers

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/366

### DIFF
--- a/packages/nextcloud/lib/src/helpers/spreed.dart
+++ b/packages/nextcloud/lib/src/helpers/spreed.dart
@@ -23,7 +23,7 @@ extension SpreedVersionSupported on spreed.Client {
 /// Conversation types.
 ///
 /// Use [value] to get the integer representation that is used in the API.
-/// See https://github.com/nextcloud/spreed/blob/master/lib/Room.php.
+/// See https://github.com/nextcloud/spreed/blob/main/lib/Room.php.
 enum RoomType {
   /// Room between two participants.
   oneToOne,
@@ -45,12 +45,22 @@ enum RoomType {
 
   /// Integer representation of the [ParticipantType].
   int get value => index + 1;
+
+  /// Converts the integer [value] representation of a [RoomType] into a [RoomType].
+  static RoomType fromValue(final int value) => RoomType.values[value - 1];
+
+  /// Whether the room is only with one other user.
+  bool get isSingleUser => switch (this) {
+        RoomType.oneToOne || RoomType.changelog || RoomType.oneToOneFormer || RoomType.noteToSelf => true,
+        _ => false,
+      };
 }
 
 /// Types of chat messages.
 ///
-/// Use `name` to get the string representation that is used in the API.
-/// See https://github.com/nextcloud/spreed/blob/master/lib/Chat/ChatManager.php.
+/// Use [EnumName.name] to get the string representation that is used in the API.
+/// Use [EnumByName.byName] to convert the string representation into a [MessageType].
+/// See https://github.com/nextcloud/spreed/blob/main/lib/Chat/ChatManager.php.
 enum MessageType {
   /// Message.
   comment,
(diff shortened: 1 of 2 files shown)


packages/nextcloud/lib/src/helpers/spreed.dart
```diff
@@ -174,30 +194,70 @@ enum ParticipantPermission {
...
+    }
+    v = v >> 1;
+  }
+
+  if (result.isEmpty) {
+    result.add(values.first);
+  }
+
+  return result;
+}
```
Leptopoda, line 263: Nice to generalize this but why not make it general extensions for all Enums?
This way we can reuse it anywhere.

the fromBinary must be an extension on List<T> (`EnumFromBinary<T extends Enum> on List<T>`) to get access to the values. This way we also wouldn't need to pass the values around.

This could be something like (caution untested):
```dart
extension EnumBinaryCollection<T extends Enum> on Set<T> {
  int get binaryValue => map((final p) => p.binaryValueFromEnumIndex).reduce((final a, final b) => a | b);
}

extension EnumToBinary on Enum {
  int get binaryValueFromEnumIndex => index == 0 ? 0 : 1 << (index - 1);
}

extension EnumFromBinary<T extends Enum> on List<T> {
  Set<T> enumFromBitValue(final int value) {
    final result = <T>{};

    var v = value;
    for (var i = 1; i <= length - 1; i++) {
      if (v.isOdd) {
        result.add(this[i]);
      }
      v = v >> 1;
    }

    if (result.isEmpty) {
      result.add(first);
    }

    return result;
  }
}
```

We do not need to export these helper methods so we don't confuse users to much.
The helper methods with `fromValue` and `value` inside each enum already handle that.

Leptopoda (changes requested):

packages/nextcloud/lib/src/helpers/spreed.dart
Leptopoda: By not exporting the extensions from the nextcloud package (putting them under `packages/nextcloud/lib/src/helpers/enum_helpers.dart` and annotate them with internal) We would be able to reuse them within the nc package but not need to think about any documentation.

The Enum needing them like `RoomType` would still implement the `fromValue` and to `value` methods that internally call these unexposed ones.

Does that make sense?

```dart
enum SomeBinaryEnum{
	value1,
	value2;

	/// Integer representation of the [SomeBinaryEnum].
	int get value => binaryValueFromEnumIndex;

	/// Converts the integer representation of multiple [SomeBinaryEnum]s to the corresponding [SomeBinaryEnum]s.
	static Set<SomeBinaryEnum> fromValue(final int value) => SomeBinaryEnum.values.enumFromBitValue(value);

}

@internal
extension EnumBinaryCollection<T extends Enum> on Set<T> {
  int get binaryValue => map((final p) => p.binaryValueFromEnumIndex).reduce((final a, final b) => a | b);
}

@internal
extension EnumToBinary on Enum {
  int get binaryValueFromEnumIndex => index == 0 ? 0 : 1 << (index - 1);
}

@internal
extension EnumFromBinary<T extends Enum> on List<T> {
  Set<T> enumFromBitValue(final int value) {
    final result = <T>{};

    var v = value;
    for (var i = 1; i <= length - 1; i++) {
      if (v.isOdd) {
        result.add(this[i]);
      }
      v = v >> 1;
    }
```

Leptopoda: This way we would have one internal method handling all of this (especially as the extensions `ParticipantPermissionsValue` and `ParticipantInCallFlagsValue` are already duplicating logic).

provokateurin: I opted for making the extension public and documented. I added the information about the binary representation and how to handle them to each of the relevant enums.

provokateurin:
I aligned the extension with the EnumName and EnumByName extensions to provide a similar interface.

Leptopoda (approved):

PR #1117, opened 2023-11-06
fix(neon): Fix UnifiedPush distributors not showing up in settings

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1115


Leptopoda (approved):
Looks fine
Can we unit test this? Looking at the imports of the file it doesn't rely on any of our blocs so it should be relatively easy.

provokateurin:
Good idea, especially with the changes coming with https://github.com/nextcloud/neon/pull/1079. I'll do a follow up to land the fix now.

PR #1121, opened 2023-11-06
chore(nextcloud): Update server specs

provokateurin:
API should be stable now for the 28 release.

### DIFF
--- a/packages/nextcloud/lib/src/api/user_status.openapi.json
+++ b/packages/nextcloud/lib/src/api/user_status.openapi.json
@@ -817,7 +817,7 @@
                                                     "$ref": "#/components/schemas/OCSMeta"
                                                 },
                                                 "data": {
-                                                    "oneOf": [
+                                                    "anyOf": [
                                                         {
                                                             "$ref": "#/components/schemas/Private"
                                                         },
--- a/packages/nextcloud/lib/src/api/user_status.openapi.json
+++ b/packages/nextcloud/lib/src/api/user_status.openapi.json
@@ -821,7 +821,10 @@
                                                         {
                                                             "$ref": "#/components/schemas/Private"
                                                         },
-                                                        {}
+                                                        {
+                                                            "type": "array",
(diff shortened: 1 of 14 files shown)


Leptopoda (approved):
Nothing obvious that looks wrong

provokateurin:
I didn't update to the latest version and committed the wrong stuff.

packages/nextcloud/lib/src/api/user_status.openapi.json
```diff
@@ -821,7 +821,10 @@
                                                         {
                                                             "$ref": "#/components/schemas/Private"
                                                         },
-                                                        {}
+                                                        {
+                                                            "type": "array",
+                                                            "maxLength": 0
+                                                        }
```
Leptopoda, line 827: This will probably need to be patched to an anyOf.
I think it was also broken before as `{}` just matches anything.

Leptopoda (changes requested):
Rejecting for now until I have more time to properly test these changes.

provokateurin:
I can also put this patch into a separate PR if you think that's better.

Leptopoda:
I haven't looked into it but I think we should unit test all patched endpoints.
Do you think this is possible to enforce in CI with some regex magic?

provokateurin:
It would be possible to detect these cases somehow, but it's a lot of effort to implement and we should really just fix dynamite. Besides that I don't expect many new of these cases so it's not like it requires constant attention to not break. Just be careful when reviewing new/updated specs.

Leptopoda:
I didn't mean to find the oneOf/anyOf bugs but a regex that ensures that all patched endpoints have unit tests.

Leptopoda (approved):
Unit tests for the patched endpoints would be great but looks good

provokateurin:
I'd say increasing the test coverage is one overall goal. I don't think adding tests for these particular cases is more helpful than testing other endpoints which don't have this problem.

PR #1129, opened 2023-11-12
fix(dynamite,nextcloud): fix generics in nested built_collection buil…

Leptopoda:


provokateurin (approved):
Does it make sense to also add a test to the e2e tests? Just so we can more easily see if there are changes in the future (although you already cover that in the unit tests so we should be fine).

Leptopoda:
We wouldn't really catch the issue in dynamite_e2e.
But it's already on my list to add further tests there.

Issue #1131, opened 2023-11-12
Validate openapi specs


Leptopoda:
We should validate our specs with an openapi validator.
This is especially useful for hand written specs in the dynamite_e2e tests and could avoid things like #1098
It could also make us more confident when applying patches to the generated specs.

provokateurin:
We could also add a script in bin/ which developers can use to validate their specs (without generating any code).

Leptopoda:
Shure.
What tool do you think fits best? I only know the `openapi-builder` that can also validate but I have little trust in that software.

provokateurin:
No I meant that we can build our own based on our spec parsing :)

Leptopoda:
We already partially do so when generating.
But a bug in the parsing for dynamite would not help us validate the e2e specs.
If they are wrong we are testing dynamite with garbage data.

provokateurin:
Of course, I mean it might just be nice to expose this separately from generating code.

PR #1134, opened 2023-11-13
fix(dynamite_end_to_end_test): fix spec validity

Leptopoda:
another issue #1131 could have caught


provokateurin (commented):
I don't understand why it was not valid?

Leptopoda:
oneOf is not an object.
Or am I wrong?

provokateurin:
I assumed that is valid as long as the objects inside oneOf are objects, but I might be wrong. Can you check what exactly the spec says?

Leptopoda:
I could not find any definitive answer.
The openapi-generator happily validates #1098 so we can't trust it either.

provokateurin:
Then I'd say we can support this, as long as the entries inside the nested someOfs are all objects.

Leptopoda:
Closing this for now.
Let's continue the discussion over at #1131

PR #1136, opened 2023-11-14
fix(dynamite): Fix application/octet-stream encoding

provokateurin:
Needed for https://github.com/nextcloud/neon/issues/761

### DIFF
--- a/packages/dynamite/dynamite/lib/src/models/type_result/base.dart
+++ b/packages/dynamite/dynamite/lib/src/models/type_result/base.dart
@@ -30,7 +30,14 @@ class TypeResultBase extends TypeResult {
           return '$object.toString()';
         }
       case 'application/octet-stream':
-        return 'utf8.encode($object) as Uint8List';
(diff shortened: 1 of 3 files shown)


packages/dynamite/dynamite/lib/src/models/type_result/base.dart
```diff
@@ -30,7 +30,11 @@ class TypeResultBase extends TypeResult {
           return '$object.toString()';
         }
       case 'application/octet-stream':
-        return 'utf8.encode($object) as Uint8List';
+        if (className == 'Uint8List') {
+          return object;
+        } else {
+          return 'utf8.encode($object.toString()) as Uint8List';
```
Leptopoda, line 36: I don't think this makes sense.
I'd rather add an elseif to check whether the type is a string and throw otherwise.
The `toString` method can be somewhat unpredictable.

Leptopoda (approved):

PR #1137, opened 2023-11-14
feat(dynamite_petstore_example): Init

provokateurin:
Depends on https://github.com/nextcloud/neon/pull/1136
https://github.com/nextcloud/neon/issues/761

### DIFF
--- a/.cspell/tools.txt
+++ b/.cspell/tools.txt
@@ -21,6 +21,7 @@ enduml
 euxo
 exiftool
 extglob
+fehguy
 freedesktop
 fullpath
 gboolean
@@ -45,9 +46,12 @@ libsqlite
 mipmap
 ndebug
 nodoc
+nonintegers
 nproc
 openrelay
 openrelayprojectsecret
+palo
+petstore
 plantuml
 precache
 puml
--- a/.github/workflows/generate.yaml
+++ b/.github/workflows/generate.yaml
@@ -34,6 +34,8 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
         with:
           fetch-depth: 0
+      - name: Submodules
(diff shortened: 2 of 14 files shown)


provokateurin:
@Leptopoda I can move the fix for the security schemes to a separate PR. With the applied patch it is not necessary anyway since it doesn't break this PR.

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -498,8 +498,14 @@ final authentication = $client.authentications.firstWhereOrNull(
 
   yield* securityRequirements.map((final requirement) {
     final securityScheme = spec.components!.securitySchemes![requirement.keys.single]!;
+    final nameParts = <String?>[
+      securityScheme.type,
+      securityScheme.scheme,
+      securityScheme.$in,
+      securityScheme.name,
+    ].whereNotNull();
```
Leptopoda, line 506: I this fits better as a method in the security scheme object

Leptopoda (approved):
Please also open followup issues for:
- the authentication issue
- to test the petstore against the server

packages/dynamite/dynamite/lib/src/builder/client.dart
provokateurin: I'll make a separate PR then

provokateurin:
yq is failing because of different implementations... I'll find a solution for that

provokateurin:
I'm now using the petstore specification from https://github.com/OAI/OpenAPI-Specification which doesn't have the authentication problems and doesn't need any patching or yq magic.

Leptopoda (approved):

Leptopoda:
Be aware that this uses the utf8 codec so #1143 is conflicting.
I suggest landing the flutter upgrade first

PR #1140, opened 2023-11-14
feat(neon): Support data URIs in NeonUrlImage

provokateurin:
Apps can also return inline SVGs as icons, so we need to support them as well.

### DIFF
--- a/packages/neon/neon/lib/src/widgets/image.dart
+++ b/packages/neon/neon/lib/src/widgets/image.dart
@@ -354,19 +354,27 @@ class NeonUrlImage extends StatelessWidget {
   Widget build(final BuildContext context) {
     final account = this.account ?? NeonProvider.of<AccountsBloc>(context).activeAccount.value!;
 
+    final uri = Uri.parse(url);
+    final dataUri = uri.data;
+
     return NeonCachedImage(
       getImage: () async {
-        final uri = account.completeUri(Uri.parse(url));
+        if (dataUri != null) {
+          return dataUri.contentAsBytes();
+        }
+
+        final completedUri = account.completeUri(uri);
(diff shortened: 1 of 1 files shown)


packages/neon/neon/lib/src/widgets/image.dart
```diff
@@ -354,19 +354,27 @@ class NeonUrlImage extends StatelessWidget {
   Widget build(final BuildContext context) {
     final account = this.account ?? NeonProvider.of<AccountsBloc>(context).activeAccount.value!;
 
+    final uri = Uri.parse(url);
+    final dataUri = uri.scheme == 'data' ? UriData.fromUri(uri) : null;
```
Leptopoda, line 358:
```suggestion
    final dataUri = uri.data;
```
I prefer inlining this in the below check as this variable would be redundant.
Also for your information `uri.isScheme('data')` would be preferred as it does the comparison case insensitive.

Leptopoda (changes requested):

packages/neon/neon/lib/src/widgets/image.dart
provokateurin: The data URI is also used in the hint, that's why I moved it outside.

Leptopoda: Any specific reason you don't like `uri.data` or just so the code doesn't look ugly.
I agree that there isn't any difference.

Leptopoda (approved):

packages/neon/neon/lib/src/widgets/image.dart
provokateurin: Oh I didn't know there was a getter for it. I'll use it instead

PR #1146, opened 2023-11-16
chore(neon_lints): update rules against dart 3.2

Leptopoda:
followup to #1143
I just noticed this when you already merged it :sweat_smile:


provokateurin (approved):
I knew about this change, but since it was not necessary for the CI to pass I didn't add it.

PR #1147, opened 2023-11-16
Update nextcloud package docs

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1087

I'm not sure how much further usage instructions we can give without going into specific endpoints. Most of it should be easily discoverable after reading the readme.

### DIFF
--- a/packages/nextcloud/README.md
+++ b/packages/nextcloud/README.md
@@ -1,11 +1,8 @@
 # nextcloud
 
-A Nextcloud client written in Dart.  
+A Nextcloud API client written in Dart.
 
-This client will become the replacement for https://github.com/provokateurin/dart-nextcloud at some point ([See](https://github.com/nextcloud/neon/issues/1)).
-
-
-## Installing
+## Installation
 
 In the future this code will be available at https://pub.dev/packages/nextcloud, but for now you have to include it via git in your pubspec.yaml:
 ```yaml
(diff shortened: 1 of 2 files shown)


packages/nextcloud/README.md
```diff
@@ -25,15 +22,46 @@ dependency_overrides:
...
+
+It is not guaranteed that an API request will work unless the app is installed and enabled on the server (and has a supported version).  
+
+To get an easier overview of the available endpoints you can browse the [server OpenAPI documentation](https://docs.nextcloud.com/server/latest/developer_manual/_static/openapi.html), but be aware that the package might not be in sync with it.  
+Alternatively you can also go to https://pub.dev/documentation/nextcloud/latest (once the package has been republished at https://pub.dev/packages/nextcloud).
+
+The endpoints are grouped by app and most apps also group their endpoints again.
+They can be accessed using getters on the `NextcloudClient`.
+
+For an example checkout the [example](./example/example.dart).
```
Leptopoda, line 61: What's the benefit of mentioning this?

packages/nextcloud/README.md
```diff
@@ -25,15 +22,46 @@ dependency_overrides:
...
+    Uri.parse('http://localhost'),
+    loginName: 'admin',
+    appPassword: 'xxxxx-xxxxx-xxxxx-xxxx-xxxxx',
+);
+```
 
-To generate such a template take a look at `../../tool/generate-nextcloud.sh`.  
-After you have generated a template, you need to fill it out. Some endpoints can or have to be discarded.  
+Not all endpoints work with just HTTP Basic auth, so it is advised to use app passwords obtained either directly in the Web UI by the user or using the [login flow](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/LoginFlow/index.html#login-flow-v2).  
+Some endpoints do not need any authentication at all or provided extended information when the request is authenticated.
```
Leptopoda, line 49: This sounds off. I guess you mean (but I'm not sure):
```suggestion
Some endpoints do not need any authentication at all or provide extended information when the request needs to be authenticated.
```

packages/nextcloud/README.md
provokateurin: Only mentioning the petstore example could lead people to think that it's the only example we have. We can also remove this if you like

packages/nextcloud/README.md
provokateurin: `is` is correct. The request can be optionally authenticated and in that case it will return more information.

packages/nextcloud/README.md
Leptopoda: Ok my fat fingers somehow messed up the lines.

I don't think we should mention the pet store example at all. The example under `./example` is ok to mention.
The petstore has nothing to with the nextcloud packge and I don't know who it would benefit.

provokateurin: Ah true, it really makes no sense haha. I messed that one up

provokateurin:

Leptopoda (approved):

PR #1151, opened 2023-11-17
fix class name collision on single member allOff

Leptopoda:
…

fixes: #1099 
needed for #1098 and #1049 

Should I keep #1099 open or create a new one for the broader naming collision check?


Leptopoda:
Pushed it under the wrong branch name

PR #1152, opened 2023-11-17
 fix class name collision on single member allOff

Leptopoda:
fixes: #1099 

Should I keep #1099 open or create a new one for the broader naming collision check?


provokateurin:
Let's create a new issue about broader name collision detection/avoidance

PR #1156, opened 2023-11-17
fix allOf with primitive types

Leptopoda:
fixes: #1098
depends on: #1155

### DIFF
--- a/packages/dynamite/dynamite_end_to_end_test/test/all_of_test.dart
+++ b/packages/dynamite/dynamite_end_to_end_test/test/all_of_test.dart
@@ -1,0 +1,78 @@
+import 'package:dynamite_end_to_end_test/all_of.openapi.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('ObjectAllOf', () {
+    final object = ObjectAllOf(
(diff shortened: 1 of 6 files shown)


provokateurin (commented):
Please add a unit test to validate that the serialization and deserialization work.

packages/dynamite/dynamite_end_to_end_test/test/all_of_test.dart
```diff
@@ -0,0 +1,78 @@
...
+
+    expect(object.toJson(), equals(json));
+    expect(OneObjectAllOf.fromJson(json), equals(object));
+  });
+
+  test('PrimitiveAllOf', () {
+    final object = PrimitiveAllOf(
+      (final b) => b
+        ..string = 'stringValue'
+        ..$int = 0,
```
provokateurin, line 38: 0 might not be the best value because it is sometimes used as a default value. I'm not sure if that could ever happen here, but I'd rather avoid it from the beginning.

provokateurin (changes requested):
LGTM, only one minor nitpick

provokateurin (approved):

PR #1157, opened 2023-11-18
refactor(nextcloud): Rework version checks

provokateurin:
Towards https://github.com/nextcloud/neon/issues/571

Allows to define proper minimum and maximum versions.

### DIFF
--- a/packages/nextcloud/lib/src/helpers/common.dart
+++ b/packages/nextcloud/lib/src/helpers/common.dart
@@ -1,2 +1,42 @@
-/// The result of a version check.
-typedef VersionSupported<T> = ({bool isSupported, T minimumVersion});
+import 'package:meta/meta.dart';
+import 'package:version/version.dart';
+
+/// Holds the [versions], [minimumVersion] and [maximumMajor] of an app.
+@immutable
+class VersionCheck {
+  /// Creates a new [VersionCheck].
+  ///
+  /// If the [maximumMajor] is `null` the compatibility of the major of the [minimumVersion] is checked.
+  VersionCheck({
+    required this.versions,
+    required this.minimumVersion,
+    required final int? maximumMajor,
+  }) : maximumMajor = maximumMajor ?? minimumVersion.major;
+
+  /// Current version of the app.
+  final List<Version>? versions;
+
+  /// Minimum version of the app.
+  final Version minimumVersion;
+
+  /// Maximum major version of the app.
+  late final int maximumMajor;
+
+  /// Whether the [versions] is allowed by the [minimumVersion] and [maximumMajor].
+  ///
+  /// If [versions] is `null` or empty it is assumed that the app is supported.
+  /// Only one of the [versions] has to be supported to return `true`.
+  bool get isSupported {
+    if (versions == null || versions!.isEmpty) {
+      return true;
+    }
+
+    for (final version in versions!) {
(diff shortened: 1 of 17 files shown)


packages/nextcloud/lib/src/helpers/common.dart
```diff
@@ -1,2 +1,43 @@
...
+  });
+
+  /// Current version of the app.
+  final Iterable<Version>? versions;
+
+  /// Minimum version of the app.
+  final Version minimumVersion;
+
+  /// Maximum version of the app.
+  final Version? maximumVersion;
```
provokateurin, line 23: For server it would make sense to have a flag to set the maximum version to be exclusive. Then we can specify the max version as 28.0.0 for example instead of 27.99.99

Leptopoda (changes requested):

packages/nextcloud/lib/src/helpers/common.dart
provokateurin: @Leptopoda what do you think?

Leptopoda: Github is garbage.
Because I replied from the review screen my comment is still "pending" :roll_eyes: 

---

Maybe we con do something with the minSupportedVersion.incrementMajor() and just check wether the version is smaller than the next major one.
Similar to what dynamite does.

provokateurin:

packages/nextcloud/test/version_check_test.dart
```diff
@@ -0,0 +1,101 @@
...
+          Version(0, 9, 9),
+          Version(1, 5, 0),
+          Version(1, 6, 0),
+          Version(2, 0, 0),
+        ],
+        minimumVersion: Version(1, 0, 0),
+        maximumExclusiveVersion: Version(1, 9, 9),
+      );
+      expect(check.isSupported, isTrue);
+    });
```
Leptopoda, line 39: Can you explain me how multiple versions work?
This list contains versions that are unsupported by the constraints like `0.9.9` and `2.0.0`. Does this mean that `isSupported` is just checking whether the list contains supported versions?

I couldn't find any helpflul information in the doc comments.

packages/nextcloud/lib/src/helpers/common.dart
```diff
@@ -1,2 +1,51 @@
...
+          }
+        } else {
+          if (version.major == minimumVersion.major) {
+            return true;
+          }
+        }
+      }
+    }
+
+    return false;
```
Leptopoda, line 49: This is still fairly complex. Why not supply a default max version in the constructor. Depending on my other question of how a list of versions should behave this could look:

```dart
  VersionCheck({
    required this.versions,
    required this.minimumVersion,
    required final Version? maximumExclusiveVersion,
  }) : maximumExclusiveVersion = maximumExclusiveVersion ?? minimumVersion.incrementMajor();

/* ... */

    for (final version in versions!) {
      if (version < minimumVersion || version > maximumExclusiveVersion) {
        return false;
      }
    }

    return true;
    
```

packages/nextcloud/test/version_check_test.dart
provokateurin: Yep it checks if at least one version is supported (used for news and notes). I'll add it to the doc comment

Leptopoda (approved):

Issue #1085, opened 2023-11-02
Add DynamiteRawResponse enpoint to webdav


Leptopoda:
Similar to the work done towards #878.

This should be implemented with an experimental tag using the DynamiteRawResponse or a compatible interface.

Leptopoda:
This does not make any sense without #878
I'll push some refinements I already make and close thin in favour of the other issue.

PR #1162, opened 2023-11-19
BytesStreamExtension add stream based xml decoding

Leptopoda:
Leftover from my investigation for #1085
I don't expect any significant performance gains from this. Some quick testing revealed a 3% gain at most which I consider margin of error.

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
@@ -3,13 +3,18 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:universal_io/io.dart';
+import 'package:xml/xml.dart';
+import 'package:xml/xml_events.dart';
 
 /// A stream of bytes.
 ///
 /// Usually a `Stream<Uint8List>`.
 typedef BytesStream = Stream<List<int>>;
 
-final _utf8JsonDecoder = utf8.decoder.fuse(json.decoder);
+final _jsonBytesConverter = utf8.decoder.fuse(json.decoder);
+
+final _xmlBytesConverter =
+    utf8.decoder.fuse(XmlEventDecoder()).fuse(const XmlNormalizeEvents()).fuse(const XmlNodeDecoder());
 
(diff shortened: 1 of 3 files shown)


packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -76,7 +76,7 @@ class DynamiteRawResponse<B, H> {
         _rawBody = switch (bodyType) {
           const FullType(Uint8List) => await response.bytes,
           const FullType(String) => await response.string,
```
provokateurin, line 78: Please also rename those to `bytesBody` and `stringBody`

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: I think we should rather drop the `Body` suffix from xml and json.
The extension methods are on `Stream<List<int>>` and therefore have nothing to do with the response itself.

What do you think?

provokateurin: Sounds good

packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
```diff
@@ -3,13 +3,18 @@ import 'dart:convert';
...
 /// A stream of bytes.
 ///
 /// Usually a `Stream<Uint8List>`.
 typedef BytesStream = Stream<List<int>>;
 
-final _utf8JsonDecoder = utf8.decoder.fuse(json.decoder);
+final _xmlConverter =
+    utf8.decoder.fuse(XmlEventDecoder()).fuse(const XmlNormalizeEvents()).fuse(const XmlNodeDecoder());
+
+final _jsonBytesConverter = utf8.decoder.fuse(json.decoder);
```
provokateurin, line 17: The names are not following the same scheme, is that intended?

provokateurin (approved):

PR #1165, opened 2023-11-20
chore: run melos bootstrap from a clean state

Leptopoda:
Ever since `neon_lints` and now with `neon_dashboard` the overrides weren't sorted alphabetically anymore.
This means running `melos bs` on a fresh clone (or after a `melos clean`) would generate a diff. This is really annoying as I regularly run into this.

Looks like it also cleans up a leftover transitive dependency that still had a registered  plugin.

### DIFF
--- a/packages/app/pubspec_overrides.yaml
+++ b/packages/app/pubspec_overrides.yaml
@@ -1,4 +1,4 @@
-# melos_managed_dependency_overrides: dynamite_runtime,file_icons,neon,neon_files,neon_news,neon_notes,neon_notifications,nextcloud,sort_box,neon_lints,neon_dashboard
+# melos_managed_dependency_overrides: dynamite_runtime,file_icons,neon,neon_dashboard,neon_files,neon_lints,neon_news,neon_notes,neon_notifications,nextcloud,sort_box
 dependency_overrides:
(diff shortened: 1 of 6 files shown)


provokateurin:
Can we somehow add it as a hook before running it so we always get a clean state?

Leptopoda:
I don't always want to run `melos clean` as this will also prune the build cache making development slow.
If you want we can add a separate command but I don't think `melos run clean_bs` is any shorter than `melos clan && melos bs`.

The only reasonable thing I can think about is running `melos clean` in our setup script so updating flutter also cleans all caches.

provokateurin:
I rather thought about only removing the pubspec_overrides.yaml files

Leptopoda:
I don't plan on adding this.
Let's get this merged and if you find value in this functionality go ahead and open a PR. We could also test this in Ci if you really wan.

Leptopoda:
Ok I must have had some issues with my pub cache.
After running `dart pub cache repair` (for something unrelated) I can not reproduce the plugin thing anymore.
The diffs in the overrides still remain.

provokateurin:
Yeah it didn't look correct to me

provokateurin (approved):

PR #1171, opened 2023-11-21
feat(dynamite): Use URI templates for path parameters

provokateurin:
https://github.com/nextcloud/neon/issues/1097

I'm not sure if this really a good improvement but at least it makes our code a little less complex. It would be possible to also put the query parameters into the template, but I think there the risk to introduce bugs is higher than with the current way.

@Leptopoda I'm not sure if you wanted something more from this or not.

I have some ideas to refactor the parameter serialization code to remove make it cleaner, but that's not related to this change.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/imports.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/imports.dart
@@ -27,6 +27,7 @@ Iterable<Spec> generateImports(final AssetId outputId, final State state) sync* 
     Directive.import('package:dynamite_runtime/utils.dart'),
     Directive.import('package:meta/meta.dart'),
     Directive.import('package:universal_io/io.dart'),
+    Directive.import('package:uri/uri.dart'),
     const Code(''),
   ];
 
--- a/packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
+++ b/packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
@@ -44,6 +44,30 @@
                     }
                 }
             }
+        },
+        "/{path-parameter}": {
+            "parameters": [
+                {
+                    "name": "path-parameter",
+                    "in": "path",
+                    "required": true,
+                    "schema": {
+                        "type": "string"
+                    }
+                }
+            ],
+            "get": {
+                "responses": {
+                    "200": {
+                        "description": "",
+                        "content": {
+                            "application/json": {
+                                "schema": {}
+                            }
+                        }
+                    }
+                }
+            }
         }
     },
     "tags": []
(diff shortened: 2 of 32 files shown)


packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
Leptopoda, line 1: Please add a new operation for the path parameter instead of altering an existing one

packages/dynamite/dynamite/lib/src/builder/imports.dart
```diff
@@ -27,6 +27,7 @@ Iterable<Spec> generateImports(final AssetId outputId, final State state) sync*
     Directive.import('package:dynamite_runtime/utils.dart'),
     Directive.import('package:meta/meta.dart'),
     Directive.import('package:universal_io/io.dart'),
+    Directive.import('package:uri/uri.dart'),
```
Leptopoda, line 30: should we already take named imports into account?
In this case `as uri` would probably be prefered.

Leptopoda (commented):
LGTM 
I would like to also handle query parameters in the future but this doesn't need to be done now.

Leptopoda:
> I have some ideas to refactor the parameter serialization code to remove make it cleaner, but that's not related to this change.

Can you explain? Can't we just rely on the serializer (by calling result.encode)?

packages/dynamite/dynamite/lib/src/builder/imports.dart
provokateurin: Let's do them all at once

provokateurin:
> Can you explain? Can't we just rely on the serializer (by calling result.serialize)?

I was thinking of refactoring it such that we can for example omit the `_queryParameters` map completely if there are no query parameters at all. Right now this results in an empty map which can't be optimized away because it is read in the Uri.

Leptopoda:
Yep we should also tackle this.
Wouldn't this be easier if all parameters (path and query) are handled by the UriTemplate?
Then we would add all parameters to the same map and would only need to handle one.

it should then be as easy as checking whether any parameters are specified in the spec and either use the template or the raw path.

provokateurin:
>  Wouldn't this be easier if all parameters (path and query) are handled by the UriTemplate?
Then we would add all parameters to the same map and would only need to handle one.

Yes, but it makes it more complex because we have to change the template and can't just use the path key. We can do it but as I said this can easily introduce new bugs.

Leptopoda:
Ok if you want we can change this after the nc package release.
In that case we should leave the issue open.

I still think that with the provided information of:
https://swagger.io/docs/specification/serialization/
we shouldn't have many bugs. We should also be able to add unit tests with the provided examples within dynamite to catch issues early on.

Leptopoda (approved):

Issue #1101, opened 2023-11-03
Optimize ofs with only one member


Leptopoda:
An of with only one member is useless and can be cleaned up to a normal parameter.
This ofs optimization by dynamite should be configurable and is similar to #1082 and relies on the rework in #1049.

PR #1173, opened 2023-11-22
feat(dynamite): use records for someOfs

Leptopoda:
fixes: #1049 
fixes: #1101

#### FollowUps:
- Other optimizations will come in followup commits (#1178)
- Re introduce the discriminator. Might get complicated :(


provokateurin:
I assume there is no way to split this PR into smaller PRs (or commits)?

packages/dynamite/dynamite/lib/src/builder/resolve_interface.dart
```diff
@@ -74,11 +74,18 @@ Method generateProperty(
...
           ..docs.addAll(description);
 
+        final builtValueFieldAnnotations = <String, Expression>{};
         if (name != propertyName) {
+          builtValueFieldAnnotations['wireName'] = literalString(propertyName);
+        }
+
+        //if (type is TypeResultSomeOf) {
+        //  builtValueFieldAnnotations['serialize'] = literalFalse;
+        //}
```
provokateurin, line 84: Unused code

packages/nextcloud/test/core_test.dart
```diff
@@ -190,7 +190,7 @@ void main() {
         });
       });
     },
-    retry: retryCount,
+    retry: 1,
```
Leptopoda, line 193: This needs reverting

packages/dynamite/dynamite/lib/src/builder/resolve_interface.dart
Leptopoda, line 84: Leftovers from before we had built_value_generator 8.8

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
```diff
@@ -0,0 +1,97 @@
...
+    if (optimizedSubTypes.length == 1) {
+      return optimizedSubTypes.single;
+    }
+
+    final record = optimizedSubTypes.map((final type) {
+      final dartType = type.nullableName;
+      final dartName = toDartName(dartType);
+      final fieldName = toFieldName(dartName, type.className);
+
+      return '$dartType $fieldName';
```
provokateurin, line 38: This could be set to final, no?

Leptopoda:
> I assume there is no way to split this PR into smaller PRs (or commits)?

Is it that bad?
I'll have a look whether I can split it up further.

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
```diff
@@ -0,0 +1,97 @@
...
+  @override
+  String? get _serializer {
+    if (optimizedSubTypes.length <= 1) {
+      return null;
+    }
+
+    return '..add(${className}Extension.serializer)';
+  }
+
+  late final List<TypeResult> optimizedSubTypes = subTypes.toList();
```
provokateurin, line 25: I know this is related to the follow-up PR, but please remove it from here since it is not used.

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
```diff
@@ -0,0 +1,243 @@
...
+      ..body = const Code('dynamite_utils.validateOneOf(_values)');
+  });
+
+  final anyOfValidator = Method((final b) {
+    b
+      ..name = 'validateAnyOf'
+      ..returns = refer('void')
+      ..lambda = true
+      ..body = const Code('dynamite_utils.validateAnyOf(_values)');
+  });
```
provokateurin, line 64: Instead of adding these methods to the interfaces the relevant places could just call the methods directly.

Edit 1: I see that the _values field is private, is that why you added these proxy methods to not expose the field directly?

Edit 2: Since the values in the _values field are not private it would be possible to drop the additonal methods and at the invocation dynamically create the list of arguments. Honestly I don't care which of these options you chose in the end :woman_shrugging:

provokateurin:
>  I'll have a look whether I can split it up further.

I think I was a bit too fast. I saw the PR touched many files, but most changes are generated anyway. So this is already ok to review :)

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
```diff
@@ -0,0 +1,243 @@
...
+    (final b) => b
+      ..name = serializerName
+      ..implements.add(refer('PrimitiveSerializer<$identifier>'))
+      ..constructors.add(Constructor((final b) => b.constant = true))
+      ..methods.addAll([
+        Method(
+          (final b) => b
+            ..annotations.add(refer('override'))
+            ..returns = refer('Iterable<Type>')
+            ..type = MethodType.getter
```
provokateurin, line 124: Can just be a const field

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
```diff
@@ -0,0 +1,243 @@
...
+            ..name = 'types'
+            ..lambda = true
+            ..body = Code('const [$identifier]'),
+        ),
+        Method(
+          (final b) => b
+            ..annotations.add(refer('override'))
+            ..returns = refer('String')
+            ..type = MethodType.getter
```
provokateurin, line 133: Can also just be a const field

provokateurin (changes requested):
Already looks very good and it's definitely a nice cleanup of our generated code!

provokateurin:
Btw why do you think discriminator will get complicated? It is basically a oneOf but instead of trying to deserialize every option you only try to deserialize one of them based on the key.

Leptopoda:
> Btw why do you think discriminator will get complicated? It is basically a oneOf but instead of trying to deserialize every option you only try to deserialize one of them based on the key.

Because built_value still works on the concept of types and not fields. This means all `({int? $int, String string})` need to have the same serializer. No matter whether it is a oneOf, anyOf or a type that needs the discriminator.

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
Leptopoda: what do you mean?
`return 'final $dartType $fieldName';`? This will not work as it would result in a record of `({final int? $int, final String string})` which isn't valid dart syntax.

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
Leptopoda: I quite like calling validate directly on the type.
Imagine:
```dart
var myRecord = (10, 0.5, 'test');
myRecord.validateOneOf();
``` 

I'll add this to the validation hook of the object having the record.
This is needed as the records are currently not validated at deserialization as a serializer can be used for both `oneOf`s and `anyOf`s.

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
Leptopoda: Nope. `avoid_field_initializers_in_const_classes`
https://dart.dev/tools/linter-rules/avoid_field_initializers_in_const_classes

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
Leptopoda: see above

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
provokateurin: Oh I thought this was possible 🫣

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
provokateurin: Anything that needs to be changed or is this done?

Leptopoda: The validation is still missing and just adding it will lead to multiple extension methods being declared on the same type.
https://dart.dev/tools/diagnostic-messages#ambiguous_extension_member_access

Should be trivial to fix but probably something I'll do on monday.

packages/dynamite/dynamite/lib/src/builder/resolve_ofs.dart
```diff
@@ -0,0 +1,69 @@
...
+            s.formattedDescription,
+          );
+
+          methods.add(property);
+        }
+      }
+    }
+
+    state.output.add(
+      buildInterface(identifier, interfaces: interfaces, methods: methods.build()),
```
provokateurin, line 59: Add a trailing comma for readability

packages/dynamite/dynamite/lib/src/builder/resolve_ofs.dart
```diff
@@ -0,0 +1,69 @@
...
+    );
+
+    state.output.add(
+      buildBuiltClass(
+        identifier,
+      ),
```
provokateurin, line 65: Here the trailing commas can be dropped

provokateurin (approved):

PR #1174, opened 2023-11-22
feat(dynamite): allow per spec configuration

Leptopoda:
I don't think we should merge this without using it.
I just wanted to get feedback and avoid letting the branch just rot on my machine

### DIFF
--- a/packages/dynamite/dynamite/README.md
+++ b/packages/dynamite/dynamite/README.md
@@ -55,9 +55,11 @@ targets:
             - 'static BuiltSet<.*> get values => _\$.*Values;'
           overrides:
             # Tighten linting rules for specific specs
+            # Mark the generated library as experimental
             lib/my_spec.openapi.json:
               analyzer_ignores:
                 - camel_case_types
+              experimental: true
 
 ```
 
--- a/packages/dynamite/dynamite/README.md
+++ b/packages/dynamite/dynamite/README.md
@@ -53,6 +53,11 @@ targets:
             - 'factory .*\.fromJson\(Map<String, dynamic> json\) => jsonSerializers\.deserializeWith\(serializer, json\)!;'
             - 'Map<String, dynamic> toJson\(\) => jsonSerializers\.serializeWith\(serializer, this\)! as Map<String, dynamic>;'
             - 'static BuiltSet<.*> get values => _\$.*Values;'
+          overrides:
+            # Tighten linting rules for specific specs
+            lib/my_spec.openapi.json:
+              analyzer_ignores:
+                - camel_case_types
 
 ```
 
--- a/packages/dynamite/dynamite/lib/src/models/dynamite_config/config.dart
+++ b/packages/dynamite/dynamite/lib/src/models/dynamite_config/config.dart
@@ -40,6 +40,14 @@ abstract class DynamiteConfig implements Built<DynamiteConfig, DynamiteConfigBui
   // overrides is reserved in the build.yaml
   BuiltMap<String, DynamiteConfig>? get overrides;
 
+  /// Whether the generated library should be marked with the `@experimental` annotation.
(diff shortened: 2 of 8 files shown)


provokateurin:
I think for testing this we could change some of the rules discussed in https://github.com/nextcloud/neon/issues/1245 for a single (separate) spec.

Leptopoda:
My long term goal was to also generate doc comments for the generated `Client` so we can just enable `public_member_api_docs` while testing our documentation at the same time.

provokateurin:
Another real world use case for this would be annotating the news, notes and uppush code as experimental because those are specs are written by hand and are not complete.

Leptopoda:
I'll rebase and take a look.

Leptopoda:
I've added the `experimental` annotation.
It's currently based on #1337 so tests might fail due to #1343

Leptopoda:
should be ready for review @provokateurin

packages/dynamite/dynamite/README.md
```diff
@@ -53,6 +53,13 @@ targets:
             - 'factory .*\.fromJson\(Map<String, dynamic> json\) => jsonSerializers\.deserializeWith\(serializer, json\)!;'
             - 'Map<String, dynamic> toJson\(\) => jsonSerializers\.serializeWith\(serializer, this\)! as Map<String, dynamic>;'
             - 'static BuiltSet<.*> get values => _\$.*Values;'
+          overrides:
+            # Tighten linting rules for specific specs
+            # Mark the generated library as experimental
+            lib/my_spec.openapi.json:
```
provokateurin, line 59: Here the spec is referenced by the full path

packages/nextcloud/build.yaml
```diff
@@ -18,3 +18,10 @@ targets:
             - 'factory .*\.fromJson\(Map<String, dynamic> json\) => jsonSerializers\.deserializeWith\(serializer, json\)!;'
             - 'Map<String, dynamic> toJson\(\) => jsonSerializers\.serializeWith\(serializer, this\)! as Map<String, dynamic>;'
             - 'static BuiltSet<.*> get values => _\$.*Values;'
+          overrides:
+            news.openapi.json:
+              experimental: true
+            notes.openapi.json:
+              experimental: true
+            uppush.openapi.json:
+              experimental: true
```
provokateurin, line 27: But here it is just the file name. It should be the whole path relative to the root of the package.

provokateurin (changes requested):

packages/nextcloud/build.yaml
Leptopoda: Should I just drop the regex altogether and always require full paths?

provokateurin: Yeah that is better to avoid future problems.

provokateurin: I doubt that this will lead to a lot of duplicate configuration

packages/dynamite/dynamite/lib/src/models/dynamite_config/config.dart
```diff
@@ -33,6 +33,66 @@ abstract class DynamiteConfig implements Built<DynamiteConfig, DynamiteConfigBui
...
+    }
+
+    DynamiteConfig? override;
+    for (final entry in overrides!.entries) {
+      if (entry.key == uri) {
+        override = entry.value;
+        break;
+      }
+    }
+
```
provokateurin, line 66: We might want to check here that the overrides do not contain overrides themselves

provokateurin (approved):

packages/dynamite/dynamite/lib/src/models/dynamite_config/config.dart
Leptopoda: it is a few lines below

PR #1176, opened 2023-11-23
feat(dynamite): allow patterns and string lengths in objects

Leptopoda:
this is also the ground work for the new validation used in #1173 
I think adding #1056 should now also be trivial.


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -452,25 +453,16 @@ Iterable<String> buildParameterSerialization(
...
+Iterable<String> buildParameterPatternCheck(
   final openapi.Parameter parameter,
 ) sync* {
-  final value = toDartName(parameter.name);
-  final name = "'$value'";
-
   final schema = parameter.schema;
-  if (result.name == 'String' && schema != null) {
-    if (schema.pattern != null) {
-      yield "checkPattern($value, RegExp(r'${schema.pattern!}'), $name); // coverage:ignore-line";
```
Leptopoda, line 465: @provokateurin feel free to suggest a regex that re introduces these checks.
or just add another commit to this branch.

provokateurin: I don't think it is needed anymore since you always call the method and skip the check if the value is null.
Even if this reduces our coverage right now I'd add it later.
I'll make a round of coverage cleanups after all the dynamite changes that are necessary for the release are merged.

Leptopoda:
@provokateurin I think it should be ready for review now.

packages/dynamite/dynamite/lib/src/builder/client.dart
provokateurin: Ok no I misunderstood this. Please just add the single line coverage ignores back to the checks. For an object it doesn't matter as I said because the methods are always called, but this is not the case for nullable parameters.

provokateurin (changes requested):
LGTM, just needs the coverage ignores to be added back.

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: Can you suggest a regex?
I'd like to not ignore it to get a full coverage report in the e2e tests (I think anything with `dynamite_utils` could be safely ignored).

provokateurin: Ah true, the ignore doesn't really make much sense. If the check is not called it should show up like that in the coverage :+1:

PR #1178, opened 2023-11-24
feat(dynamite): allow patterns and string lengths in objects

Leptopoda:
fixes: #1082 
depends on: #1173


Leptopoda:
We should probably also optimize combinations with `num` like `anyOf(num|int)`.
This would be a really bad spec but is still valid.

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
```diff
@@ -41,6 +45,33 @@ abstract class TypeResultSomeOf extends TypeResult {
...
 
+  Iterable<TypeResult> get _optimizedSubTypes sync* {
+    var optimizedNum = false;
+
+    final numbers = subTypes.where(
+      (final type) => switch (type.className) {
+        'int' || 'double' || 'num' => true,
+        _ => false,
+      },
+    );
```
provokateurin, line 56: This check is duplicated below, please move it to a separate function

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
```diff
@@ -0,0 +1,128 @@
...
+      return '$dartType $fieldName';
+    });
+
+    return TypeResultBase('({${record.join(',')}})');
+  }
+
+  Iterable<TypeResult> get _optimizedSubTypes sync* {
+    var optimizedNum = false;
+
+    final numbers = subTypes.where(
```
provokateurin, line 51:
```suggestion
    final numbersCount = subTypes.where(
```

provokateurin (changes requested):
Looks pretty good to me, only two small nitpicks

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
Leptopoda: this is a collection thus I'd not call it `count`

provokateurin: Ah true, but since you only ever access the length you could only store that instead

Leptopoda: It looks ugly like that 😅

provokateurin: Ok 🤷‍♀️

packages/dynamite/dynamite/lib/src/models/type_result/some_of.dart
```diff
@@ -49,13 +51,42 @@ abstract class TypeResultSomeOf extends TypeResult {
...
+    for (final type in optimizedSubTypes) {
       buffer.write(toDartName(type.className, uppercaseFirstCharacter: true));
     }
 
     return buffer.toString();
   }
 
+  BuiltList<TypeResult> get _optimizedSubTypes {
+    final optimized = ListBuilder<TypeResult>();
+    final numberCount = subTypes.where(_isNumber).length;
```
provokateurin, line 63: This variable could be inlined as it is only used once and doesn't make the code less readable imo.

provokateurin (approved):

PR #1130, opened 2023-11-12
generate custom serializers for all classes

Leptopoda:
Depends on #1129
Needed for #1049

PR #1182, opened 2023-11-25
refactor(dynamite): let built_value_generator generate the serializers

Leptopoda:
reverts #1130


provokateurin (approved):
Does this affect the header serialization that we still need?

Leptopoda:
No it does not affect it and we can change it long term.
But nothing that we need to do immediately.

PR #1185, opened 2023-11-26
Add separate nextcloud_test helper package

provokateurin:
Towards https://github.com/nextcloud/neon/issues/571


Leptopoda:
Can you explain me why we need a separate package?

provokateurin:
The helper code is getting quite complex and with the changes for https://github.com/nextcloud/neon/issues/571 it gets even more complex. This also brings together the docker stuff that used to live in `tool/` but rather belongs to the nextcloud (or now nextcloud_test) package.

packages/nextcloud_test/lib/src/docker_container.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// Reads the server logs.
+  Future<String> serverLogs() async => (await runExecutableArguments(
+        'docker',
+        [
+          'logs',
+          id,
+        ],
+        stdoutEncoding: utf8,
+      ))
+          .stdout as String;
```
Leptopoda, line 93: Please give this method a body like:
```dart
  Future<String> serverLogs() async {
  final result = await runExecutableArguments(
        'docker',
        [
          'logs',
          id,
        ],
        stdoutEncoding: utf8,
      );

return result.stdout as String;
}
```

packages/nextcloud_test/lib/src/docker_container.dart
```diff
@@ -0,0 +1,112 @@
...
+        'docker',
+        [
+          'exec',
+          id,
+          'cat',
+          'data/nextcloud.log',
+        ],
+        stdoutEncoding: utf8,
+      ))
+          .stdout as String;
```
Leptopoda, line 106: Please make this a method with a body (see above)

packages/nextcloud_test/lib/src/test_client.dart
```diff
@@ -0,0 +1,79 @@
...
+
+/// An extension for creating [NextcloudClient]s based on [DockerContainer]s.
+extension TestNextcloudClient on NextcloudClient {
+  /// Creates a new [NextcloudClient] for a given [container] and [username].
+  ///
+  /// It is expected that the password of the user matches the its [username].
+  /// This is the case for the available test docker containers.
+  static Future<NextcloudClient> create(
+    final DockerContainer container, {
+    final String? username = 'user1',
```
Leptopoda, line 18: In what situation does it make sense to have a null username?
Our tests currently don't make use of it.

Or do we need it to test public apis?

packages/nextcloud_test/lib/src/test_client.dart
```diff
@@ -0,0 +1,79 @@
...
+          i++;
+          await Future<void>.delayed(const Duration(milliseconds: 100));
+          if (i >= 300) {
+            throw TimeoutException('Failed to get the status of the Server. $error');
+          }
+        } else {
+          rethrow;
+        }
+      }
+    }
```
Leptopoda, line 75: I don't think that we should catch `DynamiteApiException`s.

Leptopoda (changes requested):

packages/nextcloud_test/lib/src/test_client.dart
provokateurin: Yes that will be the use-case I think. We can also make it non-nullable for now until we need it.

packages/nextcloud_test/lib/src/test_client.dart
provokateurin: I'll give it a test to see what exceptions we actually catch here, since the HttpException was only added for apache which we no longer use anywhere.

provokateurin: Ok so on my machine the web server starts fast enough to never throw any exceptions. Let's see if GH actions will run into a problem due to being slower hardware. If not we can completely remove this whole while loop and catch block and even save one http request.

provokateurin: Yay, no problems in the CI as well. Removed all of this code

PR #1187, opened 2023-11-26
fix(nextcloud): Only advertise server 27.1 compatibility

provokateurin:
We actually don't support 27.0 because of the new dashboard API.
Will be fixed with https://github.com/nextcloud/neon/issues/571


Leptopoda (approved):
So this means that you'll patch the spec in a followup to make it work again?

provokateurin:
Yep, once we also run our tests against 27.0 versions

PR #1196, opened 2023-11-28
feat(nextcloud,neon_dashboard): Add compatibility for server 27.0

provokateurin:
https://github.com/nextcloud/neon/issues/571


Leptopoda (changes requested):

Leptopoda:
Same as the spreed patch.
Please add a version number to the patch name.

packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
```diff
@@ -54,9 +54,11 @@ class DashboardBloc extends InteractiveBloc implements DashboardBlocEvents, Dash
       final response = await _account.client.dashboard.dashboardApi.getWidgets();
 
       for (final widget in response.body.ocs.data.values) {
-        if (widget.itemApiVersions.contains(2)) {
+        final itemApiVersions = widget.itemApiVersions;
+        if (itemApiVersions != null && itemApiVersions.contains(2)) {
           v2WidgetIDs.add(widget.id);
-        } else if (widget.itemApiVersions.contains(1)) {
+        } else if (itemApiVersions == null || itemApiVersions.contains(1)) {
```
Leptopoda, line 60: Please reorder the two checks so the null check is first.

provokateurin: But you also wanted to have the checks unified. Some widgets have both v1 and v2 and we prefer v2 so we need to check for v1 after v2. Either the null check has to stay here or you don't have unified checks. I'm fine with either one.

Leptopoda: I didn't know that a widget can support both.
This makes sense so lets leave it as is.

Leptopoda (approved):

PR #1198, opened 2023-11-29
chore(nextcloud): Order patches

provokateurin:


packages/nextcloud/lib/src/patches/spreed/1-compatibility-17.1.json
```diff
@@ -1,7 +1,7 @@
 [
```
Leptopoda, line 1: Why isn't the name the same as the others?
Looks more like the oneOf workaround to me

provokateurin: This is because the order changed. Now the compatibility patch is applied first so the oneOfs are gone. I only had to change the paths in the patches, the content didn't change at all.

provokateurin: I think you got confused by the diff that is displayed here. Just check how the full patch looks like, then you'll understand.

Leptopoda (approved):

PR #1205, opened 2023-11-30
feat(nextcloud,neon): Add compatibility for server 26 and spreed 16

provokateurin:
Closes https://github.com/nextcloud/neon/issues/571


provokateurin:
Hm so the CI always kills itself for unknown reasons

provokateurin:
We should try to only build the images that we test against. That will reduce the used images drastically and fix it for now.

Leptopoda:
I don't have any particular opinion but I agree that we should be able to count on semver.
It's also the easiest to fix it rn.

provokateurin:
Rebased onto https://github.com/nextcloud/neon/pull/1216 to check if the failure goes away

packages/nextcloud/lib/src/patches/spreed/1-compatibility-17.1.json
provokateurin, line 1: Uh these changes should be a different patch

PR #1215, opened 2023-12-02
fix(dynamite): serialization of someOfs with an array member

Leptopoda:
towards #1058 

Looks like that migrating to records already fixed the deserialization. And now only serialization is broken.
This is due to a limitation/oversight in the upstream `StandartJsonPlugin` that is responsible to convert the intermediate built_value wire format into a Json output.

```dart
  @override
  Object? afterSerialize(Object? object, FullType specifiedType) {
    if (object is List &&
        specifiedType.root != BuiltList &&
        specifiedType.root != BuiltSet &&
        specifiedType.root != JsonObject) {
      if (specifiedType.isUnspecified) {
        return _toMapWithDiscriminator(object);
      } else {
        return _toMap(object, _needsEncodedKeys(specifiedType));
      }
    } else {
      return object;
    }
  }
  ```
  
  I'll ask upstream. Maybe we can find a simple solution (I don't really want to write a replacement to the json plugin).

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/serializer.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/serializer.dart
@@ -62,7 +62,17 @@ List<Spec> buildSerializer(State state) => [
               ])
               .cascade('addPlugin')
               .call([
-                refer('StandardJsonPlugin', 'package:built_value/standard_json_plugin.dart').newInstance(const []),
+                refer('StandardJsonPlugin', 'package:built_value/standard_json_plugin.dart').newInstance(const [], {
+                  if (state.uniqueSomeOfTypes.isNotEmpty)
+                    'typesToLeaveAsList': literalConstSet(
+                      state.uniqueSomeOfTypes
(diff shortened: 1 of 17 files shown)


provokateurin:
This test is supposed to fail until the problem is fixed, right? If so please make it a draft for now.

provokateurin:
But it doesn't fail? :see_no_evil:

Leptopoda:
Yes it doesn't because I currently skip the failing tests.

provokateurin:
Ahhh :bulb: Maybe remove the skip for now so we can actually see that it fails (and don't forget to remove it later).

provokateurin:
Btw did you open an issue upstream and can you link it here?

Leptopoda:
Finally opened an issue upstream: https://github.com/google/built_value.dart/issues/1295

Leptopoda:
It's working :tada: 
No we only need to wait for the next release.

provokateurin (approved):

PR #1223, opened 2023-12-03
fix(dynamite): resolve arrays with a maximum length of 0 to BuiltList…

Leptopoda:
…<Never>

Fixes the issue we had with `nextcloud` and #1058.
This does not solve the root cause of the issue.


provokateurin:
user_status also has a patch that needs to be removed and please rename the remaining spreed patches so they start with 0 again.

Issue #1167, opened 2023-11-20
Check dependency constraints when generating with dynamite


provokateurin:
`dynamite_runtime` and others (to be listed here) need to be checked for compatibility (similarly to how json_serializable does it).

PR #1228, opened 2023-12-04
feat(dynamite)!: validate version of dynamite_runtime when generating

Leptopoda:
fixes: #1167

### DIFF
--- a/packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
+++ b/packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
@@ -1,0 +1,96 @@
+import 'package:build/build.dart';
+import 'package:pub_semver/pub_semver.dart';
+import 'package:pubspec_parse/pubspec_parse.dart';
+
+final dependencies = {
+  'built_collection': Version.parse('5.0.0'),
+  'built_value': Version.parse('8.0.0'),
+  'collection': Version.parse('1.0.0'),
+  'dynamite_runtime': Version.parse('0.0.1'),
+  'meta': Version.parse('1.0.0'),
+  'universal_io': Version.parse('2.0.0'),
+  'uri': Version.parse('1.0.0'),
+};
+
+final devDependencies = {
+  'built_value_generator': Version.parse('8.8.0'),
+};
+
+/// Checks whether the correct version of the dependencies are present in the pubspec.yaml file.
+Future<({bool hasFatal, String messages})> helperVersionCheck(final BuildStep buildStep) async {
+  final pubspecAsset = AssetId(buildStep.inputId.package, 'pubspec.yaml');
+
+  if (!await buildStep.canRead(pubspecAsset)) {
+    return (
+      hasFatal: false,
+      messages: 'Failed to read the pubspec.yaml file. Version constraints of required packages can not be validated.',
+    );
+  }
+
+  final pubspecContent = await buildStep.readAsString(pubspecAsset);
+  final pubspec = Pubspec.parse(pubspecContent, sourceUrl: pubspecAsset.uri);
+
+  final messages = StringBuffer();
+  var hasFatal = false;
+
+  for (final constraint in dependencies.entries) {
+    final result = _validateVersion(pubspec.dependencies, constraint.key, constraint.value);
+
+    if (result.message != null) {
+      messages.writeln(result.message);
+    }
+
+    if (result.isFatal) {
+      hasFatal = true;
+    }
+  }
+
+  for (final constraint in devDependencies.entries) {
+    final result = _validateVersion(pubspec.devDependencies, constraint.key, constraint.value);
+
+    if (result.message != null) {
+      messages.writeln(result.message);
+    }
+
+    if (result.isFatal) {
+      hasFatal = true;
+    }
+  }
+
+  return (hasFatal: hasFatal, messages: messages.toString());
+}
+
+({bool isFatal, String? message}) _validateVersion(
+  final Map<String, Dependency> dependencies,
+  final String packageName,
+  final Version minVersion,
+) {
+  final dependency = dependencies[packageName];
+  final maxVersion = minVersion.nextBreaking;
+
+  if (dependency == null) {
+    return (
+      isFatal: false,
+      message:
+          'Could not find the dependency on `$packageName` in pubspec.yaml file. Compatibility check is being skipped.',
+    );
+  } else if (dependency is HostedDependency) {
+    final constraint = dependency.version;
+    final invalidConstraintMessage =
+        'The version constraint $constraint on `$packageName` allows versions before $minVersion or after $maxVersion which is not allowed.';
+
+    if (constraint is Version && (constraint < minVersion || constraint > maxVersion)) {
+      return (isFatal: true, message: invalidConstraintMessage);
+    }
+
+    final range = constraint as VersionRange;
+    final rangeMin = range.min;
+    final rangeMax = range.max;
+
+    if (rangeMin == null || rangeMax == null || rangeMin < minVersion || rangeMax > maxVersion) {
+      return (isFatal: true, message: invalidConstraintMessage);
+    }
+  }
+
+  return (isFatal: false, message: null);
+}
--- a/packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
+++ b/packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
@@ -13,7 +13,7 @@ final dependencies = {
 };
 
 final devDependencies = {
-  'built_value_generator': Version.parse('8.8.0'),
+  'built_value_generator': Version.parse('8.8.1'),
 };
 
 /// Checks whether the correct version of the dependencies are present in the pubspec.yaml file.
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -12,6 +12,7 @@ import 'package:dynamite/src/builder/generate_schemas.dart';
 import 'package:dynamite/src/builder/imports.dart';
 import 'package:dynamite/src/builder/serializer.dart';
 import 'package:dynamite/src/builder/state.dart';
+import 'package:dynamite/src/helpers/version_checker.dart';
 import 'package:dynamite/src/models/config.dart';
 import 'package:dynamite/src/models/openapi.dart' as openapi;
 import 'package:version/version.dart';
@@ -38,6 +39,17 @@ class OpenAPIBuilder implements Builder {
 
   @override
   Future<void> build(final BuildStep buildStep) async {
+    final result = await helperVersionCheck(buildStep);
+
+    if (result.messages.isNotEmpty) {
+      if (result.hasFatal) {
+        log.severe(result.messages);
+        return;
+      } else {
+        log.info(result.messages);
+      }
+    }
+
     try {
       final inputId = buildStep.inputId;
       final outputId = inputId.changeExtension('.dart');
--- a/packages/dynamite/dynamite/README.md
+++ b/packages/dynamite/dynamite/README.md
@@ -16,7 +16,7 @@ dependencies:
   uri: ^1.0.0
 dev_dependencies:
   build_runner: ^2.4.0
-  built_value_generator: ^8.8.0
+  built_value_generator: ^8.8.1
 ```
 
 # Generating
--- a/packages/dynamite/dynamite/pubspec.yaml
+++ b/packages/dynamite/dynamite/pubspec.yaml
@@ -15,6 +15,8 @@ dependencies:
   intersperse: ^2.0.0
   meta: ^1.0.0
   path: ^1.0.0
+  pub_semver: ^2.1.4
+  pubspec_parse: ^1.2.3
   uri: ^1.0.0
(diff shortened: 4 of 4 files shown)


provokateurin (commented):
This should not only check dynamite_runtime but also every package that the generated code depends on (https://github.com/nextcloud/neon/blob/main/packages/dynamite/dynamite/lib/src/builder/imports.dart).

Leptopoda:
I don't know if this is a good idea.
both `dynamite_e2e` and `dynamite_petstore` do not import all of these as they don't depend on all of them.
And which versions should we require?

provokateurin:
I'd say if a package is missing from the pubspec then we can skip it since it might not be used as you said.
Regarding the required versions we could either test basically everything or just use the current major versions we depend on and then check the minimum minor and patch versions using pub downgrade and running tests.

Leptopoda:
Downgrading for the test should not be needed if we assume devs follow semver (and the packages we choose should come from reputable publishers like we already do).

While we (nextcloud package) might be affected by a bug/need a feature does not mean other users face the same.
An example is that a user might not use a `someOf` in their spec so in that case we should not force them to use `built_value:8.8.0`.

provokateurin:
I disagree, we should require dependencies that work for everyone, regardless of the features they use. Otherwise it makes no sense to check for dependency versions.
The downgrading was only meant as a way to figure out the lowest supported version. From our current dependency constraints it's impossible to know if we really support all the versions or depend on something that was added or fixed n a minor or patch release.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+  'uri': Version.parse('1.0.0'),
+};
+
+final devDependencies = {
+  'built_value_generator': Version.parse('8.8.0'),
+};
+
+/// Checks whether the correct version of `dynamite_runtime` is present in the
+/// pubspec.yaml file.
+Future<({bool hasFatal, String messages})> helperVersionCheck(final BuildStep buildStep) async {
```
provokateurin, line 21:
```suggestion
Future<({bool isFatal, String messages})> helperVersionCheck(final BuildStep buildStep) async {
```
For consistency

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+  'universal_io': Version.parse('2.0.0'),
+  'uri': Version.parse('1.0.0'),
+};
+
+final devDependencies = {
+  'built_value_generator': Version.parse('8.8.0'),
+};
+
+/// Checks whether the correct version of `dynamite_runtime` is present in the
+/// pubspec.yaml file.
```
provokateurin, line 20:
```suggestion
/// Checks whether the correct version of the dependencies are present in the pubspec.yaml file.
```

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+    final result = _validateVersion(pubspec.devDependencies, constraint.key, constraint.value);
+
+    if (result.message != null) {
+      messages.writeln(result.message);
+    }
+
+    if (result.isFatal) {
+      hasFatal = true;
+    }
+  }
```
provokateurin, line 59: Duplicate logic, should be a separate method

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+  final Version minVersion,
+) {
+  final dependency = dependencies[packageName];
+
+  if (dependency == null) {
+    return (
+      isFatal: false,
+      message: 'Could not find the required dependency on `$packageName` in pubspec.yaml file.'
+          ' Compatibility check is being skipped.',
+    );
```
provokateurin, line 76: Since not all dependencies are required for all specs, we might just want to ignore it. If the generated code contains a dependency that is not declared in the pubspec.yaml I don't consider it a problem we have to care about.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+  } else if (dependency is! HostedDependency) {
+    return (
+      isFatal: false,
+      message: 'The dependency on `$packageName` is not a `$HostedDependency`. '
+          'Compatibility check is being skipped.',
+    );
```
provokateurin, line 82: This could also be skipped as people should know what they are doing when they use a path or git dependency.

packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -38,6 +39,22 @@ class OpenAPIBuilder implements Builder {
...
+      final result = await helperVersionCheck(buildStep);
+
+      if (result.messages.isNotEmpty) {
+        log.warning(result.messages);
+      }
+
+      if (result.hasFatal) {
+        log.warning(result.messages);
+        return;
+      }
```
provokateurin, line 52: Shouldn't this throw an exception? Otherwise we also won't notice it in our CI.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,101 @@
...
+    if (constraint is Version && constraint < minVersion) {
+      return (isFatal: true, message: constraintTooLowMessage);
+    }
+
+    final range = constraint as VersionRange;
+    final rangeMin = range.min;
+
+    if (rangeMin == null || rangeMin < minVersion) {
+      return (isFatal: true, message: constraintTooLowMessage);
+    }
```
provokateurin, line 97: We also need to limit the upper version since a new major version can/will have breaking changes. The maximum exclusive version could be calculated on the fly as `${major+1}.0.0`.

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/openapi_builder.dart
Leptopoda: I thought returning was enough but you are right.
Thanks for catching

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
provokateurin: Let's keep it, but please remove the `required` part since it is not true for all.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
provokateurin: Less noise in the output to better highlight the actual problems.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
Leptopoda: I argue that this would make the code really ugly.

Leptopoda:
forgot to push yesterday :sweat_smile: 
Should be ready now.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+};
+
+/// Checks whether the correct version of the dependencies are present in the pubspec.yaml file.
+Future<({bool hasFatal, String messages})> helperVersionCheck(final BuildStep buildStep) async {
+  final pubspecAsset = AssetId(buildStep.inputId.package, 'pubspec.yaml');
+
+  if (!await buildStep.canRead(pubspecAsset)) {
+    throw Exception(
+      'Failed to read the pubspec.yaml file. '
+      'Version constraints of required packages can not be validated.',
```
provokateurin, line 26: Can you make these just one string? I think this is quite ugly (especially the trailing space)

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+  final String packageName,
+  final Version minVersion,
+) {
+  final dependency = dependencies[packageName];
+
+  if (dependency == null) {
+    return (
+      isFatal: false,
+      message: 'Could not find the dependency on `$packageName` in pubspec.yaml file.'
+          ' Compatibility check is being skipped.',
```
provokateurin, line 74: Same as above and even worse readability imo

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+    );
+  } else if (dependency is HostedDependency) {
+    final constraint = dependency.version;
+    final constraintTooLowMessage = 'The version constraint "$constraint" on `$packageName` allows '
+        'versions before $minVersion which is not allowed.';
```
provokateurin, line 79: And again

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+      return (isFatal: true, message: constraintTooLowMessage);
+    }
+
+    final range = constraint as VersionRange;
+    final rangeMin = range.min;
+    final rangeMax = range.max;
+
+    if (rangeMin == null || rangeMax == null || rangeMin < minVersion || rangeMax > minVersion.nextBreaking) {
+      return (isFatal: true, message: constraintTooLowMessage);
+    }
```
provokateurin, line 91: The max check needs a different error message.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+          ' Compatibility check is being skipped.',
+    );
+  } else if (dependency is HostedDependency) {
+    final constraint = dependency.version;
+    final constraintTooLowMessage = 'The version constraint "$constraint" on `$packageName` allows '
+        'versions before $minVersion which is not allowed.';
+
+    if (constraint is Version && constraint < minVersion) {
+      return (isFatal: true, message: constraintTooLowMessage);
+    }
```
provokateurin, line 83: This also needs a check for the max version, right?

provokateurin (changes requested):

provokateurin:
Please adjust the built_value_generator minimum to 8.8.1 (also in the dynamite docs): https://github.com/nextcloud/neon/pull/1274.

packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
```diff
@@ -0,0 +1,95 @@
...
+  if (dependency == null) {
+    return (
+      isFatal: false,
+      message:
+          'Could not find the dependency on `$packageName` in pubspec.yaml file. Compatibility check is being skipped.',
+    );
+  } else if (dependency is HostedDependency) {
+    final constraint = dependency.version;
+    final invalidConstraintMessage =
+        'The version constraint "$constraint" on `$packageName` allows versions before $minVersion or after $maxVersion which is not allowed.';
```
provokateurin, line 79:
```suggestion
        'The version constraint $constraint on `$packageName` allows versions before $minVersion or after $maxVersion which is not allowed.';
```

packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -38,6 +39,21 @@ class OpenAPIBuilder implements Builder {
...
+        log.info(result.messages);
+      }
+
+      if (result.hasFatal) {
+        log.warning(result.messages);
+        throw Error();
+      }
+    } on Exception catch (exception) {
+      log.warning(exception);
+    }
```
provokateurin, line 55:
```suggestion
    final result = await helperVersionCheck(buildStep);

    if (result.messages.isNotEmpty) {
      if (result.hasFatal) {
        log.severe(result.messages);
        return;
      } else {
        log.info(result.messages);
      }
    }
```
Removes duplicate messages, `Instance of 'Error''` message and early exit if the constraints are wrong.

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -38,6 +39,21 @@ class OpenAPIBuilder implements Builder {
...
+        if (result.hasFatal) {
+          log.severe(result.messages);
+          return;
+        } else {
+          log.info(result.messages);
+        }
+      }
+    } on Exception catch (exception) {
+      log.warning(exception);
+    }
```
provokateurin, line 55: I don't get why you'd want to catch these. If there is an exception thrown then the builder should fail, no?

provokateurin (commented):
LGTM, just one question

packages/dynamite/dynamite/lib/src/openapi_builder.dart
Leptopoda: The exception is only thrown when we can't read the pubspec in which case we just skip the verification and assume the deps are ok.

provokateurin: Then please just log the warning there instead of throwing an exception and handling it here.

provokateurin (approved):

PR #1239, opened 2023-12-06
chore(deps): Upgrade to gradle 8

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1004

### DIFF
--- a/packages/app/android/gradle.properties
+++ b/packages/app/android/gradle.properties
@@ -1,3 +1,4 @@
 org.gradle.jvmargs=-Xmx4G
 android.useAndroidX=true
 android.enableJetifier=true
+kotlin.jvm.target.validation.mode = IGNORE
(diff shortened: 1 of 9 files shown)


packages/app/android/gradle.properties
```diff
@@ -1,3 +1,4 @@
-org.gradle.jvmargs=-Xmx1536M
+org.gradle.jvmargs=-Xmx4G
 android.useAndroidX=true
 android.enableJetifier=true
+kotlin.jvm.target.validation.mode = IGNORE
```
Leptopoda, line 4: Can you please explain this to me?
Looks like it is needed when the java and Kotlin jvm version mismatch? Can't we avoid this by specifying the correct versions?

provokateurin: I tried specifying the correct versions but I think this problem is not in our code but in the libraries that are included. I couldn't find a way to make it work other than with this line.

Leptopoda (approved):
I don't like the ignore but other than that LGTM

PR #1242, opened 2023-12-07
fix(dynamite): Allow int enums

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1240
Depends on https://github.com/google/built_value.dart/pull/1284


provokateurin:
Depends on https://github.com/nextcloud/neon/pull/1243.
I need to test it with the new specs as I'm not sure if this will work correctly now.

provokateurin:
Oh well it generates broken code right now

provokateurin:
Ah I had already fixed the problem but it wasn't committed to this branch.
Please review now, it should be all done now.

packages/dynamite/dynamite_end_to_end_test/test/enum_test.dart
```diff
@@ -22,4 +25,22 @@ void main() {
...
     });
   });
+
+  group('int', () {
+    test('serialize', () {
+      final serialized = serializers.serialize(
+        EnumInt.$2,
+        specifiedType: const FullType(EnumInt),
+      );
+      expect(serialized, '2');
```
Leptopoda, line 35: why do we expect a string value and not a plain `2`?

packages/dynamite/dynamite/lib/src/helpers/type_result.dart
```diff
@@ -1,8 +1,8 @@
 import 'package:dynamite/src/helpers/dart_helpers.dart';
 import 'package:dynamite/src/models/type_result.dart';
 
-String valueToEscapedValue(final TypeResult result, final String value) {
-  if (result is TypeResultBase && result.name == 'String') {
+String valueToEscapedValue(final TypeResult result, final String value, {final bool forceString = false}) {
```
Leptopoda, line 4: Please add a doc comment where and why `forceString` is used.

Leptopoda (changes requested):

packages/dynamite/dynamite_end_to_end_test/test/enum_test.dart
provokateurin: EnumClass can only take Strings as the name and this means it can only be serialized as such. We need to chain it with a `String->T where Enum<T>` serializer :thinking: Since I'm a noob, can you tell me how to do that? :see_no_evil:

provokateurin: Or we change EnumClass to accept any type which I would prefer as we won't need the workaround for enums that are not strings.

provokateurin: But that would be a breaking change for built_value so might not get accepted so easily.

Leptopoda: I'm curious whether this is even intended for openapi :thinking: 
And you already guessed that breaking changes in built_value are really hard to land upstream as this means all google internal projects must adapt and this can be a really slow process.

Leptopoda (approved):
LGTM
dunno what to do with the remaining issue

packages/dynamite/dynamite_end_to_end_test/test/enum_test.dart
provokateurin: You mean if integer enums are intended?

Leptopoda: Yes

provokateurin: Yes the are. I previously also thought they were not allowed, but they are.

provokateurin: https://json-schema.org/understanding-json-schema/reference/enum

provokateurin: I think this is still a problem? I'll add a test that checks that the enum is correctly serialized in json but I'm pretty sure it will fail.

provokateurin:
This is still broken because it serializes to string which is not correct.
Marking as draft to prevent accidental merging.

provokateurin:
@Leptopoda could you take this over?

Leptopoda:
I have an idea and will take a look (hopefully tomorrow)

Leptopoda:
@provokateurin can we also delete the branch?

PR #1243, opened 2023-12-07
feat(dynamite_end_to_end_test): Add string enum with a keyword value

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1241
Depends on https://github.com/google/built_value.dart/pull/1284


Leptopoda (changes requested):
Enums can get fairly complex (all the different subschemas).
Should we add these schemas to a separate spec (thinking of "enum.openapi.dart").

Please also unit test the generated code.

Leptopoda (approved):

PR #1248, opened 2023-12-07
feat(neon_framework): Add Findable interface and extension

provokateurin:
Closes https://github.com/nextcloud/neon/issues/989


packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
```diff
@@ -109,7 +108,7 @@ class DashboardBloc extends InteractiveBloc implements DashboardBlocEvents, Dash
             Result.success(
               widgets.map(
                 (final id, final items) => MapEntry(
-                  response.body.ocs.data.values.find(id),
+                  response.body.ocs.data.values.singleWhere((final widget) => widget.id == id),
```
Leptopoda, line 111: Is it really important that we also verify that only a single value matches this?
To do so single has to always iterate over every element while first will stop on the first.

```suggestion
                  response.body.ocs.data.values.firstWhere((final widget) => widget.id == id),
```

packages/neon_framework/lib/src/utils/findable.dart
Leptopoda, line 1: Please unit test this directly.

packages/neon_framework/lib/src/utils/findable.dart
```diff
@@ -0,0 +1,20 @@
...
+abstract interface class Findable {
+  /// Unique ID of the element.
+  String get id;
+}
+
+extension FindExtension<T extends Findable> on Iterable<T> {
+  /// Returns the first element matching [id] by [Findable.id].
+  ///
+  /// If no element was found `null` is returned.
+  T? tryFind(final String? id) => firstWhereOrNull((final e) => e.id == id);
```
Leptopoda, line 13: I'd rather not support nullable parameters but we already use them throughout the code so let's early abort.

```suggestion
  T? tryFind(final String? id) {
	if(id == null){
		return null;
	}
	
        return firstWhereOrNull((final e) => e.id == id);
  }
```

Leptopoda (approved):
Other than the small change LGTM

PR #1256, opened 2023-12-08
fix(dynamite): Fix nullable someOf parameters

provokateurin:


packages/dynamite/dynamite_end_to_end_test/lib/some_of.openapi.dart
```diff
@@ -29,6 +33,79 @@ class Client extends DynamiteClient {
...
+    final queryParameters = <String, dynamic>{};
+    final headers = <String, String>{};
+    Uint8List? body;
+
+    if (oneOf != null) {
+      queryParameters['oneOf'] = oneOf.toString();
+    }
+    if (anyOf != null) {
+      queryParameters['anyOf'] = anyOf.toString();
+    }
```
provokateurin, line 90: I assume this should call serialize instead, but I'm not sure if that would be enough? @Leptopoda could you have a look at this and fix it (in a separate PR)?

provokateurin: Do you prefer merging this already (with the broken serialization) or fixing it in this PR as well?

Leptopoda: If we need it for 28 I'll review it and we can merge.

If not I'd rather add unit tests with a mocked HttpClient first.
I'll open an issue for that.

Leptopoda: I'll do both. Should be straightforward.

Leptopoda:
I think that this is a bigger issue:
https://github.com/nextcloud/neon/blob/027a37d70d964f0659f545e65fc7ef07cd38313f/packages/dynamite/dynamite/lib/src/builder/client.dart#L247-L256

Calling `dartType` here caused the nullability issue in the first place and as the returned type-result is a `TypeResultBase` the serialization is also broken.

Leptopoda:
@provokateurin what do you think about this:
[patch.txt](https://github.com/nextcloud/neon/files/13635911/patch.txt)

provokateurin:
The patch is empty, but LGTM :joy:

Leptopoda:
Rip, I had already staged the changes so git diff was empty :)
[patch.txt](https://github.com/nextcloud/neon/files/13635988/patch.txt)

provokateurin:
The diff looks good, but I don't like the introduction of the mime type. It is wrong because these parameters do not actually have a mime type. I don't know a better way though so we have to move forward that way. Your the PR with your patch will replace this PR, right?

Leptopoda:
branch is still floating around :)

PR #1259, opened 2023-12-09
nextcloud: Add Content-Length headers in WebDavClient.put

cuihaoleo:
Missing Content-Length causes Nextcloud to write empty files.

Please see the discussion here:
https://github.com/saber-notes/saber/issues/945#issuecomment-1848069199

### DIFF
--- a/packages/nextcloud/lib/src/webdav/client.dart
+++ b/packages/nextcloud/lib/src/webdav/client.dart
@@ -134,7 +134,7 @@ class WebDavClient {
         headers: _getUploadHeaders(
           lastModified: lastModified,
           created: created,
-          contentLength: null,
+          contentLength: localData.length,
         ),
       );
 


cuihaoleo:
FYI, now I have a different fix on my nginx instance: https://github.com/saber-notes/saber/issues/945#issuecomment-1848100037

And I'm not sure if adding Content-Length is the best solution. But apparently chunked encoding is troublesome.

provokateurin (approved):
Thanks for the fix. Interesting that we didn't run into any problems yet and that it only breaks in certain scenarios.

provokateurin:
The commit message should be `fix(nextcloud): Add Content-Length header in WebDavClient.put` (otherwise the CI won't pass).

The other CI failure is a problem that I didn't think about. @Leptopoda we need to make it possible to run the CI for people who are not part of the org :sweat_smile:

provokateurin:
All tests pass locally, so we can merge after the commit message is fixed.

cuihaoleo:
The commit message is fixed.

PR #1265, opened 2023-12-09
test(dynamite_end_to_end_test): test the clients generated by dynamite

Leptopoda:
I usually prefer `mocktail` instead of `mockito` but after a lot of back and forth this was the first one I got working without too much work.

This infrastructure is needed to test the issue discovered in #1256. Testing the other remaining e2e specs will come when I have more time (or someone else wants to tackle it)


provokateurin (approved):
The tests themselves look pretty useless for now, but I get that you need the mocks to verify the correct behavior of our generated code.
I think mocktail is still better because it doesn't generate 2k new lines, but that's ok for now. We can just migrate it later.

PR #1267, opened 2023-12-10
feat(neon_framework): Allow dev servers in version checks

provokateurin:
Useful when testing/developing directly against a git version of Nextcloud.


Leptopoda (approved):
Didn't test it but LGTM

Issue #1247, opened 2023-12-07
Dynamite: validate more types


Leptopoda:
Dynamite should support validating more strict types:
- array lengths ##1056 
	- minItems
	- maxItems 
	- uniqueItems (maybe utilize a Set for this)
- number:
	- minimum/exclusiveMinimum
	- maximum/exclusiveMaximum
	- multipleOf
- enum
	- check the pattern at build time (#1313)

This can all fit nicely into: https://github.com/nextcloud/neon/blob/main/packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart. We must not only consider validating these schemas inside objects but also when uses as a parameter in an operation.

provokateurin:
String enums with patterns are broken:
```json
{
	"type": "string",
	"pattern": "^a$",
	"enum": [
		"a",
		"b"
	]
}
```
The pattern match method is called but the enum it is given is not the expected String.

provokateurin:
Required keys of objects are not checked to be present in the properties keys.

Issue #1097, opened 2023-11-03
Use UriTemplates for the Uris


Leptopoda:
We currently query encode path parameters which is not required by the openapi specification. Only query parameters need to be:

```dart
    final x0 = Uri.encodeQueryComponent(x.toString());
    final y0 = Uri.encodeQueryComponent(y.toString());
	[...]
    final file0 = Uri.encodeQueryComponent(file);
    final path = '/index.php/apps/files/api/v1/thumbnail/$x0/$y0/$file0';
    final uri = Uri(path: path, queryParameters: queryParameters.isNotEmpty ? queryParameters : null);
```

https://swagger.io/docs/specification/serialization/ has a table on how to interpret parameter specifications as an RFC 6570 uri template. The google published dart library `uri` already supports these templates (https://pub.dev/documentation/uri/latest/uri/UriTemplate-class.html) so switching to them might:
- widen the supported parameter serialization types
- give us the confidence that we do it correctly
- maybe reduce complexity of our code as we'd only have a template and a collection of values to substitude

@provokateurin what do you think? Shouldn't be that hard.

provokateurin:
Oh that sounds very good, I assume it will also already handle the URL encoding for us?

Leptopoda:
I think this is part of the RFC specification.
Let's see

provokateurin:
@Leptopoda May I take this issue over? You already have so much dynamite work on your plate

Leptopoda:
sure

provokateurin:
Later on we should also use the UriTemplates for query parameters.

provokateurin:
Unassigning myself for now

Leptopoda:
removing from the milestone as we now no longer incorrectly encode path parameters.

Leptopoda:
the pr didn't auto close this issue :(
closing as done.

PR #1272, opened 2023-12-11
fix(dynamite): Fix nullable someOf parameters

Leptopoda:
#1256

### DIFF
--- a/packages/dynamite/dynamite_end_to_end_test/test/parameters_test.dart
+++ b/packages/dynamite/dynamite_end_to_end_test/test/parameters_test.dart
@@ -118,6 +118,62 @@ void main() {
         ]),
       );
     });
+    test('oneOf', () async {
+      await client.$get(oneOf: ($bool: true, string: null));
+      var captured = verify(mockHttpClient.openUrl(captureAny, captureAny)).captured;
+      expect(
+        captured,
+        equals([
+          'get',
+          Uri.parse('example.com/?oneOf=true'),
+        ]),
+      );
+      resetMockitoState();
+
+      await client.$get(oneOf: ($bool: null, string: 'value'));
+      captured = verify(mockHttpClient.openUrl(captureAny, captureAny)).captured;
+      expect(
+        captured,
+        equals([
+          'get',
+          Uri.parse('example.com/?oneOf=value'),
+        ]),
+      );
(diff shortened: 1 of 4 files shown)


packages/dynamite/dynamite_end_to_end_test/test/parameters_test.dart
```diff
@@ -121,4 +121,30 @@ void main() {
...
+      await client.getSomeOf(anyOf: ($bool: true, string: null));
+      final captured = verify(mockHttpClient.openUrl(captureAny, captureAny)).captured;
+      expect(
+        captured,
+        equals([
+          'get',
+          Uri.parse('example.com/some-of?anyOf=true'),
+        ]),
+      );
+    });
```
provokateurin, line 148: Please also test that it works with a string.

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -253,7 +253,11 @@ Iterable<Method> buildTags(
           ),
           parameter.schema!,
           nullable: dartParameterNullable,
-        ).dartType;
+        );
+
+        if (result is! TypeResultSomeOf) {
+          result = result.dartType;
+        }
```
provokateurin, line 260: Please add a comment that explains why this is needed.

Leptopoda:
I think that it's probably easier to just call serialize (and not encode) on the typeresult and let the uri template handle the rest #1097.
I'll give it a try.

provokateurin:
That sounds good yeah

provokateurin:
yet another problem related to parameter serialization: pattern checking does not work on string values. I would argue it is useless because it is an enum already, but it is a valid spec and also happens to appear in Talk.

Leptopoda:
Pattern checking requires a bigger makeover.
Can you report an example spec to #1247?
Or open a new issue and mention it from there.

provokateurin:
I commented on the issue, but I feel like the problem is just the parameter serialization. The value needs to be serialized before checking the pattern (would be at least one way to fix it).

Leptopoda:
Still bases on #1292

provokateurin (approved):

Issue #1261, opened 2023-12-09
Check WebDAV put headers in tests


provokateurin:
https://github.com/nextcloud/neon/pull/1259

To avoid future regressions.

PR #1273, opened 2023-12-11
feat(nextcloud_test): Add fixture recording and validation

provokateurin:
This system records HTTP requests and compares them to some saved pattern. This ensures we never have any changes in the raw requests that we don't know about.

Ideally we create a dynamite_test package so that we can reuse this system for dynamite_petstore_example and dynamite_end_to_end_test. I can do that in this PR already if it is favorable.

There is no performance impact to the tests:
```log
before: 08:04 +387 ~2: All tests passed!
after: 08:05 +387 ~2: All tests passed!
```

https://github.com/nextcloud/neon/issues/435


packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
```diff
@@ -11,9 +11,11 @@ import 'package:xml/xml_events.dart';
 /// Usually a `Stream<Uint8List>`.
 typedef BytesStream = Stream<List<int>>;
 
-final _jsonBytesConverter = utf8.decoder.fuse(json.decoder);
+/// Converter for UTF-8 encoded bytes to JSON.
+final jsonBytesConverter = utf8.decoder.fuse(json.decoder);
 
-final _xmlBytesConverter =
+/// Converter for UTF-8 encoded bytes to XML.
+final xmlBytesConverter =
```
Leptopoda, line 18: Can we mark them as visible for testing?

packages/nextcloud_test/lib/src/proxy_http_client.dart
```diff
@@ -0,0 +1,123 @@
...
+
+    for (var value in values) {
+      if (name == HttpHeaders.authorizationHeader) {
+        value = '${value.split(' ').first} mock';
+      }
+      if (name == 'destination') {
+        value = Uri.parse(value).replace(port: 80).toString();
+      }
+
+      buffer.write('\n$name: $value');
```
Leptopoda, line 110: writeln

packages/nextcloud_test/lib/src/proxy_http_client.dart
```diff
@@ -0,0 +1,123 @@
...
+    }
+  });
+
+  if (body.isNotEmpty) {
+    try {
+      buffer.write('\n${utf8.decode(body)}');
+    } catch (_) {
+      buffer.write('\n${base64.encode(body)}');
```
Leptopoda, line 118: again writeln

packages/nextcloud_test/lib/src/proxy_http_client.dart
```diff
@@ -0,0 +1,123 @@
...
+    });
+
+    return mockRequest;
+  });
+
+  return mockClient;
+}
+
+String _formatHttpRequest(final HttpClientRequest request, final Uint8List body) {
+  final buffer = StringBuffer('${request.method} ${request.uri.replace(port: 80)}');
```
Leptopoda, line 95: in order to work with the other writeln changes this would need to be adapted to also call writeln in a cascade way

packages/nextcloud/test/user_status_test.dart
```diff
@@ -19,6 +19,7 @@ void main() {
         if (Invoker.current!.liveTest.errors.isNotEmpty) {
           print(await container.allLogs());
         }
+        validateFixture(preset);
```
Leptopoda, line 22: Every test now calls validateFixtures.
can't this be done directly in the nextcloud_test package (inside the `presets` function)?

Leptopoda (commented):
looks good from a quick glance

packages/nextcloud/test/user_status_test.dart
provokateurin: Ah true, this can easily be added in a second teardown. With printing the log on failure it was not possible because the container needed to be still around at that time.

packages/nextcloud_test/lib/src/proxy_http_client.dart
provokateurin: This is done so we don't end with a trailing newline.

Leptopoda (approved):
LGTM 
I tjink #1261 can also be marked to be closed.

provokateurin:
Sure. Do you want me to create a dynamite_test package out of this so we can also utilize it for petstore and the e2e tests?

Leptopoda:
I think that these packages can utilise nextxloud_test directly. I don't see the benefit of a separate package.

provokateurin:
The nextcloud_test package has a lot of nextcloud specific stuff so it's not very clean to add it to the dynamite packages which are not nextcloud specific.

Leptopoda:
But it's only for testing 🤷‍♀️
If you prefer a separate package I'm fine with it.

provokateurin:
Let's keep it as is for now but once we want this in other packages as well we should move it to a separate package.

PR #1276, opened 2023-12-12
feat(dynamite): Expose serializers for testing

provokateurin:
Makes it possible to directly use the serializers in tests if needed. Fun diff to review :)


Leptopoda (approved):
Changes to dynamite look good and nothing too obvious in the generated code.

Issue #1172, opened 2023-11-22
Remove unused parameter maps


provokateurin:
Currently we generate maps for query, path and header parameters even if the operations don't use any of those parameter types. We should not generate those maps at all if they are not used.

Leptopoda:
the pr didn't auto close this issue :(
there are still minor changes that can be made but overall this I consider this done for now

PR #1278, opened 2023-12-12
fix(dynamite): object serialization in query parameters

Leptopoda:
fixes: #1097 
fixes: #1172 
And more :)


packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
```diff
@@ -84,6 +86,13 @@ final _dartKeywords = [
   'with',
 ];
 
+const _dartTypes = [
```
provokateurin, line 89: Why not put them to the other keywords? I don't see any special handling because they are types :woman_shrugging:

packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
```diff
@@ -9,7 +9,7 @@
             "get": {
                 "parameters": [
                     {
-                        "name": "content-string",
+                        "name": "content_string",
```
provokateurin, line 12: This change doesn't seem necessary?

packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
```diff
@@ -21,7 +21,7 @@
                         }
                     },
                     {
-                        "name": "content-parameter",
+                        "name": "content_parameter",
```
provokateurin, line 24: Same as above

packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
```diff
@@ -31,6 +31,57 @@
...
+                    {
+                        "name": "string_binary",
+                        "in": "query",
+                        "schema": {
+                            "type": "string",
+                            "format": "binary"
+                        }
+                    },
+                    {
+                        "name": "integer",
```
provokateurin, line 65:
```suggestion
                        "name": "int",
```
To properly check for collision with `int`

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -54,9 +59,111 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
...
+    final prefix = switch (style) {
+      ParameterStyle.simple => null,
+      ParameterStyle.label => '.',
+      ParameterStyle.matrix => ';',
+      ParameterStyle.form => isFirst ? '?' : '&',
+      ParameterStyle.spaceDelimited || ParameterStyle.pipeDelimited => explode
+          ? isFirst
+              ? '?'
+              : '&'
+          : throw UnsupportedError(error),
```
provokateurin, line 81: Very ugly formatting

provokateurin (changes requested):
LGTM, I think for this case alone it was nice to have the fixtures because we can be sure there were no changes. Only a small failure in the CI due to a missing serializer but I guess that can be fixed easily.

packages/dynamite/dynamite_end_to_end_test/lib/parameters.openapi.json
Leptopoda: variable names can only consist of `ALPHA / DIGIT / "_" / pct-encoded`.
https://datatracker.ietf.org/doc/html/rfc6570#section-2.3

As `-` is valid for query parameters but not for UriTemplates. I don't know how to support them.
At least for path parameters openapi says:
> The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by RFC3986: forward slashes (/), question marks (?), or hashes (#).

provokateurin: Hm I see

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -317,12 +316,30 @@ Iterable<Method> buildTags(
...
+          pathBuilder
+            ..write('{?')
+            ..writeAll(queryParams, ',')
+            ..write('}');
+        }
+
+        final path = pathBuilder.toString();
+
+        // Sanity check the uri at build time.
+        UriTemplate(path);
```
provokateurin, line 340: Exceptions should be caught and a better error message should be logged instead.

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -54,9 +59,111 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
     return null;
   }
 
+  /// Builds the uri template value for this parameter.
+  ///
+  /// Throws a StateError if the parameter does not support a uri template.
```
provokateurin, line 64:
```suggestion
  /// Throws a `StateError` if the parameter does not support a uri template.
```

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -54,9 +59,111 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
...
+  String uriTemplate({
+    final bool isFirst = false,
+    final bool withPrefix = true,
```
provokateurin, line 67: Add doc comments that explain what they do.

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -54,9 +59,111 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
...
+  }) {
+    final error = 'The parameter $name is not supported by RFC 6570 UriTemplates. Dynamite can not yet work with it.';
```
provokateurin, line 69:
```suggestion
    final errorUnsupportedParameter = 'The parameter $name is not supported by RFC 6570 UriTemplates. Dynamite can not yet work with it.';
```

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -54,9 +59,111 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
...
+
+    if (explode) {
+      buffer.write('*');
+    }
+
+    return buffer.toString();
+  }
+
+  /// The pct encoded name of this parameter.
+  String get pctName => Uri.encodeQueryComponent(name);
```
provokateurin, line 103:
```suggestion
  String get percentEncodedName => Uri.encodeQueryComponent(name);
```
It's better to use full names instead of abbreviations

provokateurin (changes requested):
I'll give it another review once rebased and the CI is fixed.

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: I agree but how do we easily quit generation?
Just rethrow?

provokateurin: Throw a new exception with the improved error message (and link to the rfc would be good as well).

packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
Leptopoda: I disagree. The RFC6570 only calls them `pct-encoded`

provokateurin: Then `pctEncodedName` please

provokateurin (approved):

Issue #1281, opened 2023-12-13
send default parameters when explicitly specified.


Leptopoda:
PR is coming :)

PR #1292, opened 2023-12-13
feat(dynamite): always send parameters even when being the default

Leptopoda:
fixes: #1281


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -263,16 +259,12 @@ Iterable<Method> buildTags(
                 ..name = toDartName(parameter.name)
                 ..required = parameterRequired
                 ..type = refer(result.nullableName);
-
-              if (parameter.schema!.$default != null) {
-                b.defaultTo = Code(valueToEscapedValue(result, parameter.schema!.$default!.toString()));
-              }
```
provokateurin, line 269: This needs to stay to indicate the default value as it will be used.

provokateurin (changes requested):
The change to the parameter doesn't make sense. Only the check if the value matches the default value has to be removed.

provokateurin (changes requested):
Looking at the fixtures it seems like empty values are added?

Also the fixtures need to keep the regexes otherwise they don't work.

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: This is fundamentally broken.
If we want to support arbitrary objects to have default values this does not work.

The UriTemplate pr also needs the parameters to be built_collections as the serializer can't work with plain Lists or Maps.

It'll make semse once I push it tomorrow.

Leptopoda:
> Looking at the fixtures it seems like empty values are added?
> 
> Also the fixtures need to keep the regexes otherwise they don't work.

I'll double check. Maybe I need to do everything at once in a big PR. UriTemplate does not do so so probably some issue got introduced when rebasing.

Leptopoda:
I think you where referring to the parameters that default to an empty string.


I'll fix the fixtures (re add the reggexes that got removed for the parameters) and try to figure out why the tests are failing.

.vscode/settings.json
```diff
@@ -13,4 +13,4 @@
   "files.associations": {
     "*.arb": "json",
   }
-}
+}
```
provokateurin, line 16: Unnecessary change.

provokateurin (approved):
LGTM, but it would have been easier if you explained from the start why the default values can no longer be present in the method header. Just from this diff alone it doesn't make sense so it is better to explain why you did it.

provokateurin:
I find it still confusing that there is no equals sign when the value is an empty string but if that is correct I don't care :woman_shrugging:

Leptopoda:
I'll take a look. Totally forgot to verify that.
Thanks for mentioning

Leptopoda:
Ok it is wrong
https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.9
>        {&x,y,empty}       &x=1024&y=768&empty=

it will get even funnier as we must also add the `=`.
I don't think this is possible in the current way. Should I close this PR and just have both commits in the other one?

provokateurin:
Yes I think we need to have both changes at once. I think you could change the target branch of the other PR to this branch and then we merge the other PR into this one and merge this PR in the end

provokateurin (approved):
Please rebase on the main branch to ensure everything is working correctly.

Leptopoda:
This is odd :thinking: 
on 26 we now get 0 Items back for the dashboard. Maybe an upstram bug @provokateurin?

provokateurin:
This is really strange indeed. I can reproduce it locally :thinking:

provokateurin:
Also fails on the main branch. Switching to 27.0.1 works

provokateurin:
I rebuild the image without any cache locally and it works again...

provokateurin:
I deleted the tagged version of the image and restarted the CI and now it works :woman_shrugging:

PR #1313, opened 2023-12-18
fix(dynamite): Fix pattern check for enums

provokateurin:
The pattern has to be checked after the value has been serialized (not only applies to enums, but that's the only use case so far).

### DIFF
--- a/packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart
+++ b/packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart
@@ -4,18 +4,17 @@ import 'package:dynamite/src/models/openapi/schema.dart';
 Iterable<String> buildPatternCheck(
   final openapi.Schema schema,
   final String value,
+  final String name,
 ) sync* {
-  final name = "'$value'";
-
   if (schema.type == SchemaType.string) {
     if (schema.pattern != null) {
-      yield "dynamite_utils.checkPattern($value, RegExp(r'${schema.pattern!}'), $name);";
+      yield "dynamite_utils.checkPattern($value as String?, RegExp(r'${schema.pattern!}'), '$name');";
     }
     if (schema.minLength != null) {
-      yield 'dynamite_utils.checkMinLength($value, ${schema.minLength}, $name);';
+      yield "dynamite_utils.checkMinLength($value as String?, ${schema.minLength}, '$name');";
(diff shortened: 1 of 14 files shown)


Leptopoda (changes requested):
Please add a unit test to dynamite_e2e.
I'd suggest a new spec just for pattern checking.

For enums we should be able to verify the pattern at generation time as the value of an enum is constant. Should we do so long term? If so we should open an issue for that.

provokateurin:
Yes, I forgot to add the test before creating the PR :sweat_smile:
Checking enums at build time sounds good, for other cases we still need this change. Can you open an issue for that?

Leptopoda (approved):
I think that it'll fit better into:
https://github.com/nextcloud/neon/blob/main/packages/dynamite/dynamite_end_to_end_test/lib/pattern_check.openapi.json

Other than that LGTM

provokateurin:
Right but in this case I don't want to test the pattern checking itself but that it works with serialized parameters (which was broken before).
Also I made a mistake which I'm fixing right now.

packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart
Leptopoda, line 1: Why did you change the return type?

We need to return multiple values as a parameter can have a length restriction and pattern at the same time. I think our pattern_check_test in e2e should catch that and error out.

Also why do the null check in here when dynamite_runtime already does so? just don't cast to `String` and cast to `String?`.

Leptopoda (changes requested):
blocking for now

packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart
Leptopoda: Ok indeed the test caught it :sweat_smile:

provokateurin: Oops, more oversights :sweat_smile: I only did this dirty on the branch for the new specs without giving too much thought

provokateurin:
A null input for the pattern checks is always allowed. This was not handled correctly and the tests for it were also wrong. If the input is not allowed to be null then it needs to be checked elsewhere and not with the patterns.

Leptopoda (approved):

PR #1315, opened 2023-12-18
refactor(nextcloud_test): Remove version handling mess

provokateurin:
Instead of manually keeping track of if the minor and patch versions are set we can just increment the major version and make it exclusive.


Leptopoda (approved):
Thanks for the cleanup :)
I didn't test it but code looks good

PR #1317, opened 2023-12-19
refactor(neon_framework): Cleanup settings and options related class names

provokateurin:
Needed for https://github.com/nextcloud/neon/pull/600
The `Specific` and `App` parts were never needed in the class names. Also more aligned with the rest, using `AppImplementation` instead of `NextcloudApp`.


Leptopoda (commented):
I'm hesitant to approving this as we wanted to call them `client` instead of app but when I last tried changing them this conflicted with the api clients we have.

What do you think?

provokateurin:
I still prefer AppImplementation over Client, but either way we should make the names consistent. If we change to Client later then we can do it, but these names are not aligned so that needs to be fixed first.

Leptopoda (approved):
ok let's do it

PR #1319, opened 2023-12-19
fix(dynamite): Allow int enums

### LINKED ISSUE #1240: Integer enums are broken
A schema like
```json
"enum-int": {
    "type": "integer",
    "enum": [
        0,
        1,
        2
    ]
}
```
does not work because we expect the possible enum values to be only strings.

Leptopoda:
fixes #1240 
replaces: #1242


packages/dynamite/dynamite/lib/src/models/type_result/enum.dart
```diff
@@ -20,7 +20,7 @@ class TypeResultEnum extends TypeResult {
     final bool onlyChildren = false,
     final String? mimeType,
   }) {
-    if (subType.name == 'String') {
+    if (subType.name == 'String' || subType.name == 'int') {
       return '$object.name';
     }
```
provokateurin, line 25: Shouldn't this return the name in all cases regardless of the subtype? I guess that was left over from the changes I did

provokateurin (changes requested):
Great job!

packages/dynamite/dynamite/lib/src/models/type_result/enum.dart
Leptopoda: I'd rather merge #1323 as it will remove this mess altogether :)

provokateurin: Ok let's get that merged first then

Leptopoda:
I'm not sure this is the best way to do it.
I think that enum_value.name should always return a `String`.
What about introducing a separate attribute called `value` that is of type `T`?

provokateurin:
I like the idea, please go ahead

provokateurin:
I rebased the branch and confirmed it works with the latest specs. If you want I can push that branch if you haven't done it locally already.

Leptopoda:
I've already rebased and done this:
>  What about introducing a separate attribute called value that is of type T?

I'll push it in a few minutes just want to make sure it can be easily adapted without any breaking changes.

Leptopoda:
I left it on a separate commit for now so we can roll it back easier.
We can squash them before merging.

provokateurin:
I'm confused what you did in the second commit. It looks like you reverted a lot of the changes from the first commit?

Leptopoda:
You are right.
In the second commit we just have the value getter and zan use the built_value_generator for the rest of the enum so no need to do it ourselves.

packages/dynamite/dynamite_end_to_end_test/test/enum_test.dart
```diff
@@ -3,23 +3,96 @@ import 'package:dynamite_end_to_end_test/enum.openapi.dart';
...
+          result.value,
+          specifiedType: const FullType(EnumDynamic),
+        );
+        expect(deserialized, result.key);
+      }
+    });
+  });
+
+  group('wrapped', () {
+    test('serialize', () {
```
provokateurin, line 82: Deserialize test for the wrapped would be good too.

provokateurin (approved):

Issue #1255, opened 2023-12-08
Skip generating client class if no endpoints are defined


provokateurin:
https://github.com/nextcloud/neon/blob/main/packages/nextcloud/lib/src/api/comments.openapi.dart has a client class but no endpoint. The client therefore is unnecessary and can just be omitted in the output.

PR #1324, opened 2023-12-19
refactor(dynamite): do not generate _parameters map when not needed

Leptopoda:
Fixes: #1255

And various other similar houskeepings.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/client.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/client.dart
@@ -189,6 +189,9 @@ Iterable<Method> buildTags(
         ...?pathEntry.value.parameters,
         ...?operation.parameters,
       ]..sort(sortRequiredParameters);
+      final hasUriParameters = parameters
+          .where((final p) => p.$in == openapi.ParameterType.path || p.$in == openapi.ParameterType.query)
+          .isNotEmpty;
       final name = toDartName(filterMethodName(operationName, tag ?? ''));
 
       var responses = <openapi.Response, List<String>>{};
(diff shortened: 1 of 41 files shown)


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -315,35 +321,40 @@ Iterable<Method> buildTags(
           toDartName(identifierBuilder.toString(), uppercaseFirstCharacter: true),
         );
 
-        final queryParams = <String>[];
-        for (final parameter in parameters) {
-          if (parameter.$in != openapi.ParameterType.query) {
-            continue;
-          }
+        if (!hasUriParameters) {
+          code.writeln("const _path = '${pathEntry.key}';");
```
provokateurin, line 356: This could already parse the Uri instead of first storing the path as a variable.

provokateurin (approved):
Very nice cleanup

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: What exactly do you mean?
Uri does not have a constant constructor.

provokateurin: But do we need it? In my eyes this makes the output code longer and a bit more complex which is not needed.

Leptopoda: Ok. 
What about leaving the path variable and just calling `Uri.Parse(path)` in the assignment for the request?
In theory we don't need it to be an URI at all as the DynamiteClient will just call `toString` to concatenate it with the baseUrl.

So just duplicate parsing :thinking:

provokateurin: So you just want to pass the full path as a String instead of a Uri? Sounds like a good plan to remove duplicate parsing

Issue #1245, opened 2023-12-07
Discuss used lint rules


Leptopoda:
We currently enable the `prefer_final_parameters` rule this is not common in dart and also affects #1244.
Disabling it should not affect performance or code size (to be investigated) and for `parameter_assignments` has us covered for the rest.

provokateurin:
And instead we enable `avoid_final_parameters`, right?

Leptopoda:
I did some testing a few days ago and as expected I was not able to find any difference.
Sadly from a `avoid_final_parameters` does not have a quick fix so I'd just disable it in the generated code for now. And the rest can be migrated in the future (if we really want to).

On a side note should we also ignore `no_leading_underscores_for_local_identifiers` in our generated code?

provokateurin:
Yes that could be ignored as well. dart fix will change it but it doesn't need to changed.

provokateurin:
`no_leading_underscores_for_local_identifiers` should still be changed

PR #1325, opened 2023-12-19
style: change parameter style to avoid_final_parameters

Leptopoda:
fixes: #1245

This does not migrate all of our packages and I don't see the need to do this any time soon.
For the bigger ones like dynamite or the framework we could also just keep it for ever.

### DIFF
--- a/packages/app/analysis_options.yaml
+++ b/packages/app/analysis_options.yaml
@@ -1,7 +1,1 @@
 include: package:neon_lints/flutter.yaml
-
-linter:
-  rules:
-    # TODO: migrate package to new lint rules
-    prefer_final_parameters: true
-    avoid_final_parameters: false
--- a/packages/app/analysis_options.yaml
+++ b/packages/app/analysis_options.yaml
@@ -1,1 +1,7 @@
 include: package:neon_lints/flutter.yaml
+
+linter:
+  rules:
+    # TODO: migrate package to new lint rules
+    prefer_final_parameters: true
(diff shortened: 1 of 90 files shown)


provokateurin:
I'd prefer a single style for all packages. Either remove them everywhere or keep them everywhere.

Leptopoda:
Can we still do this incrementally?

provokateurin:
Sure

packages/nextcloud/build.yaml
```diff
@@ -12,6 +12,7 @@ targets:
             - public_member_api_docs
             - unreachable_switch_case
             - camel_case_extensions
+            - no_leading_underscores_for_local_identifiers
```
provokateurin, line 15: Please introduce this in a separate PR as it has nothing to do with the final parameters.

provokateurin (changes requested):
Please only change the rules for the packages that you are migrating in this PR instead of changing the default and overriding it in all other packages.

Leptopoda:
> Please only change the rules for the packages that you are migrating in this PR instead of changing the default and overriding it in all other packages.

I disagree. While I will try to tackle this in the coming days the individual migrations (especially the framework) will take some time.

If we where to introduce a new package in that time it should automatically have the new rules.

provokateurin:
Ok I get that point. Then please add a TODO comment everywhere where the rules are manually changed.

provokateurin (approved):

PR #1326, opened 2023-12-20
refactor(nextcloud): Reuse containers in tests

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1300

It was possible to migrate some of the test suite without even changing the test content or fixtures. I could extract those into a separate PR so we can get those merged already. I'm not sure how long it will take to merge this complete PR, so we might want to do this to get some small improvements already.

@Leptopoda please let me know if you like this approach. It makes all tests very fast, except for news which I haven't migrated yet as it is quite hard. I might do some profiling of the news app to check why it is so slow in the first place and if that could be solved at all. If it turns out that adding the feed is slow then we won't really get much benefit from reusing the same container.


packages/nextcloud/test/user_status_test.dart
```diff
@@ -11,17 +11,22 @@ void main() {
...
         if (Invoker.current!.liveTest.errors.isNotEmpty) {
           print(await container.allLogs());
         }
         container.destroy();
       });
 
+      Future<void> resetStatus() async {
+        await client.userStatus.userStatus.setStatus(statusType: 'online');
+        await client.userStatus.userStatus.clearMessage();
+      }
```
Leptopoda, line 28: i think in these cases the reset can be called in a `setup` hook so we don't need to explicitly add it everywhere

provokateurin: Yes, but the problem here was that not all tests need it and I wanted to avoid unnecessary resets. Using a group would fix it, but then the grouping with the controllers/tags is not in place anymore. Not sure which option is better.

Leptopoda: In this case we can probably leave it. I prefer the grouping as is.

provokateurin:
I'm giving up on the news tests for now. The spec and tests need to be rewritten completely (not even talking about generating them). They are the oldest part of this project and it really shows.

Leptopoda (approved):

PR #1328, opened 2023-12-20
style: disable no_leading_underscores_for_local_identifiers for generated clients

Leptopoda:
…disable no_leading_underscores_for_local_identifiers for generated clients

dynamite_petstore is not really affected as it doesn't have any linting rules configured. PR for that is also coming.


provokateurin (approved):

provokateurin:
Commit type should be style ;)

PR #1332, opened 2023-12-21
chore(nextcloud): Update specs to 28.0.1

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1226

Draft for now, I'll wait until 28.0.1 has been tagged on server and notifications (although there are no diffs using the stable28 branches).

### DIFF
--- a/packages/nextcloud/lib/src/api/files_sharing.openapi.json
+++ b/packages/nextcloud/lib/src/api/files_sharing.openapi.json
@@ -538,7 +538,11 @@
                     },
                     "hide_download": {
                         "type": "integer",
-                        "format": "int64"
+                        "format": "int64",
+                        "enum": [
+                            0,
+                            1
+                        ]
                     },
                     "id": {
                         "type": "string"
@@ -579,7 +583,11 @@
                     },
                     "mail_send": {
                         "type": "integer",
-                        "format": "int64"
+                        "format": "int64",
+                        "enum": [
+                            0,
+                            1
+                        ]
                     },
                     "mimetype": {
                         "type": "string"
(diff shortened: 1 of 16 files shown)


packages/nextcloud/lib/src/api/files_sharing.openapi.json
```diff
@@ -2793,26 +2801,20 @@
...
-                                        "type": "array",
-                                        "items": {
-                                            "type": "integer",
-                                            "format": "int64"
-                                        }
                                     }
-                                ]
-                            }
+                                }
+                            ]
```
Leptopoda, line 2817: I'd like a unit test on that one.
It should not be affected by #1058 but I'd like to have it tested.

packages/nextcloud/lib/src/api/files_sharing.openapi.json
```diff
@@ -2916,26 +2918,20 @@
...
```
Leptopoda, line 2934: same here

Leptopoda (commented):
LGTM.

The unit tests do not need to be in NC if it is hard to reproduce both available cases.
Maybe add a dynamite_e2e test with `oneof(int|array<int>)` and `oneof(String|array<Object>)` (the later one being used in '/ocs/v2.php/apps/spreed/api/{apiVersion}/signaling/{token}'

provokateurin:
The spreed case of `oneof(String|array<Object>)` is already tested. Seems like oneOf with array is completely broken :/ I'm opening a separate PR with a test so we can merge this already.

provokateurin:
I think we can merge this already because the bug is in dynamite and doesn't block this change.

packages/nextcloud/lib/src/api/files_sharing.openapi.json
provokateurin: Nope this is exactly https://github.com/nextcloud/neon/pull/1215 :/

Leptopoda (approved):

Issue #1333, opened 2023-12-21
Mark generated news, notes and uppush code as experimental


provokateurin:
The specs are written by hand and are not complete. They might change unexpectedly once we have proper generate specs in place. Needs https://github.com/nextcloud/neon/pull/1174

Issue #1244, opened 2023-12-07
Reduce the dependance on `dart fix` in dynamite


Leptopoda:
Dynamite should build better code in the first place.
This will also reduce the time `generate-nextcloud` currently takes.

PR #1337, opened 2023-12-21
feat(dynamite): generate library for every spec

Leptopoda:
needed so we can mark libs as experimental for #1333 

Before you complain about the added package. I plan to also use other parts of it like string the [string escaping](https://pub.dev/documentation/source_helper/latest/source_helper/escapeDartString.html). This should help us a lot with #1244.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/openapi_builder.dart
+++ b/packages/dynamite/dynamite/lib/src/openapi_builder.dart
@@ -2,19 +2,19 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:build/build.dart';
-import 'package:built_collection/built_collection.dart';
 import 'package:checked_yaml/checked_yaml.dart';
 import 'package:code_builder/code_builder.dart';
 import 'package:dart_style/dart_style.dart';
 import 'package:dynamite/src/builder/client.dart';
 import 'package:dynamite/src/builder/generate_ofs.dart';
 import 'package:dynamite/src/builder/generate_schemas.dart';
-import 'package:dynamite/src/builder/imports.dart';
 import 'package:dynamite/src/builder/serializer.dart';
 import 'package:dynamite/src/builder/state.dart';
+import 'package:dynamite/src/helpers/dart_helpers.dart';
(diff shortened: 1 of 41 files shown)


packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -78,16 +78,45 @@ class OpenAPIBuilder implements Builder {
...
-        ..addAll(buildSerializer(state))
-        ..insertAll(0, generateImports(outputId, state));
+      final output = Library((final b) {
+        final analyzerIgnores = state.buildConfig.analyzerIgnores;
+        if (analyzerIgnores != null) {
+          b.ignoreForFile.addAll(analyzerIgnores);
+        }
+
+        b
+          ..name = toLibraryName(p.basenameWithoutExtension(inputId.path))
```
provokateurin, line 88: The `_openapi` part of the resulting library names should be dropped.

packages/dynamite/dynamite/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
   path: ^1.0.0
   pub_semver: ^2.1.4
   pubspec_parse: ^1.2.3
+  source_helper: ^1.3.4
```
provokateurin, line 20: All these constraints should probably be not so strict.

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/openapi_builder.dart
Leptopoda: I fear that we potentially get naming conflicts if we do so.

Leptopoda: And it is needed to avoid reserved names to be used.

`enum` and `$enum` are not allowed but `enum_openapi` is. We face this in the e2e specs

provokateurin:
I think you closed this unintentionally?

Leptopoda:
huh that's odd I only force pushed :thinking: 
And now I can't reopen it due to that force push.

Sorry but I'll have to make a new one.

PR #1338, opened 2023-12-22
Fix oneOf with array

provokateurin:
Found in https://github.com/nextcloud/neon/pull/1332


provokateurin:
Ah well this is exactly https://github.com/nextcloud/neon/pull/1215.

Issue #1336, opened 2023-12-21
Add linting to dynamite_petstore


Leptopoda:
Before we forget it again.

PR #1341, opened 2023-12-22
feat(dynamite_petstore_example): add linting

Leptopoda:
fixes: #1336

### DIFF
--- a/packages/dynamite/dynamite_petstore_example/pubspec.yaml
+++ b/packages/dynamite/dynamite_petstore_example/pubspec.yaml
@@ -13,6 +13,8 @@ dependencies:
     git:
       url: https://github.com/nextcloud/neon
       path: packages/dynamite/dynamite_runtime
+  meta: ^1.0.0
+  universal_io: ^2.0.0
   uri: ^1.0.0
 
 dev_dependencies:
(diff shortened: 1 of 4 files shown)


packages/dynamite/dynamite_petstore_example/pubspec.yaml
```diff
@@ -13,6 +13,8 @@ dependencies:
     git:
       url: https://github.com/nextcloud/neon
       path: packages/dynamite/dynamite_runtime
+  meta: ^1.11.0
+  universal_io: ^2.2.2
```
provokateurin, line 17: We won't publish this so we don't exactly need to care about the constraints, but someone might copy them so it's best to set the correct ones.

packages/dynamite/dynamite_petstore_example/pubspec.yaml
```diff
@@ -22,3 +24,7 @@ dev_dependencies:
     git:
       url: https://github.com/nextcloud/neon
       path: packages/dynamite/dynamite
+  neon_lints:
+    git:
+      url: https://github.com/nextcloud/neon
+      path: packages/neon_lints
```
provokateurin, line 30: This should also appear in pubspec_overrides.yaml :thinking:

provokateurin (changes requested):

provokateurin (approved):

PR #1342, opened 2023-12-22
refactor: Make bloc implementations private

provokateurin:
Some places accessed parts of Blocs that should have been in the interface. Now it is no longer possible to do that.

### DIFF
--- a/packages/neon/neon_notes/lib/src/blocs/note.dart
+++ b/packages/neon/neon_notes/lib/src/blocs/note.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 
 import 'package:flutter/foundation.dart';
 import 'package:neon_framework/blocs.dart';
+import 'package:neon_framework/models.dart';
 import 'package:neon_notes/src/blocs/notes.dart';
 import 'package:neon_notes/src/options.dart';
 import 'package:nextcloud/nextcloud.dart';
@@ -9,21 +10,37 @@ import 'package:nextcloud/notes.dart' as notes;
 import 'package:queue/queue.dart';
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotesNoteBlocEvents {
+sealed class NotesNoteBloc implements InteractiveBloc {
+  factory NotesNoteBloc(
+    final NotesBloc notesBloc,
+    final Account account,
+    final notes.Note note,
+  ) =>
+      _NotesNoteBloc(
+        notesBloc,
+        account,
+        note,
+      );
+
   void updateContent(final String content);
 
   void updateTitle(final String title);
 
   void updateCategory(final String category);
-}
 
-abstract interface class NotesNoteBlocStates {
   BehaviorSubject<String> get category;
+
+  NotesOptions get options;
+
+  String get initialContent;
+
+  String get initialTitle;
 }
 
-class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, NotesNoteBlocStates {
-  NotesNoteBloc(
+class _NotesNoteBloc extends InteractiveBloc implements NotesNoteBloc {
+  _NotesNoteBloc(
     this._notesBloc,
+    this.account,
     final notes.Note note,
   ) {
     _emitNote(note);
@@ -52,12 +69,17 @@ class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, Note
     });
   }
 
-  late final NotesOptions options = _notesBloc.options;
+  @override
+  NotesOptions get options => _notesBloc.options;
+
   final NotesBloc _notesBloc;
+  final Account account;
   final _updateQueue = Queue();
 
   late final int id;
+  @override
   late final String initialContent;
+  @override
   late final String initialTitle;
   late String _etag;
 
@@ -71,12 +93,12 @@ class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, Note
   BehaviorSubject<String> category = BehaviorSubject<String>();
 
   @override
-  void refresh() {}
+  Future<void> refresh() async {}
 
   @override
   void updateCategory(final String category) {
     _wrapAction(
-      (final etag) async => _notesBloc.account.client.notes.updateNote(
+      (final etag) async => account.client.notes.updateNote(
         id: id,
         category: category,
         ifMatch: '"$etag"',
@@ -87,7 +109,7 @@ class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, Note
   @override
   void updateContent(final String content) {
     _wrapAction(
-      (final etag) async => _notesBloc.account.client.notes.updateNote(
+      (final etag) async => account.client.notes.updateNote(
         id: id,
         content: content,
         ifMatch: '"$etag"',
@@ -98,7 +120,7 @@ class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, Note
   @override
   void updateTitle(final String title) {
     _wrapAction(
-      (final etag) async => _notesBloc.account.client.notes.updateNote(
+      (final etag) async => account.client.notes.updateNote(
         id: id,
         title: title,
         ifMatch: '"$etag"',
--- a/packages/neon/neon_notes/lib/src/blocs/notes.dart
+++ b/packages/neon/neon_notes/lib/src/blocs/notes.dart
@@ -8,7 +8,16 @@ import 'package:neon_notes/src/options.dart';
 import 'package:nextcloud/notes.dart' as notes;
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotesBlocEvents {
+sealed class NotesBloc implements InteractiveBloc {
+  factory NotesBloc(
+    final NotesOptions options,
+    final Account account,
+  ) =>
+      _NotesBloc(
+        options,
+        account,
+      );
+
   void createNote({
     final String title = '',
     final String category = '',
@@ -24,20 +33,21 @@ abstract interface class NotesBlocEvents {
   });
 
   void deleteNote(final int id);
-}
 
-abstract interface class NotesBlocStates {
   BehaviorSubject<Result<List<notes.Note>>> get notesList;
+
+  NotesOptions get options;
 }
 
-class NotesBloc extends InteractiveBloc implements NotesBlocEvents, NotesBlocStates {
-  NotesBloc(
+class _NotesBloc extends InteractiveBloc implements NotesBloc {
+  _NotesBloc(
     this.options,
     this.account,
   ) {
     unawaited(refresh());
   }
 
+  @override
   final NotesOptions options;
   final Account account;
 
--- a/packages/neon/neon_notifications/lib/src/blocs/notifications.dart
+++ b/packages/neon/neon_notifications/lib/src/blocs/notifications.dart
@@ -7,21 +7,25 @@ import 'package:neon_notifications/src/options.dart';
 import 'package:nextcloud/notifications.dart' as notifications;
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotificationsBlocEvents {
+sealed class NotificationsBloc implements NotificationsBlocInterface, InteractiveBloc {
+  factory NotificationsBloc(
+    final NotificationsOptions options,
+    final Account account,
+  ) =>
+      _NotificationsBloc(options, account);
+
+  @override
   void deleteNotification(final int id);
 
   void deleteAllNotifications();
-}
 
-abstract interface class NotificationsBlocStates {
   BehaviorSubject<Result<List<notifications.Notification>>> get notificationsList;
 
   BehaviorSubject<int> get unreadCounter;
 }
 
-class NotificationsBloc extends InteractiveBloc
-    implements NotificationsBlocInterface, NotificationsBlocEvents, NotificationsBlocStates {
-  NotificationsBloc(
+class _NotificationsBloc extends InteractiveBloc implements NotificationsBlocInterface, NotificationsBloc {
+  _NotificationsBloc(
     this.options,
     this._account,
   ) {
@@ -35,7 +39,6 @@ class NotificationsBloc extends InteractiveBloc
     _timer = TimerBloc().registerTimer(const Duration(seconds: 30), refresh);
   }
 
-  @override
   final NotificationsOptions options;
   final Account _account;
   late final NeonTimer _timer;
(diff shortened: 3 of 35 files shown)


packages/neon/neon_notes/lib/src/blocs/note.dart
```diff
@@ -2,28 +2,45 @@ import 'dart:async';
...
+import 'package:neon_framework/models.dart';
 import 'package:neon_notes/src/blocs/notes.dart';
 import 'package:neon_notes/src/options.dart';
 import 'package:nextcloud/nextcloud.dart';
 import 'package:nextcloud/notes.dart' as notes;
 import 'package:queue/queue.dart';
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotesNoteBlocEvents {
+abstract interface class NotesNoteBloc implements InteractiveBloc {
```
Leptopoda, line 13:
```suggestion
sealed class NotesNoteBloc implements InteractiveBloc {
```

packages/neon/neon_notes/lib/src/blocs/note.dart
```diff
@@ -71,12 +93,12 @@ class NotesNoteBloc extends InteractiveBloc implements NotesNoteBlocEvents, Note
   BehaviorSubject<String> category = BehaviorSubject<String>();
 
   @override
-  void refresh() {}
+  Future<void> refresh() async {}
```
Leptopoda, line 96:
```suggestion
  FutureOr<void> refresh() {}
```

packages/neon/neon_notes/lib/src/blocs/notes.dart
```diff
@@ -8,7 +8,16 @@ import 'package:neon_notes/src/options.dart';
 import 'package:nextcloud/notes.dart' as notes;
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotesBlocEvents {
+abstract interface class NotesBloc implements InteractiveBloc {
```
Leptopoda, line 11:
```suggestion
sealed class NotesBloc implements InteractiveBloc {
```

packages/neon/neon_notifications/lib/src/blocs/notifications.dart
```diff
@@ -7,21 +7,25 @@ import 'package:neon_notifications/src/options.dart';
 import 'package:nextcloud/notifications.dart' as notifications;
 import 'package:rxdart/rxdart.dart';
 
-abstract interface class NotificationsBlocEvents {
+abstract interface class NotificationsBloc implements NotificationsBlocInterface, InteractiveBloc {
```
Leptopoda, line 10:
```suggestion
sealed class NotificationsBloc implements NotificationsBlocInterface, InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/accounts.dart
```diff
@@ -22,34 +22,38 @@ import 'package:rxdart/rxdart.dart';
 
 const _keyAccounts = 'accounts';
 
-/// Events for the [AccountsBloc].
-@internal
-abstract interface class AccountsBlocEvents {
+/// The Bloc responsible for managing the [Account]s
+abstract interface class AccountsBloc implements Disposable {
```
Leptopoda, line 26:
```suggestion
sealed class AccountsBloc implements Disposable {
```

packages/neon_framework/lib/src/blocs/apps.dart
```diff
@@ -17,19 +17,28 @@ import 'package:nextcloud/nextcloud.dart';
 import 'package:provider/provider.dart';
 import 'package:rxdart/rxdart.dart';
 
-/// Events for the [AppsBloc].
+/// The Bloc responsible for managing the [AppImplementation]s.
 @internal
-abstract interface class AppsBlocEvents {
+abstract interface class AppsBloc implements InteractiveBloc {
```
Leptopoda, line 22:
```suggestion
sealed class AppsBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/user_details.dart
```diff
@@ -9,16 +9,14 @@ import 'package:nextcloud/provisioning_api.dart' as provisioning_api;
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class UserDetailsBlocEvents {}
+abstract interface class UserDetailsBloc implements InteractiveBloc {
```
Leptopoda, line 12:
```suggestion
sealed class UserDetailsBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/unified_search.dart
```diff
@@ -11,24 +11,29 @@ import 'package:nextcloud/core.dart' as core;
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class UnifiedSearchBlocEvents {
+abstract interface class UnifiedSearchBloc implements InteractiveBloc {
```
Leptopoda, line 14:
```suggestion
sealed class UnifiedSearchBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/timer.dart
```diff
@@ -4,74 +4,77 @@ import 'dart:ui';
 import 'package:meta/meta.dart';
 import 'package:neon_framework/src/bloc/bloc.dart';
 
-@internal
-abstract interface class TimerBlocEvents {
+abstract interface class TimerBloc extends Bloc {
```
Leptopoda, line 7:
```suggestion
sealed class TimerBloc extends Bloc {
```

packages/neon_framework/lib/src/blocs/push_notifications.dart
```diff
@@ -15,14 +15,19 @@ import 'package:nextcloud/notifications.dart' as notifications;
 import 'package:unifiedpush/unifiedpush.dart';
 
 @internal
-abstract interface class PushNotificationsBlocEvents {}
-
-@internal
-abstract interface class PushNotificationsBlocStates {}
+abstract interface class PushNotificationsBloc {
```
Leptopoda, line 18:
```suggestion
sealed class PushNotificationsBloc {
```

packages/neon_framework/lib/src/blocs/next_push.dart
```diff
@@ -5,21 +5,29 @@ import 'package:meta/meta.dart';
...
 import 'package:neon_framework/src/blocs/accounts.dart';
 import 'package:neon_framework/src/models/account.dart';
+import 'package:neon_framework/src/models/disposable.dart';
 import 'package:neon_framework/src/utils/global_options.dart';
 import 'package:nextcloud/uppush.dart' as uppush;
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class NextPushBlocEvents {}
+abstract interface class NextPushBloc implements Disposable {
```
Leptopoda, line 14:
```suggestion
sealed class NextPushBloc implements Disposable {
```

packages/neon_framework/lib/src/blocs/login_flow.dart
```diff
@@ -10,18 +10,16 @@ import 'package:nextcloud/nextcloud.dart';
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class LoginFlowBlocEvents {}
+abstract interface class LoginFlowBloc implements InteractiveBloc {
```
Leptopoda, line 13:
```suggestion
sealed class LoginFlowBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/login_check_server_status.dart
```diff
@@ -10,18 +10,15 @@ import 'package:nextcloud/nextcloud.dart';
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class LoginCheckServerStatusBlocEvents {}
+abstract interface class LoginCheckServerStatusBloc implements InteractiveBloc {
```
Leptopoda, line 13:
```suggestion
sealed class LoginCheckServerStatusBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/login_check_account.dart
```diff
@@ -11,18 +11,24 @@ import 'package:nextcloud/provisioning_api.dart' as provisioning_api;
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class LoginCheckAccountBlocEvents {}
+abstract interface class LoginCheckAccountBloc implements InteractiveBloc {
```
Leptopoda, line 14:
```suggestion
sealed class LoginCheckAccountBloc implements InteractiveBloc {
```

packages/neon_framework/lib/src/blocs/first_launch.dart
```diff
@@ -2,21 +2,23 @@ import 'dart:async';
 
 import 'package:meta/meta.dart';
 import 'package:neon_framework/src/bloc/bloc.dart';
+import 'package:neon_framework/src/models/disposable.dart';
 import 'package:neon_framework/src/settings/models/storage.dart';
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class FirstLaunchBlocEvents {}
+abstract interface class FirstLaunchBloc implements Disposable {
```
Leptopoda, line 10:
```suggestion
sealed class FirstLaunchBloc implements Disposable {
```

packages/neon_framework/lib/src/blocs/capabilities.dart
```diff
@@ -9,17 +9,15 @@ import 'package:nextcloud/core.dart' as core;
 import 'package:rxdart/rxdart.dart';
 
 @internal
-abstract interface class CapabilitiesBlocEvents {}
+abstract interface class CapabilitiesBloc implements InteractiveBloc {
```
Leptopoda, line 12:
```suggestion
sealed class CapabilitiesBloc implements InteractiveBloc {
```

Leptopoda (changes requested):
I think that the interfaces should be sealed.
1. I don't like interface classes with constructors/factories
2. We prevent our users to extend these interfaces.

Long term the concrete implementations should not contain private fields as the classes themselves are already private. You already did so for the  TimerBloc. But please not in this PR as it's already large enoguh :)

provokateurin:
> I don't like interface classes with constructors/factories

How should I fix it?

> Long term the concrete implementations should not contain private fields as the classes themselves are already private

Agreed, I didn't do it to keep the diff more readable.

Leptopoda:
> How should I fix it?

a sealed class can no longer have the interface modifier :)
So just make them sealed as suggested in the review.

provokateurin:
But that doesn't make the factories go away, no? I thought you saw that as a problem too.

Leptopoda:
I'm fine with them :)

provokateurin:
Making the classes sealed prevents us from creating mocks (at least with mocktail) :/

I think we should still go for it as we shouldn't need to have mocks for blocs once we untangle them.

packages/neon_framework/lib/src/bloc/bloc.dart
```diff
@@ -35,7 +35,7 @@ abstract class InteractiveBloc extends Bloc {
   /// Refreshes the state of the bloc.
   ///
   /// Commonly involves re-fetching data from the server.
-  FutureOr<void> refresh();
+  Future<void> refresh();
```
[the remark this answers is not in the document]
provokateurin: The problem is that FutureOr doesn't play nice with tear-offs and other async code. I'll push a second commit but you will see that it is very ugly. I'd just go with Future to avoid this

Leptopoda:
> Making the classes sealed prevents us from creating mocks (at least with mocktail) :/

I haven't tried it yet and don't know if it would help but `meta` also has a `sealed` annotation. This would only produce lint warnings instead of compile time errors when one tries to extend it. Maybe this is a better idea :thinking:

packages/neon_framework/lib/src/bloc/bloc.dart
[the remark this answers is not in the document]
Leptopoda: I fear that this will not play nice when interpreted (like in an js environment). Maybe the compiler can optimize this :thinking: 
I think that we have two solutions.
1. add a separate synchronous method like `backgroundRefresh()`
2. just let it return a future and ignore the lint error. When thinking about the bloc pattern we have a stream of events that lead to a different stream of events being fired (just a recap). Adding an event to a stream happen synchronously so there should be no reason to actually await it. The emitted states should already handle changes in the UI. If somebody really needs to await the event this means that business logic is located in the UI code which is what we try to avoid in the first place.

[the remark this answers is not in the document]
provokateurin: The problem is we rely on the future for the refresh indicator in lists. There is no easy way to make the refresh indicators stateful so that we can hook them to the streams. We could also just ignore that and make the function a void and then the refresh indicators always hide instantly because there's nothing to await. I prefer that as we already have the linear progress indicator to show that something is loading at the moment.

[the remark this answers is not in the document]
Leptopoda: ok let's keep it until we tackle https://github.com/nextcloud/neon/issues/800

Leptopoda:
I'll review this later today

packages/neon_framework/lib/src/blocs/accounts.dart
```diff
@@ -22,34 +23,40 @@ import 'package:rxdart/rxdart.dart';
 
 const _keyAccounts = 'accounts';
 
-/// Events for the [AccountsBloc].
-@internal
-abstract interface class AccountsBlocEvents {
+/// The Bloc responsible for managing the [Account]s
+// TODO: Make this class properly sealed once we do not need mocks for the Bloc anymore.
```
Leptopoda, line 27: Under what circumstance should we no longer need to mock this bloc?
I'm still not entirely sure that sealed is the best thing.

At least it's not breaking if we decide to change it later on.

provokateurin: As far as I understood we want to untangle the Blocs in a way that the blocs don't know about each other. In that case we don't need to mock it anymore. I'm not sure if this will actually work, as all the widgets still use the bloc directly.

provokateurin: Let's remove the TODO for now and keep the sealed annotation. When we need to mock other blocs we can do the same thing and just ignore the lint.

Leptopoda: > I'm not sure if this will actually work, as all the widgets still use the bloc directly.

That's exactly the problem. We might not need it for testing individual blocs anymore but other widgets might still need a mock.

Leptopoda (approved):

packages/neon_framework/lib/src/blocs/accounts.dart
provokateurin: Yeah, but I think the sealed annotation is a good compromise for that

PR #1343, opened 2023-12-22
fix(ci): run dart tests with dart instead of flutter

Leptopoda:
needed to fix #1337

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -27,6 +27,8 @@ scripts:
   format: dart format --fix --line-length 120 .
   format:check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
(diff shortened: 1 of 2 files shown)


melos.yaml
```diff
@@ -27,6 +27,6 @@ scripts:
   format: dart format --fix --line-length 120 .
   format:check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: dart analyze --fatal-infos .
-  test: melos exec --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
+  test: melos exec --no-flutter --concurrency=1 --fail-fast --dir-exists=test -- "dart test --concurrency=$(nproc --all)" && melos exec --flutter --concurrency=1 --fail-fast --dir-exists=test -- "flutter test --concurrency=$(nproc --all)"
```
provokateurin, line 30: Is it possible to use the multiline syntax here? Then it would be more readable

provokateurin (approved):

PR #1348, opened 2023-12-23
chore: Set correct publish_to

provokateurin:
Leaves nextcloud, dynamite and dynamite_runtime as non-private packages.


Leptopoda (approved):

provokateurin:
Ah well I can't do this for nextcloud until dynamite_runtime is released and we no longer need the git dependency.

Leptopoda:
> Ah well I can't do this for nextcloud until dynamite_runtime is released and we no longer need the git dependency.

but we are close #1349 :P

PR #1350, opened 2023-12-23
fix(dynamite): too long names for someOf typedefs

Leptopoda:


provokateurin (commented):
Ugh ok no we can't do this, it's a horrible user experience. We are just working around a problem with dart doc here, we need to fix it there.

Leptopoda:
What do you mean?

In theory we can make them visibleForTesting. Users would loose the ability to have a `fromJson` reviver but I doubt that  this is a problem.
I also don't think that there is a way to fix this other than dartdoc generating html files with hashed names and using some sort of javaScript routing to handle the correct url client side. And I don't see them investing is such work.

packages/dynamite/dynamite_end_to_end_test/lib/type_defs.openapi.dart
```diff
@@ -58,30 +58,31 @@ abstract class NestedRedirect implements $NestedRedirectInterface, Built<NestedR
 }
 
 typedef SomeOfRedirect = ({Base? base, int? $int, JsonObject? jsonObject});
-typedef $BaseIntJsonObject = ({Base? base, int? $int, JsonObject? jsonObject});
+typedef $e1c7ecea8e5fdae7b94cd86c0dc4f1ba = ({Base? base, int? $int, JsonObject? jsonObject});
```
Leptopoda, line 61: As one can see here there is always going to be a typedef with a usable name.
The hashed name is only needed for the serializer to match all typedefs with those usable names.

provokateurin:
Well in some places one needs to create objects, e.g. spreed signaling. Using some weird hash name is not ok there

Leptopoda:
> Well in some places one needs to create objects, e.g. spreed signaling. Using some weird hash name is not ok there

Please see the above code comment.

packages/dynamite/dynamite_end_to_end_test/lib/type_defs.openapi.dart
provokateurin: I disagree about the "usable". Also in the generate docs, the names would be meaningless. There's no way we can push this onto the package consumers.

provokateurin: I'd say not having a typedef is more user friendly in this case.

Leptopoda: The problem is that we need the typedef for the serializer as it can not work with raw records (the runtimetype of them is not constant while the one of a typedef is. So it also has performance implications).

Therefore we generate a typedef for use by the devs (`SomeOfRedirect` and one to use by the serializer `$e1c7ecea8e5fdae7b94cd86c0dc4f1ba`).
A dev should never need to use the later one as everything (other than the fromJson method that's static to the extension).

Leptopoda: If the fromJson reviver is that important We can generate extensions for the human readable version to only handle that while having a common extension (on the hashed name) that handles the non static methods.

Leptopoda: And if you are curious we can not generate extension methods with the non static methods for every typedef as ones this would lead to conflicts between someOfs with the same type.

provokateurin: Ok, but how would you generate human readable extension names that are not conflicting and not too long?

Leptopoda: We already do generate human readable ones.
That's the whole point.

The not human readable ones are not needed by the devs. I just made them private and will push these changes later. Maybe this helps in understanding.

Leptopoda: Sorry, I didn't find the time earlier.
Does it make sense now?

provokateurin:
Just to be sure I understand the approach: The serializers and records are private and hashed, but we have proxy extensions that are human readable. There might be multiple proxy extensions pointing to the same type as they would otherwise have to have the long names we had problems. Is this correct?

One problem I still see is that someone could create a schema that has >255 characters, but we can't do much about that and it would be quite stupid anyway.

Leptopoda:
> One problem I still see is that someone could create a schema that has >255 characters, but we can't do much about that and it would be quite stupid anyway.

The good thing is that dynamite would not fail. They would just not be able to create documentation for their generated code (which is out of our responsibility).

provokateurin:
Can you confirm that I understand this change correctly?

provokateurin:
The commit message has a typo btw (`too` not `to`)

Leptopoda:
Ah sorry.
Yes you understand that correctly.

cspell.json
```diff
@@ -11,7 +11,9 @@
     "**.mocks.dart",
     "**/assets",
     "**/l10n/!(en.arb)",
+    "**/*.openapi.dart",
```
provokateurin, line 14:
```suggestion
    "**.openapi.dart",
```

cspell.json
```diff
@@ -11,7 +11,9 @@
...
     "external",
+    "packages/dynamite/dynamite_end_to_end_test/test/*.dart",
```
provokateurin, line 16: This shouldn't be needed anymore

Leptopoda:
> The commit message has a typo btw (`too` not `to`)

Is the rest fine? So I only need to push once more?

packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
```diff
@@ -17,7 +17,7 @@ String toDartName(
     }
   }
 
-  if (_reservedNames.contains(result) || RegExp(r'^[0-9]+$', multiLine: true).hasMatch(result)) {
+  if (_reservedNames.contains(result) || RegExp(r'^[0-9]+$', multiLine: true).hasMatch(result[0])) {
```
provokateurin, line 20: Why this change?

packages/dynamite/dynamite/pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   checked_yaml: ^2.0.0
   code_builder: ^4.5.0
   collection: ^1.0.0
+  crypto: ^3.0.3
```
provokateurin, line 14:
```suggestion
  crypto: ^3.0.0
```
Should be enough

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
```diff
@@ -112,8 +137,9 @@ Iterable<Spec> generateSomeOf(
 
   yield Extension(
     (final b) => b
-      ..name = '${identifier}Extension'
+      ..name = '${identifier}Extension'.nonPrivate
       ..on = refer(identifier)
+      ..docs.add('// ignore: library_private_types_in_public_api')
```
provokateurin, line 142: The whole extension could be private so we don't need this, right?

provokateurin: Then we also wouldn't need to hash the name as it won't appear in the docs anyway :thinking:

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
Leptopoda: Not needed anymore as we now prefix all hashed names with a `$`.
I'll put it in a separate PR

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
Leptopoda: This would also hide the `toJson` and `validate` methods` which might be interesting to devs.

packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
provokateurin: But then the expression will always be false because it never starts with a number?

packages/dynamite/dynamite/lib/src/builder/generate_ofs.dart
provokateurin: But we could expose those through the proxy extension? I like this way more as it will solve this problem the right way imo

Leptopoda: But we could expose those through the proxy extension? 
Nope as you would have multiple extension declaring the same method.

Which one should be used?
```dart
print('value'.name);

extension Extension1 on String{
  String get name => '1';
}

extension Extension2 on String{
  String get name => '2';
}
```
I know that one could explicitly specify this like `Extension1('value').name`. But that's tedious.

provokateurin: Ah I see :/
I would still prefer to make the whole extension private. We expose the serializers so if someone really needs toJson they can do it manually.

provokateurin: And we can still make it public later, it wouldn't be a breaking change. Keeping it public now and making it private later will be a breaking change.

Leptopoda: The seriailizers with all registered builder factories needed are still annotated with `visibleForTesting`.

provokateurin: But that we can also change later without it being a breaking change if necessary.

provokateurin: But I realized in this case we can also argue against the proxy extension and just make it all private. If serializing/deserializing is necessary then we need to expose the jsonSerializers.

Leptopoda: I think the current implementation is fine.
We can still add doc comments to the ugly extension later.

provokateurin: But do we really want to have the hashed extensions appear in the docs? Is there maybe an annotation that removes it from the docs but makes it still accessible (not like `@internal`)?

provokateurin (approved):
Please fix the typo in the commit message and pr title before merging

Leptopoda:
Can I get a new review?
I made the serializer getter private and added one per human readable name.
As it was static this would have been one (rare) case, where a dev might have encountered the hashed name.

provokateurin (approved):

PR #1352, opened 2023-12-24
fix(neon_framework): Remove progress indicator height jumping around

provokateurin:
We always want to display it with the exact height to prevent items in a listview or column from jumping around because of the height difference.


Leptopoda (approved):
Does this also fix https://github.com/nextcloud/neon/issues/890?

provokateurin:
Nope, that one is about the images actually reloading and not just jumping up and down a few pixels.

Leptopoda:
I just thought that by not needing to re layout the list view this would also fix this one at the same time :(

provokateurin:
Wait let me check, maybe it does fix it. I just thought it wouldn't because I did it to fix a different thing.

provokateurin:
Nope it doesn't fix it.

PR #1355, opened 2023-12-25
chore(neon_framework): Switch to file-level missing docs ignores

provokateurin:
When adding new things now we are properly warned when there are missing docs.


Leptopoda (approved):
Should we do the same for files,news,notifications and notes?

provokateurin:
All of them have basically no docs so I don't see the benefit for now. At some point we should disable the lint completely though

PR #1363, opened 2023-12-26
Release nextcloud package

provokateurin:
Remaining problems before it can be released.


packages/nextcloud/generate_exports.dart
```diff
@@ -1,4 +1,3 @@
-// ignore_for_file: depend_on_referenced_packages
 import 'dart:io';
```
Leptopoda, line 1: I'd rather put it into tools/

packages/nextcloud/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## 4.0.0
```
Leptopoda, line 1: Should we also add the old changelogs from the old package?

Otherwise our changelog will start at 4.0 which is odd.

provokateurin: I don't think the old changelog is useful because literally everything has changed.

packages/nextcloud/generate_exports.dart
provokateurin: Yeah, I also don't like this. I thought about recreating this with bash but I think it is too much headache...

Leptopoda (approved):
Let's get the release done.
We can still change these minor things

PR #1367, opened 2023-12-27
refactor(dynamite): only use raw strings where necessary

Leptopoda:
towards: #1244
The main issue is some missing comma in the fulltype but propperly fixing this is a bit harder.
I've some eperiments locally but they require major restructures of dynamite so probably something I'll come back to after the neon release.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/client.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/client.dart
@@ -477,11 +477,14 @@ String buildParameterSerialization(
   }
   final dartName = toDartName(parameter.name);
   final serializedName = '\$$dartName';
-
-  final buffer = StringBuffer()..write('var $serializedName = ${result.serialize(dartName)};');
+  final buffer = StringBuffer();
 
   if ($default != null) {
-    buffer.writeln('$serializedName ??= $defaultValueCode;');
+    buffer
+      ..write('var $serializedName = ${result.serialize(dartName)};')
+      ..writeln('$serializedName ??= $defaultValueCode;');
+  } else {
+    buffer.write('final $serializedName = ${result.serialize(dartName)};');
   }
 
   if (parameter.schema != null) {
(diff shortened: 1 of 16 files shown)


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -472,15 +473,18 @@ String buildParameterSerialization(
...
   final serializedName = '\$$dartName';
-
-  final buffer = StringBuffer()..write('var $serializedName = ${result.serialize(dartName)};');
+  final buffer = StringBuffer();
 
   if ($default != null) {
-    buffer.writeln('$serializedName ??= $defaultValueCode;');
+    buffer
+      ..write('var $serializedName = ${result.serialize(dartName)};')
+      ..writeln('$serializedName ??= $defaultValueCode;');
```
provokateurin, line 485: Why not directly use ?? and make the variable final

packages/dynamite/dynamite/lib/src/models/type_result/type_result.dart
```diff
@@ -107,13 +109,13 @@ sealed class TypeResult {
...
       ..write(object)
-      ..write(', specifiedType: const $fullType)');
+      ..write(', specifiedType: $fullType)');
 
     if (!nullable) {
       buffer.write('!');
     }
 
-    return '($buffer as $name)';
+    return '$buffer as $name';
```
provokateurin, line 118: This is usually needed for safety. If someone would call `.toString()` or anything else on it it will result in broken code. I assume this never happens as otherwise the code in this PR would be broken?

Leptopoda: it only happens when validating a serialized someOf (we did need it before as the encoding would call methods on the serialized object).

We should not need this anymore.

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: I think the output is way cleaner (and we already do so).

provokateurin (approved):

PR #1374, opened 2023-12-28
refactor: Only make bloc factories internal

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1372

For the neon_* packages this has not effect since they still have the global ignore set (which we should change eventually).

### DIFF
--- a/packages/neon_framework/lib/src/blocs/login_check_account.dart
+++ b/packages/neon_framework/lib/src/blocs/login_check_account.dart
@@ -1,3 +1,5 @@
+// ignore_for_file: public_member_api_docs
+
 import 'dart:async';
 
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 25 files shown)


packages/neon_framework/lib/src/blocs/login_check_account.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: public_member_api_docs
```
Leptopoda, line 1: IIRC the constructors had docs before all these changes. Please restore them or add new ones.

Leptopoda (changes requested):
It's not that hard to write propper documentation here and we should do so moving forward.

packages/neon_framework/lib/src/blocs/login_check_account.dart
provokateurin: Nope that was not the case.

Leptopoda: Then please add docs.

provokateurin:
I agree and I plan to do that, but for now I only want to change this to enable it. I'll create another issue to document all the blocs.

provokateurin:
As the commit says this is only a refactor to enabled easier detection of missing docs. I do not want to add the docs in this PR already.

Leptopoda (approved):
I still disagree.
It's not a big deal to add a few lines of docs. I fear that any open issue will just rot (similar to https://github.com/nextcloud/neon/issues/1119)

PR #1375, opened 2023-12-28
refactor(neon_framework): Allow relative time formatting of Durations

provokateurin:
Needed for https://github.com/nextcloud/neon/issues/442


Leptopoda (changes requested):
missing unit test (would also be great if the DateTime one also gets one)

provokateurin:
Neat, adding the tests even uncovered a bug that has been there for a long time :see_no_evil:

Leptopoda (approved):

PR #1378, opened 2023-12-29
feat(neon_framework): Support setting user status

provokateurin:
Depends on https://github.com/nextcloud/neon/pull/1375
Closes https://github.com/nextcloud/neon/issues/442

TODO:
 - [x] Fix grid layout
 - [x] Write tests


packages/neon_framework/lib/src/widgets/user_status_dialog.dart
```diff
@@ -0,0 +1,438 @@
...
+          SizedBox(
+            width: (constraints.maxWidth - 24 * 2) / 2,
+            child: Column(
+              crossAxisAlignment: CrossAxisAlignment.stretch,
+              children: children.sublist((children.length / 2).ceil(), children.length),
+            ),
+          ),
+        ],
+      );
+}
```
provokateurin, line 391: I'm trying to use a Wrap here to correctly display it on multiple screen sizes, but I can't get it to work at all without giving it a predefined height which of course breaks different screen sizes. @Leptopoda do you have an idea how this could work? Would it be possible to solve this mess using slivers?

Leptopoda:
Initial imoression: the picker looks beautiful :heart_eyes:

packages/neon_framework/lib/src/widgets/user_status_dialog.dart
Leptopoda: It should be enough to just make the two grids a GridView.
I don't think that slivers are really needed here (but there is also nothing against it).

provokateurin: I couldn't get the GridViews to work either. The reason why I want to use Wrap is that it will wrap them on smaller screens. The grid will look weird because the tiles need the horizontal space and should avoid line breaks in the text.

provokateurin: I still didn't get it to work. Can you have a look at it?

provokateurin:
Just setting the icon/emoji will not work here because of a bug in server: https://github.com/nextcloud/server/pull/42532

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+              .map(
+                (final child) => SizedBox(
+                  height: 72,
+                  width: constraints.maxWidth / 2,
+                  child: child,
+                ),
+              )
+              .toList(),
+        ),
+      );
```
Leptopoda, line 948:
```suggestion
  Widget buildGrid(
    final BoxConstraints constraints,
    final List<Widget> children,
  ) =>
      Flexible(
        child: Wrap(
          children: children
              .map(
                (final child) => SizedBox(
                  height: 72,
                  width: constraints.maxWidth / 2,
                  child: child,
                ),
              )
              .toList(),
        ),
      );
```

Does this really simple trick work?

provokateurin: Nope, already tried it and just checked again.

Leptopoda: Then I don't really understand what you want.
Can you describe it? or two mockups for different screen sizes (figma/penpot?)

provokateurin: It is quite simple: If there is enought horizontal screen space then the tiles should be displayed in two columns (achieved by setting the width to half of the max width). If the space is not enough then it should fallback to a single column (which is done by the Wrap automatically).

Leptopoda: And the wrapped tile should take the entire width of the dialog or still the same size as if it wasn't collapsed?

If it should resize the solution is a GridView if it should stay the same size (which I find more pleasant when resizing on the desktop the above solution should work (I'll double check with my local changes but I got it to work yesterday).

provokateurin: Yeah it should stay the same size. Please tell me how you got it working :D

provokateurin:
(just a rebase)

packages/neon_framework/lib/src/widgets/dialog.dart
Leptopoda: My bad you'll also need to remove the nested Columns that are used for the loading and error handling (I dislike this pattern so much that I immediately changed it before trying the Flexible but forgot to mention it yesterday).

Leptopoda: Should I do a complete review as there are also other code changes that I'd like to be made?

provokateurin: I haven't written any tests yet so that is still to follow, but you can review the existing code if you like to do it already.

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
         ],
       );
 }
+
+/// Shows an emoji picker.
+///
+/// When the user selects an emoji the dialog will pop and return the emoji.
+class NeonEmojiPickerDialog extends StatelessWidget {
```
Leptopoda, line 562: The emoji picker has an unconstrained hight.
This looks bad when the window is full screen

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return NeonDialog(
+      content: SizedBox(
+        width: NeonDialogTheme.of(context).constraints.maxWidth,
+        child: EmojiPicker(
```
Leptopoda, line 575: NeonDialog already does constrain the width for us (oversight in a prior rebase)

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+                  child: ListTile(
+                    leading: NeonServerIcon(icon: 'user-status-${onlineStatus.key}'),
+                    title: Text(title),
+                    subtitle: subtitle != null ? Text(subtitle) : null,
+                  ),
+                ),
+              ),
+            );
+          },
+        ).toList();
```
Leptopoda, line 853: Please externalize this card into a separate widget that can be reused by the predefined statuses

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+                },
+                child: Center(
+                  child: ListTile(
+                    leading: NeonServerIcon(icon: 'user-status-${onlineStatus.key}'),
+                    title: Text(title),
+                    subtitle: subtitle != null ? Text(subtitle) : null,
+                  ),
+                ),
+              ),
+            );
```
Leptopoda, line 851:
```suggestion
            return Card(
              clipBehavior: Clip.hardEdge,
              shape: types.contains(result.data?.status) ? selectedBorder : null,
              child: ListTile(
                leading: NeonServerIcon(icon: 'user-status-${onlineStatus.key}'),
                title: Text(title),
                subtitle: subtitle != null ? Text(subtitle) : null,
                onTap: () {
                  bloc.setStatusType(onlineStatus.key);
                },
              ),
            );
```

Just set the clip behavior on the card if you want to avoid the splash overflowing.

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+  }
+
+  Widget _buildPredefinedStatuses() => ResultBuilder.behaviorSubject(
+        subject: bloc.predefinedStatuses,
+        builder: (context, predefinedStatusesResult) => ResultBuilder.behaviorSubject(
+          subject: bloc.status,
+          builder: (context, statusResult) {
+            final selectedPredefinedStatus = statusResult.hasData && statusResult.requireData is user_status.Private
+                ? (statusResult.requireData as user_status.Private).messageId
+                : null;
```
Leptopoda, line 878:
```suggestion
            final statusResultData = statusResult.data;
            String? selectedPredefinedStatus;
            if (statusResultData is user_status.Private) {
              selectedPredefinedStatus = statusResultData.messageId;
            }
```
or
```suggestion
            final statusResultData = statusResult.data;
            final selectedPredefinedStatus = statusResultData is user_status.Private? 
              selectedPredefinedStatus = statusResultData.messageId: null;
```


Local variables can be promoted by typechecks. I think this is cleaner.

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +555,395 @@ class NeonUnifiedPushDialog extends StatelessWidget {
...
+              .map(
+                (child) => SizedBox(
+                  height: 72,
+                  width: constraints.maxWidth / 2,
+                  child: child,
+                ),
+              )
+              .toList(),
+        ),
+      );
```
Leptopoda, line 948: Should both be handled by the common widget.

Leptopoda (commented):
Screenshot_20240101_182939

On small display sizes the clear at selector should be underneath the message box.

packages/neon_framework/lib/src/widgets/dialog.dart
provokateurin: Without setting the width we end up with `LayoutBuilder does not support returning intrinsic dimensions`.

packages/neon_framework/lib/src/widgets/dialog.dart
provokateurin: According to https://m3.material.io/components/dialogs/specs there is no height constraint, but I agree that it doesn't look good. What do you think about setting the constraint in our dialog theme to twice the width? I think that is a reasonable limit that will still look good.

packages/neon_framework/lib/src/widgets/dialog.dart
provokateurin: wdym?

Leptopoda: https://github.com/nextcloud/neon/pull/1378#discussion_r1439063763

packages/neon_framework/lib/src/widgets/dialog.dart
Leptopoda: Ideally it should be the same size as the NeonUserStatusDialog

provokateurin: But it is generic and will be used in other places too (like the message input in Talk).

Leptopoda: shouldn't talk use a bottom modal that replaces the keyboard as many other messaging apps do?
It's not that important for now so feel free to use the suggested 1.5x factor.

provokateurin:
Putting the message input and clear at in the same line was a bit too complex and I think this now looks better than before. No tests yet, but apart from that it is done.

provokateurin:
I will rebase this onto https://github.com/nextcloud/neon/pull/1418 once it is merged and change the clear at dropdown so it is more usable.

provokateurin:
just a rebase, i still need to add the tests :/ haven't looked further into mocking the client or request manager.

provokateurin:
@Leptopoda please review the bloc tests. I'm not sure yet if the way I did it is good, but I feel like it is the best we can do right now. Once we have more blocs with tests I'd like to merge the common logic, but there is no point in that right now.

provokateurin:
@Leptopoda added widget tests as well, this should now have all the tests we need

packages/neon_framework/test/user_status_bloc_test.dart
```diff
@@ -0,0 +1,218 @@
...
+      bloc.predefinedStatuses.transformResult((e) => e.single.id),
+      emitsInOrder([
+        Result<String>.loading(),
+        Result.success('id'),
+        Result.success('id').asLoading(),
+        Result.success('id'),
+      ]),
+    );
+    // The delay is necessary to avoid a race condition with loading twice at the same time
+    await Future<void>.delayed(const Duration(milliseconds: 1));
```
provokateurin, line 156: @Leptopoda I saw you simply used expectLater without awaiting it in https://github.com/nextcloud/neon/pull/1500, I guess I need to do the same thing here

provokateurin: Nope wrong, the delay is really necessary to avoid the race condition.

packages/neon_framework/lib/src/blocs/user_status.dart
```diff
@@ -41,16 +74,33 @@ class _UserStatusBloc extends InteractiveBloc implements UserStatusBloc {
...
-    await Future.wait(statuses.value.keys.map((username) => load(username, force: true)));
+    await Future.wait([
+      RequestManager.instance.wrapNextcloud(
+        account.id,
+        'user_status-predefined-statuses',
+        predefinedStatuses,
+        account.client.userStatus.predefinedStatus.findAllRaw(),
+        (response) => response.body.ocs.data,
+      ),
+      ...statuses.value.keys.map((username) => load(username, force: true)),
```
Leptopoda, line 102: This is really dangerous.
The load function is asynchonous so the statuses subject can change between calling and execution.
Please just map over the entries and give the load function both the username and result (similar to `updateStatus`) or the MapEntry of them.

This cleans up a lot of the checks you needed to do there.

packages/neon_framework/test/user_status_bloc_test.dart
Leptopoda, line 1: I'm not conviced by your approach.
By only mocking the HttpClient you also run through the RequestManager. It should be possible to write the tests in a way that is independent of it without also unit testing this part.

I'll make some changes and send them to you. Should make the tests a lot cleaner.

Leptopoda (commented):
If you want feel free to send the `clearAt` stuff as a separate PR as this looks good

packages/neon_framework/test/user_status_bloc_test.dart
provokateurin: Well I had the problem with the nested clients which got ugly really fast. Please send me the patch, I'm interested :)

packages/neon_framework/lib/src/blocs/user_status.dart
provokateurin: I understand what you mean, but I'm not sure if I understand how you would like to see it work.

packages/neon_framework/lib/src/bloc/result.dart
```diff
@@ -1,6 +1,12 @@
 import 'package:flutter/widgets.dart';
 import 'package:rxdart/rxdart.dart';
 
+/// An extension that calls [Result.transform] on every [Result] emitted.
+extension ResultStreamTransform<T> on Stream<Result<T>> {
+  /// Transforms every [Result] in the `Stream` using [Result.transform].
+  Stream<Result<R>> transformResult<R>(R Function(T data) callback) => map((e) => e.transform(callback));
+}
+
```
Leptopoda, line 9: Not used in production so either `@visibleForTesting` or move it into the test itself

packages/neon_framework/lib/src/models/account.dart
```diff
@@ -32,13 +33,15 @@ class Account implements Credentials, Findable {
     required this.username,
     this.password,
     this.userAgent,
+    Client? httpClient,
```
Leptopoda, line 36: Please `@visibleForTesting`

packages/neon_framework/lib/src/blocs/user_status.dart
```diff
@@ -89,18 +142,103 @@ class _UserStatusBloc extends InteractiveBloc implements UserStatusBloc {
...
+      if (predefinedStatus != null) {
+        correctedResult = correctedResult.rebuild(
+          (b) => b
+            ..message = predefinedStatus.message
+            ..icon = predefinedStatus.icon,
+        );
+      }
+    }
+
+    return correctedResult;
```
Leptopoda, line 170:
```suggestion
    if (status is user_status.Private && status.messageIsPredefined) {
      final id = status.messageId;
      final predefinedStatus =
          predefinedStatuses.valueOrNull?.data?.firstWhereOrNull((predefined) => predefined.id == id);

      if (predefinedStatus != null) {
        return status.rebuild(
          (b) => b
            ..message = predefinedStatus.message
            ..icon = predefinedStatus.icon,
        );
      }
    }

    return status;
```

Sorry for the back and forth

Leptopoda (changes requested):
two things:

1. please try to get a higher coverage in for the user status bloc
	- return a 204 status code (should retry)
	- return a not 204 status code (should emit error)
	- the `if (!force && !result.hasError) {` check
	- `correctStatus` with a predefined status set
	- whatever the line with `find(userId: username);` 
	
I mean you are able to read a coverage report :P

2. On a fresh app start/hot restart the status remains in the loading state forever.

Leptopoda (approved):

PR #1386, opened 2023-12-29
fix(dynamite): test documentation of generated libraries

Leptopoda:
Currently depends on the next release of code_builder.
Once it is done I'll regenerate the specs for every commit and drop the current dependency overrides.

pana output for nextcloud: `7701 out of 20941 API elements (36.8 %) have documentation comments.`


provokateurin:
Would it be possible to already separate and merge some of the commits?

Leptopoda:
I think all other than 885b6c1 (and the documentation unit test) would already be possible.
Do you prefer single PRs per commit or one with al changes?

provokateurin:
Just group them in a way you think makes sense :)

Leptopoda:
looks like we'll get a release really soon
https://github.com/dart-lang/code_builder/pull/445

packages/dynamite/dynamite/example/lib/petstore.openapi.dart
```diff
@@ -1,10 +1,25 @@
...
+///
+/// You can contact the Swagger API Team team under:
+///   Email: `apiteam@swagger.io`.
+///   Website: `http://swagger.io`.
+///
+/// Use of this source code is governed by a Apache 2.0 license.
+/// It can be obtained at https://www.apache.org/licenses/LICENSE-2.0.html.
+///
+/// Usage of these apis must adhere to the terms of service: `http://swagger.io/terms/`.
+library; // ignore_for_file: no_leading_underscores_for_library_prefixes
```
Leptopoda, line 21: This is ugly. I'll open an issue (maybe PR) upstream.
But not a blocker for this PR.

Leptopoda:
just a rebase for now.
@provokateurin our current dynamite code does not allow easily documenting interfaces or typedefs. Should I just add a local ignore for them?

provokateurin:
> our current dynamite code does not allow easily documenting interfaces or typedefs. Should I just add a local ignore for them?

Can you explain to me why? I'm fine with an ignore if there is nothing better right now.

Leptopoda:
1. only direct interfaces have doc comments. All interfaces generated by AllOfs can't be documented.
2. adding them to top level typedefs would require adding a docs field to the TypeResult that is only used by the typedefs.
I don't like having it in there as it would only further clutter the typeresults

packages/dynamite/dynamite/example/lib/petstore.openapi.dart
Leptopoda: Reported upstream https://github.com/dart-lang/code_builder/pull/448

provokateurin:
Now we need to add a latin dictionary :joy:

packages/dynamite/dynamite/example/lib/petstore.openapi.dart
provokateurin: Shall we wait for it?

packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
```diff
@@ -21,11 +21,15 @@ Iterable<Spec> generateSchemas(
 
       // TypeDefs should only be generated for top level schemas.
       if (result is TypeResultBase || result.isTypeDef) {
-        yield TypeDef(
-          (b) => b
+        yield TypeDef((b) {
+          if (!(state.buildConfig.analyzerIgnores?.contains('public_member_api_docs') ?? false)) {
+            b.docs.add('// ignore: public_member_api_docs');
+          }
```
provokateurin, line 27: Is it required to avoid duplicate ignores?

provokateurin (changes requested):
The links to the licenses should be quoted to avoid the final dot being part of the link.

packages/dynamite/dynamite/lib/src/builder/generate_schemas.dart
Leptopoda: yep

provokateurin (approved):

PR #1392, opened 2023-12-29
refactor(dynamite)!: do not name generated libraries

Leptopoda:


provokateurin (approved):
If I see this correctly the library directive is now completely omitted if it has no annotation? Would at least make sense since it is useless in that case

Leptopoda:
Yep. And once code_builder 4.10.0 comes with support for doc comments the library will not be omitted when having docs.

Issue #1154, opened 2023-11-17
Find and test naming collisions and string escaping within dynamite


Leptopoda:
continuation of #1099

Leptopoda:
Issues:
- [ ] `Object` is considered a valid class name. This will not produce a compile time error but is fairly inconvenient to use.

provokateurin:
Enum generation ignores keywords in class name:
```json
"String": {
    "type": "string",
    "enum": [
        "a",
        "b",
        "c"
    ]
},
```
leads to
```dart
class String extends EnumClass {
  const String._(super.name);

  static const String a = _$stringA;

  static const String b = _$stringB;

  static const String c = _$stringC;

  static BuiltSet<String> get values => _$stringValues;

  static String valueOf(final String name) => _$valueOfString(name);

  static Serializer<String> get serializer => _$stringSerializer;
}
```
This might apply to normal object generation as well.

Leptopoda:
We should also document that we require class names to be unique when parsed to be camelCase.

There is no way we can reasonably support both a  'my-class' and 'MyClass' in the same spec.

provokateurin:
Interfaces getters need to be escaped as well (otherwise built_value generates broken code).

Leptopoda:
since https://github.com/nextcloud/neon/pull/1276 we can conflict with our serializers.

Leptopoda:
RegExp is another class name that need sescaping

PR #1400, opened 2023-12-30
Feat/dynamite/reduce-name-collisions

Leptopoda:
fixes: #1154

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/serializer.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/serializer.dart
@@ -5,15 +5,22 @@ List<Spec> buildSerializer(final State state) => [
       const Code('// coverage:ignore-start\n'),
       Field((final b) {
         b
-          ..docs.add('''
+          ..docs.add(r'''
 /// Serializer for all values in this library.
 /// 
 /// Serializes values into the `built_value` wire format.
-/// See: [jsonSerializers] for serializing into json.''')
+/// See: [$jsonSerializers] for serializing into json.''')
           ..annotations.add(refer('visibleForTesting'))
           ..modifier = FieldModifier.final$
           ..type = refer('Serializers')
-          ..name = 'serializers';
+          ..name = r'$serializers'
+          ..assignment = const Code(r'_$serializers');
+      }),
+      Field((final b) {
+        b
+          ..modifier = FieldModifier.final$
+          ..type = refer('Serializers')
+          ..name = r'_$serializers';
 
         final serializers =
             state.resolvedTypes.map((final type) => type.serializers).expand((final element) => element).toSet();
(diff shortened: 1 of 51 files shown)


packages/dynamite/dynamite/lib/src/builder/serializer.dart
```diff
@@ -5,15 +5,22 @@ List<Spec> buildSerializer(final State state) => [
...
           ..type = refer('Serializers')
-          ..name = 'serializers';
+          ..name = r'$serializers'
+          ..assignment = const Code(r'_$serializers');
+      }),
+      Field((final b) {
+        b
+          ..modifier = FieldModifier.final$
+          ..type = refer('Serializers')
+          ..name = r'_$serializers';
```
provokateurin, line 23: Why have a public and a private variable that are the same? One public one with the $ prefix is enough

packages/dynamite/dynamite/lib/src/builder/serializer.dart
```diff
@@ -31,15 +38,22 @@ List<Spec> buildSerializer(final State state) => [
...
           ..type = refer('Serializers')
-          ..name = 'jsonSerializers';
+          ..name = r'$jsonSerializers'
+          ..assignment = const Code(r'_$jsonSerializers');
+      }),
+      Field((final b) {
+        b
+          ..modifier = FieldModifier.final$
+          ..type = refer('Serializers')
+          ..name = r'_$jsonSerializers';
```
provokateurin, line 56: Same as above

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/builder/serializer.dart
Leptopoda: Nope it isn't:
In the method body of an operation we need private variables.
Imagine an operation with a parameter called `jsonSerializers`:

```dart
  DynamiteRawResponse<JsonObject, void> $getRaw({
    String? jsonSerializers,
  }) {
    final _parameters = <String, dynamic>{};
    const _headers = <String, String>{
      'Accept': 'application/json',
    };

    final $jsonSerializers = jsonSerializers.serialize( // need a private serializer here
      contentString,
      specifiedType: const FullType(ContentString, [
        FullType(BuiltMap, [FullType(String), FullType(JsonObject)]),
      ]),
    );
    _parameters['json-serializers'] = $jsonSerializers;
    
    // ...
  }
```

provokateurin: But something like
```dart
    final _$jsonSerializers = $jsonSerializers.serialize( // need a private serializer here
      jsonSerializers,
      specifiedType: const FullType(String),
    );
```
should work.
The serialized name just needs to be prefixed with underscore.

Leptopoda: but then a parameter called `headers` or `path` (we already faced the later one once) would conflict with our already private variables.

Yes we can prefix them with `_$` but I don't know if we should rework such a major thing now.

provokateurin: Alright then

provokateurin (changes requested):
Please make the public serializers getters pointing to the private serializers.

Leptopoda:
I'll add an e2e test for both collision types we discussed above. Just in case we forget this in the future.
 
 > Please make the public serializers getters pointing to the private serializers.

What do you mean?

provokateurin:
> What do you mean?

```dart
Serializers get $jsonSerializers => _$jsonSerializers;
final Serializers _$jsonSerializers = (_$serializers.toBuilder()
```

Now that I think about it, this isn't really fixing the problem and rather working around it. In case there is a parameter called `jsonSerializers` the `$jsonSerializers` will be declared twice. Of course it works, but it's rather ugly imo

Leptopoda:
Yes we rely on the parameter shadowing the library level variable.
Any better ideas?

provokateurin:
Other than what I said above, no. Go ahead then

Leptopoda:
If you want to change it please open an issue. I'm not against prefixing our stuff with a `$` and the serialized variables with a `_`.

I'd just like to think more about the implications and not rush such a change with this PR.

provokateurin (approved):

Issue #1399, opened 2023-12-30
dynamite name all imports to avoid naming collisions


Leptopoda:
to avoid manually handling this we could let code_builder take over the work:
```dart
refer('Built', 'package:built_value/built_value.dart');
```

But this means that we need to also migrate code that does not yet rely on code_builder.

provokateurin:
I like that idea a lot, should definitely help us a lot!

PR #1401, opened 2023-12-30
Feat/dynamite/named-imports

Leptopoda:
currently based on #1400 (I was to lazy to rebase)
towards: #1399
also helps with: #1244 as it will only declare the import if it is referenced anywhere in the code.

remaining imports with explanation:
```dart
// din't touch dart imports yet.
dart:convert
dart:typed_data
// Typeresults need bigger refactors to make this work
package:built_collection/built_collection.dart
package:dynamite_runtime/models.dart
package:built_value/json_object.dart
// Needs bigger refactors but I have a branch from a while back
// when I already tried to tackle it
package:built_value/serializer.dart
// only exports extension methods so making it does not help
// Maybe I'll make it named anyways so we can't conflict with the extension name
package:collection/collection.dart
// Will never be possible as the built_value_generator would fail to run
package:built_value/built_value.dart
```


packages/dynamite/dynamite/lib/src/helpers/version_checker.dart
Leptopoda, line 1: Also needs to be removed from the readme

Leptopoda:
@provokateurin should be ready now

provokateurin (approved):
This should completely close https://github.com/nextcloud/neon/issues/1399, right?

Leptopoda:
Nope.

Some imports need bigger refractors and will come in the future

provokateurin:
Ok but in the last commit you completely remove the old imports, that seems wrong if not all are migrated yet?

Leptopoda:
I should have done a separate PR for that one sorry.
The last commit is just unused code. A leftover from before we generated a library for every spec.

Should I make a separate PR?

provokateurin:
Yes please

Issue #1128, opened 2023-11-08
Support builtlist in sortbox


Leptopoda:
We should consider supporting `ListBuilder`s in sortbox so someone can use `builtList.rebuild((b) => b.sort())` istead of `(builtList.toList()..sortByPriority()).toBuiltList()`

PR #1410, opened 2024-01-01
perf: avoid unwrapping BuiltList into a List

Leptopoda:
This should bring a minimal performance improvement and also helps with unwanted side effects.
once we have #1128 we should be able to further keep the lists immutable.


provokateurin (approved):
Please set a better PR title next time

PR #1420, opened 2024-01-03
perf(dynamite): Only initialize clients with tag once

provokateurin:
Previously they were initialized every time the were used. Now it's only once and also only when needed by using late.


packages/dynamite/dynamite_end_to_end_test/test/tags_test.dart
```diff
@@ -0,0 +1,11 @@
+import 'package:dynamite_end_to_end_test/tags.openapi.dart';
+import 'package:test/expect.dart';
+import 'package:test/scaffolding.dart';
+
+void main() {
+  test('Nested clients', () {
+    final client = $Client(Uri());
+    expect(client.first, isA<$FirstClient>());
+    expect(client.second, isA<$SecondClient>());
+  });
```
Leptopoda, line 10: So we can be sure they really are identical.

```suggestion
  test('Nested clients', () {
    final client = $Client(Uri());
    expect(client.first, isA<$FirstClient>());
    expect(identical(client.first, client.first), isTrue);

    expect(client.second, isA<$SecondClient>());
    expect(identical(client.second, client.second), isTrue);
  });
```

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -119,15 +119,14 @@ super(
         for (final tag in tags) {
           final client = clientName(tag.name);
 
-          b.methods.add(
-            Method(
+          b.fields.add(
+            Field(
```
Leptopoda, line 123: Is there a reason they are not marked as final?

I don't think there is a reason we should allow changing the value later on.

Leptopoda (changes requested):

packages/dynamite/dynamite/lib/src/builder/client.dart
provokateurin: Nope, just forgot it

PR #1422, opened 2024-01-03
feat(sort_box): support sorting a ListBuilder

Leptopoda:

fixes: #1128

### DIFF
--- a/packages/sort_box/lib/src/sort_box.dart
+++ b/packages/sort_box/lib/src/sort_box.dart
@@ -1,3 +1,4 @@
+import 'package:built_collection/built_collection.dart';
 import 'package:meta/meta.dart';
 
 /// Signature of a function returning a [Comparable].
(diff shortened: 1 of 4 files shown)


packages/sort_box/lib/src/sort_box.dart
```diff
@@ -51,6 +54,32 @@ class SortBox<T extends Enum, R> {
...
+      return;
+    }
+
+    final boxes = {
+      ...?presort,
+      box,
+      ...?this.boxes[box.property],
+    };
+
+    input.sort((item1, item2) => _compare(item1, item2, boxes.iterator..moveNext()));
```
provokateurin, line 80: I don't like that this is completely duplicate with the existing method (but I know it won't work because the sort methods are not the same. How about calling a common method and give it `input.sort` as a callback?

provokateurin (changes requested):

packages/sort_box/lib/src/sort_box.dart
Leptopoda: I think this would only make it more complex and ugly.
With the entire package being only 100 LOC we should consider leaving this duplication for now.

provokateurin: Ok

provokateurin (approved):

PR #1433, opened 2024-01-09
refactor(neon_framework): Move save file dialog to platform

provokateurin:
Required for implementing web (which doesn't support either of the current implementations).


packages/neon_framework/lib/src/platform/platform.dart
```diff
@@ -82,11 +83,9 @@ abstract interface class NeonPlatform {
...
 
-  /// Whether this platform should use file dialog.
-  ///
-  /// This is needed to compensate lacking support of `https://pub.dev/packages/file_picker`.
-  abstract final bool shouldUseFileDialog;
-
   /// Initializes this platform.
   FutureOr<void> init();
+
+  /// Saves a file with the user choosing the location.
```
Leptopoda, line 89: Should also mention what the returned String is doing

Leptopoda (commented):
I'm not that sure if this should be in the NeonPlatform,
All other dialogs the framework exposes are just through methods like `showConfirmationDialog()`. Maybe we should add a similar method that intern uses the information provided by the NeonPlatform. 

The current way fragments the way a consumer of the framework shows the dialogs.

provokateurin:
Ok we can also go that way, but then the we need to make this platform option an enum to allow other implementations (e.g. for web).

Leptopoda:
I changed my mind. Having it in here (NeonPlatform) allows us to better mock it when needed.

provokateurin:
Good point, I didn't even think about that.

PR #1444, opened 2024-01-12
fix(nextcloud): Fix flaky dashboard tests

provokateurin:
The dashboard seems to have a bug where in some cases no items are returned. This is not our fault so we have to work around it :/


Leptopoda (approved):
Should we report this upstream?
Or did you already do so?

provokateurin:
It's hard to report it when at least I have no clue what is going on and there is no way to reproduce it really.

Issue #1307, opened 2023-12-15
Switch mockito usage to mocktail


provokateurin:
Currently we use both mockito and mocktail in the project, but we consider mocktail as better because it doesn't need to generate thousands of lines of code to work.

provokateurin:
mockito also includes it's version number in the generated code so our dependency constraint would need to pin an exact version all the time (would be fine though as it is a dev_dependency).

Issue #788, opened 2023-09-18
Configure HttpClient


Leptopoda:
- add  follow redirects (discuss the security implications!)
- add keepalive (persistent connection)

Leptopoda:
Looks like the `persistentConnection` is only a header indicating whether a persistent connection is kept.
The property to change would be `idleTimeout` wich already defaults to a reasonable 15 seconds.

I think this is enough but @provokateurin do cou think we should raise it nevertheless? 


https://api.flutter.dev/flutter/dart-io/HttpClient/idleTimeout.html
https://api.flutter.dev/flutter/dart-io/HttpHeaders/persistentConnection.html

provokateurin:
I think that is enough, we also don't want people to wait forever while the network is not working at all.

The other points should still be investigated.

provokateurin:
Shall we switch to cronet_http_embedded now?

Leptopoda:
In theory yes but we should not rush it.

1. the embedded version is probably going to be deprecated in the long run while the normal version is going to take a compile time option to distinguish between both. Not a stopper for us as migration would be straight forward.

2. With cronet using ffi to communicate to the native lib we loose the benefits of streams. While they expose streaming interfaces all calls are collected into a single buffer copied over to the native code and then sent (similar for receiving). This would not be a big issue for the api requests but with large webdav files we might hit memory limitations. 
Don't quote me on this point as I haven't had the time to dig into this deeper and do actual tests and I might have misunderstood some code I read  a few weeks ago.

provokateurin:
Oh, no streams support is kind of a deal breaker. Users should be able to download a 4GB file on a device with only 2GB RAM.

Leptopoda:
to my knowledge webdav isn't really benefiting from http2 or http3 as there probably aren't many concurrent dowloads ever happening.
We could just use dart:io client for webdav.

Leptopoda:
Moving to the neon_framework as dynamite should accept any client

provokateurin:
Anything that still needs to be done?

Leptopoda:
I still want to use a client that supports http2 or even http3.
On Android my favorite would be the `cronet` client, but that either relies on the play services or bundling it ourselves would increase the apk size by roughly 14MB.
A different approach would be the `ok_http` client. It only supports http2 but doesn't bloat the apk size (https://github.com/dart-lang/http/tree/master/pkgs/ok_http).

provokateurin:
[...]

PR #1448, opened 2024-01-12
feat(dynamite)!: allow clients from the http package

Leptopoda:

towards: #788 
fixes: #1307 

Sorry for the big PR but it couldn't be done in smaller ones.
The fixtures now only include the headers explicitly added and not the ones added by the cleint itself.

I haven't checked every edge case but regarding the NC code (without the reexported dynamite_runtime) the only breaking change should be that the `httpClient` parameter now takes a `Client` instead of an `HttpClient`.


packages/nextcloud_test/pubspec.yaml
```diff
@@ -8,8 +8,8 @@ environment:
 dependencies:
   built_collection: ^5.0.0
   dynamite_runtime: ^0.1.0
+  http: ^1.1.0
```
provokateurin, line 11: 1.0.0 should also work (please test with pub downgrade).

packages/dynamite/dynamite_end_to_end_test/test/authentication_test.dart
```diff
@@ -31,172 +16,262 @@ void main() {
...
+        expect(request.bodyBytes.length, 0);
+        expect(request.headers, equals({'Accept': 'application/json'}));
+
+        return Response('{}', 200);
+      }),
+    );
     await client.noAuthentication();
-    var captured = verify(headers.add(captureAny, captureAny)).captured;
-    expect(captured, equals(['Accept', 'application/json']));
+    //var captured = verify(headers.add(captureAny, captureAny)).captured;
```
provokateurin, line 32: Remove

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -397,8 +397,11 @@ class DynamiteClient {
   /// The base headers added to each request.
   final Map<String, String>? baseHeaders;
 
+  /// The value sent in the `user-agent` header.
+  final String? userAgent;
```
provokateurin, line 401: Let's remove this field, it can simply be passed with the baseHeaders

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -425,33 +428,42 @@ class DynamiteClient {
...
     }
 
-    final response = await request.close();
-    if (cookieJar != null) {
-      await cookieJar!.saveFromResponse(uri, response.cookies);
+    final response = await httpClient.send(request);
+
+    final cookieHeader = response.headers['set-cookie'];
+    if (cookieHeader != null && cookieJar != null) {
+      final cookies = cookieHeader.split(RegExp(',(?=[^ ])')).map(Cookie.fromSetCookieValue).toList();
```
provokateurin, line 465: This looks dangerous, does it follow some standard on how to split cookies or did you copy this implementation from somewhere else where it is known to work?

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -425,33 +428,42 @@ class DynamiteClient {
...
+      unawaited(cookieJar!.saveFromResponse(uri, cookies));
```
provokateurin, line 466: Should be awaited to ensure queued requests always use the correct requests.

packages/dynamite/dynamite_runtime/pubspec.yaml
```diff
@@ -16,14 +16,13 @@ dependencies:
   built_value: ^8.0.1
   collection: ^1.0.0
   cookie_jar: ^4.0.7
+  http: ^1.1.0
```
provokateurin, line 19: 1.0.0 should also work

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -20,39 +55,34 @@ class WebDavClient {
...
+      method,
+      url,
+      data: data,
+      dataStream: dataStream,
+      headers: headers,
+    );
 
     request.headers.addAll({
-      HttpHeaders.contentTypeHeader: 'application/xml',
+      if (rootClient.userAgent != null) HttpHeaders.userAgentHeader: rootClient.userAgent!,
```
provokateurin, line 74: The userAgent should be dropped here as well

packages/nextcloud/pubspec.yaml
```diff
@@ -18,7 +18,8 @@ dependencies:
   crypto: ^3.0.0
   crypton: ^2.0.0
   dynamite_runtime: ^0.1.0
-  intl: ">=0.17.0 <=0.19.0"
+  http: ^1.1.0
```
provokateurin, line 21: Also 1.0.0

packages/nextcloud/test/webdav_test.dart
```diff
@@ -305,7 +305,7 @@ void main() {
 
       test('Filter files', () async {
         final response = await client.webdav.put(utf8.encode('test'), PathUri.parse('filter.txt'));
-        final id = response.headers['oc-fileid']!.first;
+        final id = response.headers['oc-fileid']!.split(RegExp(r'\s*,\s*')).first;
```
provokateurin, line 308: What's the purpose of this? I think you got confused by the multiple header thing. There will only ever one fileid that is returned, so you can directly take that.

provokateurin (changes requested):
Looks really good!

packages/nextcloud/test/webdav_test.dart
Leptopoda: Yes I got confused. For these cases we should have used `single` over `first` to make that clear.

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: The api of this package is garbage (it should use FutureOr).
Cookies are not persisted on web (the browser does that for us) so it will always use the `DefaultCookieJar` which implements this in a synchronous way.
I'd like not to interrupt the execution here so I wrapped it in unawaited.

Ideally the interface would be FutureOr with both default implementations implementing this synchronous. If I feel like it I'll open an Issue/PR upstream but I doubt that they'd respond in a reasonable time.

packages/nextcloud/test/webdav_test.dart
provokateurin: True, but you need neither of those now as you can just take the complete header value.

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
provokateurin: But for now we should still await it.

Do you think it would make sense to write our own implementation?

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
provokateurin: Neat :D

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: The code is really complex. I'm against writing our own for now.

And awaiting it will interrupt the execution (async functions always do so. Even if they complete synchronously).

provokateurin: But we need to await it to ensure the integrity of the cookies? Or am I missing something?

Leptopoda:
making this a draft as there is no timeline yet for the next http package release

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -3,15 +3,50 @@ import 'dart:convert';
...
+  WebDavRequest(
+    super.method,
+    super.url, {
+    this.dataStream,
+    this.data,
+    Map<String, String>? headers,
+  }) : assert(dataStream == null || data == null, 'Only one of dataStream or data can be specified.') {
+    this.headers.addAll({
+      HttpHeaders.contentTypeHeader: 'application/xml',
+      ...?headers,
```
provokateurin, line 27: Shouldn't the content type be last so it doesn't get overridden accidentally?

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -210,24 +239,24 @@ class WebDavClient {
     File file, {
     void Function(double progress)? onProgress,
   }) async {
-    final sink = file.openWrite();
     final response = await getStream(path);
-    if (response.contentLength > 0) {
+    final contentLength = response.contentLength;
+    if (contentLength != null && contentLength > 0) {
+      final sink = file.openWrite();
```
provokateurin, line 245: If the response is empty, we still want to write into the file.

packages/nextcloud/test/webdav_test.dart
```diff
@@ -6,7 +6,7 @@ import 'package:nextcloud/nextcloud.dart';
 import 'package:nextcloud_test/nextcloud_test.dart';
 import 'package:test/test.dart';
 import 'package:test_api/src/backend/invoker.dart';
-import 'package:universal_io/io.dart';
+import 'package:universal_io/io.dart' hide HttpClient;
```
provokateurin, line 9: Unnecessary hide

packages/dynamite/dynamite_end_to_end_test/pubspec.yaml
```diff
@@ -18,10 +18,9 @@ dev_dependencies:
   build_runner: ^2.4.8
   built_value_generator: ^8.8.1
   dynamite: ^0.1.0
-  mockito: ^5.4.4
+  http: ^1.1.2
```
provokateurin, line 21:
```suggestion
  http: ^1.2.0
```

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -425,33 +424,41 @@ class DynamiteClient {
     return executeRawRequest(
       method,
       uri,
-      {...?baseHeaders, ...headers},
+      {
+        ...?baseHeaders,
+        ...headers,
```
provokateurin, line 429: Should we change the order to ensure the base headers are not overridden or do we consider it as a feature?

provokateurin (changes requested):
The dependency check for the http package is missing in dynamite

packages/nextcloud/test/webdav_test.dart
Leptopoda: I intentionally added it so we do not accidentally use it over the http package.
I wanted to show that the universal_io import should only be used for the `file`  features.

Should I change this?

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: I think that a header specified in the specs should have higher priority.
If an operation has a `useragent` header parameter it should be used over the client one.

packages/nextcloud/test/webdav_test.dart
provokateurin: Yeah, I don't think it really matters.

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
provokateurin: Sounds good.

packages/nextcloud/lib/src/webdav/client.dart
Leptopoda: let's make a separate PR.
I just carried over the old approach

provokateurin: But let's fix it later if needed.

provokateurin (approved):

PR #1452, opened 2024-01-13
refactor(tool)!: Pin our own version of openapi-extractor

provokateurin:
It allows us to get improvements and fixes from openapi-extractor much quicker. This is especially helpful since there are multiple repos we get the specs from and each of those would need to individually update their version.
It was like this some time ago and I removed it to avoid running into compatibility problems, but I don't think any problems will appear in the future, so it should be safe to track our own version again.

### DIFF
--- a/packages/nextcloud/lib/src/api/theming.openapi.json
+++ b/packages/nextcloud/lib/src/api/theming.openapi.json
@@ -163,7 +163,11 @@
                         "description": "Let the browser decide the CSS priority",
                         "schema": {
                             "type": "integer",
-                            "default": 0
+                            "default": 0,
+                            "enum": [
+                                0,
+                                1
+                            ]
                         }
                     },
                     {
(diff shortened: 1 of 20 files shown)


packages/nextcloud/lib/src/api/theming.openapi.json
```diff
@@ -629,151 +629,6 @@
                 }
             }
         },
-        "/index.php/apps/theming/background/{type}": {
```
Leptopoda, line 632: Why did this operation get removed?

Leptopoda (approved):
LGTM

packages/nextcloud/lib/src/api/theming.openapi.json
provokateurin: There were some changes in the detection of some routes and this was previously detected as available, but in reality it is not. This isn't breaking as the endpoint would have never worked anyway.

PR #1460, opened 2024-01-16
feat(neon_framework): Add emoji picker dialog

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/1378 and added a test so it can be merged already.


packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -523,3 +524,38 @@ class NeonUnifiedPushDialog extends StatelessWidget {
         ],
       );
 }
+
+/// Shows an emoji picker.
+///
+/// When the user selects an emoji the dialog will pop and return the emoji.
```
Leptopoda, line 530:
```suggestion
/// When the user selects an emoji the dialog will pop and return the emoji as a `String`.
```

Leptopoda (approved):
Small nitpick but LGTM otherwise

PR #1465, opened 2024-01-16
fix(neon_framework): warning in the dialog unit tests 

Leptopoda:
When hitting enter on the keyboard the dialog should (and does) automatically submit the result and pop.
We do not need to tap the save action anymore. Doing so produces a warning that the save action is no longer present on the screen.

### DIFF
--- a/packages/neon_framework/test/dialog_test.dart
+++ b/packages/neon_framework/test/dialog_test.dart
@@ -128,7 +128,6 @@ void main() {
         await widgetTester.pumpAndSettle();
         await widgetTester.enterText(find.byType(TextFormField), 'My new value');
         await widgetTester.testTextInput.receiveAction(TextInputAction.done);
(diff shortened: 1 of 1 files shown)


provokateurin:
Wrong PR name and the description should mention the relevant issue

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1459 (this is what I meant with the relevant issue :sweat_smile:)

provokateurin (approved):

Leptopoda:
> Fixes #1459 (this is what I meant with the relevant issue 😅)

Sorry I did not know we had an issue for this. I just found the bug while reviewing your PR (where you probably also found the issue).

provokateurin:
No problem, I thought you fixed it because you saw the issue :see_no_evil:

Issue #446, opened 2023-07-05
Huawei sqflite bug


Leptopoda:
I was going to test the performance on a few of my test devices and started getting:

```
E/SQLiteLog(11817): (1)
I/flutter (11817): DatabaseException(near "ON": syntax error (Sqlite code 1 SQLITE_ERROR): , while compiling: INSERT INTO cache (key, value) VALUES (?, ?) ON CONFLICT(key) DO UPDATE SET value = excluded.value, (OS error - 2:No such file or directory)) sql 'INSERT INTO cache (key, value) VALUES (?, ?) ON CONFLICT(key) DO UPDATE SET value = excluded.value' args [84d5a5767ba3dd1009d002da41c57487c65eedbc-capabilit..., ["ocs",["meta",["status","ok","statuscode",200,"me...]
I/flutter (11817): #0      wrapDatabaseException
exception_impl.dart:11
I/flutter (11817): <asynchronous suspension>
I/flutter (11817): #1      SqfliteDatabaseMixin.txnRawQuery.<anonymous closure>
database_mixin.dart:559
I/flutter (11817): <asynchronous suspension>
I/flutter (11817): #2      BasicLock.synchronized
basic_lock.dart:33
I/flutter (11817): <asynchronous suspension>
I/flutter (11817): #3      SqfliteDatabaseMixin.txnSynchronized
database_mixin.dart:490
I/flutter (11817): <asynchronous suspension>
I/flutter (11817): #4      RequestManager._wrap
request_manager.dart:102
I/flutter (11817): <asynchronous suspension>
I/flutter (11817): #5      CapabilitiesBloc.refresh
capabilities.dart:43
I/flutter (11817): <asynchronous suspension>
I/flutter (11817):
Lost connection to device.
```

Looks like we are hitting an upstream issue of sqflite:
https://github.com/tekartik/sqflite/issues/1023

With the little number of huawei users in the western world (where Id argue nextcloud has most active users) I don't think it is a high priority for us.

provokateurin:
Good to know, but since it's an upstream bug we can't do much about it

provokateurin:
Maybe we should wrap the exception and just don't interact with the cache? Then it's not broken for users (but of course they won't have any caching).

Leptopoda:
:+1:  It's better than nothing.

Maybe we can also find an alternative query as this is the only query that has failed (it must have been the insert one)

Leptopoda:
I can also reproduce this on a motoG5.
[...]

PR #1466, opened 2024-01-16
fix(neon_framework): change sqflite queries to work on more devices

Leptopoda:
fixes: #446


provokateurin (approved):
Interesting trick, I didn't know the problem was an old sqlite version on those devices.

provokateurin:
I don't know :/

PR #1469, opened 2024-01-17
chore(neon_framework): only depend on one sqlite package

Leptopoda:


provokateurin (approved):
I didn't know that you could remove the sqflite one if you have sqflite_common_ffi :+1:

PR #1488, opened 2024-01-21
fix(neon_framework): Add missing drawer while loading

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/484
I am pretty sure that the drawer also wasn't updated correctly, in case the cached apps list wasn't the same as the newly loaded apps list, but that was hard to trigger since the drawer state was only initialized when the drawer was opened and the loading was always faster.


Leptopoda (commented):
This should be fairly easy to unit test. You'll only need to mock the appsBloc.
Can you please add some?

Leptopoda (approved):
Thanks :+1:

PR #1490, opened 2024-01-21
Fix quick actions not working

provokateurin:
Fix https://github.com/nextcloud/neon/issues/1454
The problem was that the active app was overridden after it was set by the quick action shortcut.
To prevent it we need to skip setting the active app if there is one already, but that breaks updating the active app in case the current active app is not supported anymore, so I fixed that too.


Leptopoda (commented):
Is it hard to unit test?
Code looks clean and I'm ok with merging if adding unit tests is out of scope for this.

provokateurin:
I think the whole bloc would need unit tests first which should be doable. I can add them first and then add another test with this PR if you want

Leptopoda (approved):
I don't have the capacity either so let's get it in

PR #1493, opened 2024-01-21
fix(ci): Enable self-hosted runners

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/540
Let's see if this works...


provokateurin:
Nope, tests are still failing and the runner seems to be slower than the ones provided by GH.

PR #1495, opened 2024-01-21
fix(neon_framework): Rewrite SVGs with percent dimensions to display them correctly

provokateurin:
This is quite a hack, but the flutter_svg package can not display geometries that have their dimensions declared in percent. This trick rewrites all height and width attributes to with correct values so they can still be displayed.


packages/neon_framework/lib/src/utils/image_utils.dart
```diff
@@ -0,0 +1,41 @@
...
+
+    for (final dimension in ['width', 'height']) {
+      final rootDimensionAttribute = document.xpath('/svg/@$dimension').firstOrNull as XmlAttribute?;
+      final rootDimension = double.parse(
+        rootDimensionAttribute != null ? rootDimensionAttribute.value : viewBox[dimension]!,
+      );
+
+      for (final element in document.descendantElements) {
+        final elementDimensionString = element.getAttribute(dimension);
+        if (elementDimensionString != null && elementDimensionString.contains('%')) {
```
Leptopoda, line 32:
```suggestion
        if (elementDimensionString != null && elementDimensionString.endsWith('%')) {
```

packages/neon_framework/lib/src/utils/image_utils.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:xml/xml.dart';
+import 'package:xml/xpath.dart';
+
+/// Utilities for working with images.
+class ImageUtils {
+  ImageUtils._();
+
+  /// Rewrites percent dimensions in SVGs to be absolute.
+  ///
+  /// This is required because the `flutter_svg` package can not handle percent dimensions.
```
Leptopoda, line 10: Is there an upstream issue we can link to?

packages/neon_framework/lib/src/utils/image_utils.dart
```diff
@@ -0,0 +1,41 @@
...
+  static String rewriteSvgDimensions(String data) {
+    final document = XmlDocument.parse(data);
+
+    final viewBoxAttribute = [
+      ...document.xpath('/svg/@viewBox'),
+      ...document.xpath('/svg/@viewbox'),
+    ].first as XmlAttribute;
+    final viewBoxValues = viewBoxAttribute.value.split(' ');
```
Leptopoda, line 18: Why add it to a list only to discard it a bit later? If we find a match in the we'd also save a second lookup.
```suggestion
    final viewBoxAttribute = document.xpath('/svg/@viewBox').singleOrNull ?? document.xpath('/svg/@viewbox').singleOrNull;
    final viewBoxValues = viewBoxAttribute!.value!.split(' ');
```

But I'm not sure this is worth it and your approach might be somewhat more readable.

packages/neon_framework/lib/src/widgets/image.dart
```diff
@@ -100,7 +101,10 @@ class NeonCachedImage extends StatefulWidget {
       return cached;
     }
 
-    final data = await getImage();
+    var data = await getImage();
+    try {
+      data = utf8.encode(ImageUtils.rewriteSvgDimensions(utf8.decode(data)));
+    } catch (_) {}
```
Leptopoda, line 107: I think we should use the svg hint (similar to the check we do when rendering.

packages/neon_framework/lib/src/utils/image_utils.dart
provokateurin: There are many open issues about that, I doubt it makes sense to link to them

packages/neon_framework/lib/src/utils/image_utils.dart
provokateurin: Yeah I originally did it for readability, but right now it is readable in this way too.

packages/neon_framework/lib/src/widgets/image.dart
provokateurin: For the hint we speculatively decode the content to check if it contains an SVG. Adding that would lead to decoding the same data twice and even if the parameter is sent in the widget we already know we have to decode the data so there is no point in that.
I get that it is cleaner, but it won't save us anything and will only complicate the code (but I can still do it if you want).

Leptopoda (approved):

PR #1498, opened 2024-01-22
fix(neon_framework): Disallow null values in settings

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1494
The problem was the after the dialog closes when selecting a value there is a null check which is necessary to avoid setting the value if the dialog was just canceled. This also meant that any null value that was valid and selected wouldn't be set as the new value. The only way to fix that is to disallow any null value in the settings.


provokateurin:
Replaced by https://github.com/nextcloud/neon/pull/1503

PR #1499, opened 2024-01-22
feat(nextcloud_test): Add workaround for bugged password confirmation

provokateurin:
https://github.com/nextcloud/neon/issues/1492

This is not a reasonable fix as we need cookies for some endpoints and apppasswords are much faster for authentication (and cookies help with that as well).


Leptopoda (approved):
Can you add a doc comment to the named parameter linking to the issue?
I tend to forget where these hacks come from.

Other than that LGTM

provokateurin:
I'm not intending to merge this. The problem has to be fixed in the server.

provokateurin:
The cookies aren't even a problem, it's just the app password.

PR #1500, opened 2024-01-23
feat(neon_framework): allow to mock the request manager cache

Leptopoda:
replaces: #1468 

because there is a lot of async stuff going on there is no reasonable way to unit test the different parts individually leading to such a big test.


packages/neon_framework/lib/src/bloc/result.dart
```diff
@@ -151,10 +153,16 @@ class Result<T> {
...
+      other.isLoading == isLoading &&
+      other.data == data &&
+      other.error?.toString() == error?.toString();
+
+  @override
+  int get hashCode => Object.hash(data, error?.toString(), isLoading, isCached);
 
   @override
-  int get hashCode => Object.hash(data, error, isLoading, isCached);
+  String toString() => 'Result: { data: $data, error: $error, isLoading: $isLoading, isCached: $isCached }';
```
provokateurin, line 165:
```suggestion
  String toString() => 'Result(data: $data, error: $error, isLoading: $isLoading, isCached: $isCached)';
```
I prefer to have it like the constructor would look like.

packages/neon_framework/lib/src/bloc/result.dart
```diff
@@ -135,8 +135,10 @@ class Result<T> {
   /// A result may both have an error and data.
   bool get hasData => data != null;
 
-  /// Returns whether this snapshot [hasData] and [isCached] is not true.
-  bool get hasUncachedData => hasData && !isCached;
+  /// Returns whether this snapshot is equal to a [Result.success].
+  ///
+  /// A successful result has data, has no error and is neither loading or cached.
```
provokateurin, line 140:
```suggestion
  /// A successful result has data, has no error and is neither loading nor cached.
```

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -125,36 +128,43 @@ class RequestManager {
...
     String k,
     BehaviorSubject<Result<T>> subject,
     NextcloudApiCallback<R> call,
     UnwrapCallback<T, R> unwrap,
     SerializeCallback<R> serialize,
     DeserializeCallback<R> deserialize, [
+    // ignore: avoid_positional_boolean_parameters
     bool disableTimeout = false,
     bool emitEmptyCache = false,
+    Duration timeLimit = kDefaultTimeout,
```
provokateurin, line 146: Why not call it timeout? It's much more common and this name is a bit confusing.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -181,12 +191,15 @@ class RequestManager {
...
           emitEmptyCache,
+          timeLimit,
           retries + 1,
         );
       } else {
         _emitError<T>(e, subject);
       }
     }
+
+    await cacheFuture;
```
provokateurin, line 202: Why is this necessary?

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -254,18 +274,28 @@ class RequestManager {
   Future<T> timeout<T>(
     NextcloudApiCallback<T> callback, {
     bool disableTimeout = false,
-    Duration timeout = kDefaultTimeout,
+    Duration timeLimit = kDefaultTimeout,
```
provokateurin, line 277: Again, why change it?

packages/neon_framework/test/request_manager_test.dart
```diff
@@ -0,0 +1,927 @@
...
+  tearDown(() {
+    RequestManager.instance = null;
+    Cache.instance = null;
+  });
+
+  group('Cache', () {
+    test('singleton', () {
+      expect(identical(Cache(), Cache()), isTrue);
+    });
+  });
```
provokateurin, line 30: Cache should also have tests

packages/neon_framework/test/request_manager_test.dart
```diff
@@ -0,0 +1,927 @@
...
+    group('wrap without cache', () {
+      test('successful request', () async {
+        var subject = BehaviorSubject<Result<String>>();
+
+        // ignore: unawaited_futures
+        expectLater(
+          subject.stream,
+          emitsInOrder([
+            equals(Result<String>.loading()),
+            equals(Result.success('VGVzdCB2YWx1ZQ==')),
```
provokateurin, line 75: Please do the encoding on the fly so when reading the test you can know what the value is

packages/neon_framework/test/request_manager_test.dart
```diff
@@ -0,0 +1,927 @@
...
+            emitsDone,
+          ]),
+        );
+
+        await RequestManager.instance.wrap<String, Uint8List>(
+          'clientID',
+          'key',
+          subject,
+          () async => utf8.encode('Test value'),
+          (deserialized) => base64Encode(deserialized),
```
provokateurin, line 85:
```suggestion
          (deserialized) => base64.encode(deserialized),
```

packages/neon_framework/test/request_manager_test.dart
```diff
@@ -0,0 +1,927 @@
...
+            equals(Result<String>.error('DynamiteApiException(statusCode: 500, headers: {}, body: )')),
+            emitsDone,
+          ]),
+        );
+
+        await RequestManager.instance.wrap<String, Uint8List>(
+          'clientID',
+          'key',
+          subject,
+          () async => throw const DynamiteApiException(500, {}, ''),
```
provokateurin, line 268: You should verify that this method is actually called multiple times.

provokateurin (changes requested):
Looks pretty good.
It would be great if the tests could somehow run in a loop to reduce duplicate code just with different parameters, but that would be quite complex for all these different scenarios and behaviors.
I didn't bother repeating my comments in the tests, but please adjust all occurrences.

packages/neon_framework/lib/src/utils/request_manager.dart
Leptopoda: because the method is already called like that so the parameter would shadow the method.

provokateurin: Ah, then the method should be renamed to something better

packages/neon_framework/test/request_manager_test.dart
Leptopoda: https://github.com/nextcloud/neon/pull/1466#issuecomment-1895546285

If we add tests it should be for the lowest supported sqlite version.
I'll open an issue to investigate this. Maybe we can specify a custom binary to use and add it to our test setup.

packages/neon_framework/lib/src/utils/request_manager.dart
Leptopoda: Dart calls the method timeout and the parameter timeLimit (https://api.dart.dev/stable/3.2.5/dart-async/Future/timeout.html).
I'd rather stick with this.

provokateurin: Ok sure :)

packages/neon_framework/test/request_manager_test.dart
provokateurin: This one and a few other ones are still missing

provokateurin (changes requested):

packages/neon_framework/test/request_manager_test.dart
Leptopoda: I added an explicit test for this further up.
I like to think that the other tests are to unit test the subjects. If the retry count logic got screwed I want that particular test to fail so I can easily identify the issue and not every test of the request manager.

provokateurin: Yeah that makes sense, but the test is called `throwing DynamiteException should retry` so I would expect that it also checks that it correctly retries?

Leptopoda: you are right. Before I added 97d1b92a3055b598594174be95d4b460e5a93d26 this was obvious as the subject would emit multiple loading events. Now with the global test it makes sense to remove the individual tests as the cache should not play a role in this logic.

Should I remove them (lightening the file) or just add the verifications?

provokateurin: Yes, please remove the tests :)

provokateurin:
PR title is wrong btw

Leptopoda:
> PR title is wrong btw

I also noticed that but what should I use? Technically this is mainly a fix for a few race conditions in the cache. 
Is the old `fix(neon_framework): fix race conditions in request manager` ok?

provokateurin:
For me the main thing is `test(neon_framework): unit test request manager`, but you can also use something like `Add unit tests for RequestManager and fix some race conditions`

provokateurin (approved):

PR #1505, opened 2024-01-23
refactor(dynamite_runtime,nextcloud)!: make DynamiteApiException exte…

Leptopoda:
…nd ClientException.

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
@@ -5,6 +5,7 @@ import 'dart:typed_data';
 import 'package:built_value/serializer.dart';
 import 'package:cookie_jar/cookie_jar.dart';
 import 'package:dynamite_runtime/src/http_extensions.dart';
+import 'package:dynamite_runtime/src/utils/debug_mode.dart';
 import 'package:dynamite_runtime/src/utils/uri.dart';
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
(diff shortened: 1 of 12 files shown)


packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -466,7 +442,7 @@ class DynamiteClient {
     if (validStatuses?.contains(response.statusCode) ?? true) {
       return response;
     } else {
-      throw await DynamiteApiException.fromResponse(response);
+      throw DynamiteStatusCodeException(response.statusCode);
```
provokateurin, line 445: It's usually very helpful to also see what the body (and headers) are for debugging purposes.

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -78,10 +78,8 @@ class WebDavClient {
     final response = await rootClient.httpClient.send(request);
 
     if (response.statusCode >= 300) {
-      throw DynamiteApiException(
+      throw DynamiteStatusCodeException(
         response.statusCode,
-        response.headers,
-        await response.stream.string,
       );
```
provokateurin, line 83: Same here

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: But for the body we'd need to await the entire response and there is no reason to have the code wait.

provokateurin: Please add it regardless, otherwise debugging will be a pain.

Leptopoda: Can we make it only do that in debug mode?

provokateurin: Sure

Leptopoda: How do you like this approach?
I'm not that sure myself.

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -466,7 +471,11 @@ class DynamiteClient {
     if (validStatuses?.contains(response.statusCode) ?? true) {
       return response;
     } else {
-      throw await DynamiteApiException.fromResponse(response);
+      if (kDebugMode) {
+        throw await DynamiteStatusCodeException.fromResponse(response);
+      } else {
+        throw DynamiteStatusCodeException(response.statusCode);
+      }
```
provokateurin, line 478: Can't you move this logic into the factory? That would make it easier and impossible to make a mistake when only using one of them

Leptopoda: But the method is asynchronous. So the event loop would still be interrupted for no reason.
I'll think about a better solution. Now that this is only used in debug mode we can just block the execution by synchronously waiting.

provokateurin: Ah I see

provokateurin (approved):

PR #1507, opened 2024-01-24
refactor(dynamite): do not store cookies of failed requests

Leptopoda:


provokateurin:
I don't think this is correct. Even on a failed request cookies could be updated to contain a new session. If we don't store them, then the next request might fail (not with Nextcloud, but in general).

provokateurin:
Even in that case we should store it for the same reasons. Just because the spec is wrong it doesn't mean that the cookies are not relevant.

provokateurin:
But yeah other than the cookies the request should be discarded.

Leptopoda:
I don't really understand the reasoning but ok.

provokateurin:
I think regardless of any failure we have to store cookies because they could contain an updated session. If we miss that then we might no longer have a valid session afterwards. The status code doesn't matter in this case (or if the status code was correctly documented).

Issue #1398, opened 2023-12-30
Enum generation ignores keywords in class name:


Leptopoda:
Enum generation ignores keywords in class name:
```json
"String": {
    "type": "string",
    "enum": [
        "a",
        "b",
        "c"
    ]
},
```
leads to
```dart
class String extends EnumClass {
  const String._(super.name);

  static const String a = _$stringA;

  static const String b = _$stringB;

  static const String c = _$stringC;

  static BuiltSet<String> get values => _$stringValues;

  static String valueOf(final String name) => _$valueOfString(name);

  static Serializer<String> get serializer => _$stringSerializer;
}
```
This might apply to normal object generation as well.

_Originally posted by @provokateurin in https://github.com/nextcloud/neon/issues/1154#issuecomment-1847117809_

upstream built_value limitation tracked in: https://github.com/google/built_value.dart/issues/1292

Leptopoda:
Should get fixed by
https://github.com/google/built_value.dart/pull/1297

PR #1511, opened 2024-01-24
test(dynamite): test enums with reserved names

Leptopoda:
fixes: #1398
fixes: #1398


provokateurin (approved):
You linked the same issue twice

PR #1513, opened 2024-01-24
test(neon_framework): add tests for NeonValidationTile

Leptopoda:


packages/neon_framework/test/validation_tile_test.dart
```diff
@@ -0,0 +1,43 @@
...
+  group('NeonValidationTile', () {
+    testWidgets('NeonValidationTile material', (widgetTester) async {
+      for (final state in ValidationState.values) {
+        final widget = Material(child: NeonValidationTile(title: 'title', state: state));
+        await widgetTester.pumpWidget(wrapWidget(widget));
+
+        if (state == ValidationState.loading) {
+          expect(find.byType(CircularProgressIndicator), findsOneWidget);
+        } else {
+          expect(find.byType(Icon), findsOneWidget);
```
provokateurin, line 31: Please check which Icon it is and which color it has

provokateurin (changes requested):

packages/neon_framework/test/validation_tile_test.dart
```diff
@@ -0,0 +1,66 @@
...
+    supportedLocales: NeonLocalizations.supportedLocales,
+    locale: locale,
+    home: widget,
+  );
+}
+
+void main() {
+  group('NeonValidationTile', () {
+    testWidgets('NeonValidationTile material', (widgetTester) async {
+      var state = ValidationState.loading;
```
provokateurin, line 24: It's not necessary to store the state in a variable

packages/neon_framework/test/validation_tile_test.dart
```diff
@@ -0,0 +1,66 @@
...
+      var widget = Material(child: NeonValidationTile(title: 'title', state: state));
+      await widgetTester.pumpWidget(wrapWidget(widget));
```
provokateurin, line 26: The  widget also doesn't need to be stored in a variable first since it is only used once.

provokateurin (changes requested):

packages/neon_framework/test/validation_tile_test.dart
Leptopoda: inlining the widget makes the test less readable

provokateurin (approved):

PR #1514, opened 2024-01-24
feat(nextcloud)!: Update server to 28.0.2

provokateurin:
Depends on https://github.com/nextcloud/neon/issues/1441 (then the first commit won't be breaking).

### DIFF
--- a/packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
+++ b/packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
@@ -159,6 +159,7 @@ const _reservedNames = [
   'super',
   'switch',
   'sync',
+  'Type',
   'this',
   'throw',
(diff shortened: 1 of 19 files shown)


Leptopoda (approved):

packages/dynamite/dynamite/lib/src/helpers/dart_helpers.dart
```diff
@@ -159,6 +159,7 @@ const _reservedNames = [
   'super',
   'switch',
   'sync',
+  'Type',
```
Leptopoda, line 162: Can you maybe add it to the e2e escaping test?

provokateurin: I thought about this as well, but then we should test all the keywords which I find pointless.

PR #1517, opened 2024-01-25
refactor(neon_framework): Move authorization headers to Account

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1462

In the end I needed this change for https://github.com/nextcloud/neon/issues/1125 and this way it was easy to add tests.

### DIFF
--- a/packages/neon_framework/lib/src/models/account.dart
+++ b/packages/neon_framework/lib/src/models/account.dart
@@ -117,13 +117,12 @@ class Account implements Credentials, Findable {
   /// Get the necessary `Authorization` headers for a given [uri].
   ///
   /// This method ensures no credentials are sent to the wrong server.
-  Map<String, String> getAuthorizationHeaders(Uri uri) {
-    // Only send the authentication headers when sending the request to the server of the account
-    if (uri.toString().startsWith(serverURL.toString()) && client.authentications.isNotEmpty) {
-      return client.authentications.first.headers;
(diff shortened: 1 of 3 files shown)


packages/neon_framework/lib/src/models/account.dart
```diff
@@ -114,6 +114,18 @@ class Account implements Credentials, Findable {
...
+  /// This method ensures no credentials are sent to the wrong server.
+  Map<String, String> getAuthorizationHeaders(Uri uri) {
+    // Only send the authentication headers when sending the request to the server of the account
+    if (uri.toString().startsWith(serverURL.toString()) && client.authentications.isNotEmpty) {
+      return client.authentications.first.headers;
+    }
+
+    return <String, String>{};
+  }
+
```
Leptopoda, line 128:
```suggestion
  Map<String, String>? getAuthorizationHeaders(Uri uri) {
    // Only send the authentication headers when sending the request to the server of the account
    if (uri.toString().startsWith(serverURL.toString())) {
      return client.authentications.firstOrNull?.headers;
    }

    return null;
  }
```

I would prefer null. Feel free to make the parameter for the dynamite client nullable (I wanted to change for the non raw method anyways).

provokateurin: Agreed, I only didn't do it because the request method expected a non-null value (which you fixed now).

Leptopoda (approved):

PR #1518, opened 2024-01-25
refactor(neon_framework): Migrate NeonUrlImage to use Uri

provokateurin:
Towards https://github.com/nextcloud/neon/issues/1125


Leptopoda (commented):
How safe is it to directly parse the uri?
Should we consider using `tryParse`? 

I know we should be able to trust the server but I also don't like having it throw in production.

provokateurin:
I think we just have to trust it. I agree that there is a possibility for failure, but usually it never happens (I myself have never seen it so far) and I doubt it is likely to happen.

Leptopoda (approved):
Ok :+1:

PR #1525, opened 2024-01-25
refactor(neon_framework): Make RequestManager take Account directly

provokateurin:


Leptopoda (commented):
I'm not sure if this is the best thing to do longterm.
If we want to do all our requests through the request manager (to also handle timeouts, queue, ...) we should probably also allow requests without any account.
But I think that this would only be ever needed for the login flow

provokateurin:
In case of guest access (which we don't have yet but will at some point) and the login pages we need special (semi-)temporary accounts, but they will work with accounts as well. So I think this is not really a problem here.

Leptopoda:
In this case I'd rather have a null value passed to the request manager and it figuring out a different cache key.

Leptopoda (approved):

provokateurin:
You're right null makes more sense for those cases.

PR #1529, opened 2024-01-26
refactor(neon_framework): Make all RequestManager arguments named

provokateurin:


packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -80,71 +80,70 @@ class RequestManager {
...
-    NextcloudApiCallback<R> call,
-    UnwrapCallback<T, R> unwrap,
-    SerializeCallback<R> serialize,
-    DeserializeCallback<R> deserialize, [
-    // ignore: avoid_positional_boolean_parameters
+  Future<void> wrap<T, R>({
+    required Account account,
+    required String cacheKey,
+    required BehaviorSubject<Result<T>> subject,
+    required NextcloudApiCallback<R> callback,
```
Leptopoda, line 139:
```suggestion
    required NextcloudApiCallback<R> requestt,
```

Feel free to also change the typedef name (and docs) if want to keep it aligned.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -249,8 +252,8 @@ class RequestManager {
   /// Otherwise the returned future will be completed with a [TimeoutException].
   /// If the timeout is disabled through [disableTimeout] the future of the
   /// callback is returned immediately.
-  Future<T> timeout<T>(
-    NextcloudApiCallback<T> callback, {
+  Future<T> timeout<T>({
+    required NextcloudApiCallback<T> callback,
```
Leptopoda, line 256: Maybe a positional parameter is still better here.
In most cases we do not need to specify anything more.

Leptopoda (changes requested):
Can  you please also adjust the documentations now that it is possible to explain what parameter does what.

packages/neon_framework/lib/src/utils/request_manager.dart
provokateurin: I just removed the entire typedef as I don't find it useful. It's just a simple mapping to anther type with the same generics and nothing else. Also the typedef wasn't used outside the file, so it didn't serve any purpose.

Leptopoda (approved):

Issue #1472, opened 2024-01-18
If the response is empty, we still want to write into the file.


Leptopoda:
If the response is empty, we still want to write into the file.

_Originally posted by @provokateurin in https://github.com/nextcloud/neon/pull/1448#discussion_r1457023068_

Leptopoda:
I'll also add a unit test for that case :)

PR #1534, opened 2024-01-27
fix(nextcloud): create file even if webdav response is empty

Leptopoda:
fixes: #1472

### DIFF
--- a/packages/nextcloud/lib/src/webdav/client.dart
+++ b/packages/nextcloud/lib/src/webdav/client.dart
@@ -232,6 +232,8 @@ class WebDavClient {
       );
 
   /// Gets the content of the file at [path].
+  ///
+  /// If the response is empty the file will be created with no data.
   Future<void> getFile(
     PathUri path,
     File file, {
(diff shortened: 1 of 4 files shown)


packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -232,29 +232,36 @@ class WebDavClient {
...
-        }
-      });
-      await completer.future;
-      await sink.close();
+
+    if (contentLength == null || contentLength <= 0) {
+      await file.create();
+      onProgress?.call(1);
+      return;
     }
```
provokateurin, line 249: Did you test if just opening the sink, adding no elements and then closing it also creates the file? Then we wouldn't need an special handling

Leptopoda: Yep this is working but we would still need to emit the `1` as the progress again creating a separate condition:
Is this really easier to read?

```
  Future<void> getFile(
    PathUri path,
    File file, {
    void Function(double progress)? onProgress,
  }) async {
    final response = await getStream(path);
    final contentLength = response.contentLength;

    final sink = file.openWrite();

    if (contentLength != null && contentLength > 0) {
      final completer = Completer<void>();
      var downloaded = 0;

      response.stream.listen((chunk) async {
        sink.add(chunk);
        downloaded += chunk.length;
        onProgress?.call(downloaded / contentLength);
        if (downloaded >= contentLength) {
          completer.complete();
        }
      });

      await completer.future;
    } else {
      onProgress?.call(1);
    }

    await sink.close();
  }
```

packages/nextcloud/test/webdav_test.dart
```diff
@@ -653,6 +653,26 @@ void main() {
...
+
+        await client.webdav.putFile(
+          source,
+          source.statSync(),
+          PathUri.parse('empty-file'),
+        );
+        await client.webdav.getFile(
+          PathUri.parse('empty-file'),
+          destination,
+          onProgress: (progress) => expect(progress, equals(1)),
```
Leptopoda, line 670: This test case needs to be changed as it only checks that the progress equals `1` if called and does not check that the progress is called at all.

packages/nextcloud/lib/src/webdav/client.dart
provokateurin: I forgot about that, then this is indeed cleaner.

provokateurin (approved):

Issue #1536, opened 2024-01-27
Persist cookies on disk


provokateurin:
The cookies should have been stored on disk to allow caching the credentials between runs.

provokateurin:
Ah but it is not so easy to separate the cookies between accounts since one might have multiple accounts on the same server.

Leptopoda:
But shouldn't neon still persist them? Just on a per account basis.
Thankfully the cookie_jar package provides nice abstracted interfaces so we can implement it into our existing sqlite or shared_prefs workflow.

provokateurin:
Yes they should be persisted but separate per account with a custom storage.

Issue #1434, opened 2024-01-10
Stop re-exporting packages to avoid breaking changes


provokateurin:
When we re-export another package then the breaking changes of that package also become our breaking changes.

provokateurin:
I wonder if this actually makes sense. Do we consider updating a dependency with a breaking change to be a breaking change for our users as well?
From the user perspective it doesn't matter if we re-export a package and the package has a breaking change or we update a dependency and pull in a breaking change. We could mitigate it by using version ranges, but that is a lot of maintenance and error-prone.
I'm not sure if renovate correctly adds the `!` when the update is a major version bump (but we can probably configure that).
@Leptopoda what do you think?

provokateurin:
I couldn't find an option to configure the `!` in the commit messages with renovate and there was also no relevant issue to the problem.

provokateurin:
Either way we need to mark major dependency updates as breaking changes. Afterwards we can still decided if we want to keep re-exporting the packages. I'd be in favor of that because it makes it easier for consumers.

Leptopoda:
I would not consider both cases the same. 
See this discussion in the dependency_validator issue tracker (https://github.com/Workiva/dependency_validator/issues/6). The person (from the dart-core team) argues that in cases you depend on the interface of a transitive dependency you should just add it to your pubspec to avoid such issues.

A breakdown on all packages (correct me if I missed something):
- nextcloud (export of dynamite_runtime)
	it remains to be seen how often we need to bump the runtime but we should consider it an integral part of the generated package. In that case we should probably stay with the mentality of "breaking dynamite change should result in a breaking NC bump". So we could (any maybe should) keep the export.
	
- dynamite_runtime (export of cookie_jar)
	The cookie jar is only used by the nextcloud_test ([link](https://github.com/nextcloud/neon/blob/3b3ad8926357cfcf0363ce2ad18912bb9c26379e/packages/nextcloud_test/lib/src/test_client.dart#L56-L57)) package and client (should be switched to a direct dependency and import). 
[...]

PR #1537, opened 2024-01-27
refactor(dynamite_runtime): stop exporting cookie_jar

Leptopoda:
…t is persiststent

#1434 for dynamite_runtime
towards:: #1536 

Long term we should use the neon sqflite backend for this (I'll open an issue after this lands).

### DIFF
--- a/packages/neon_framework/lib/src/models/account.dart
+++ b/packages/neon_framework/lib/src/models/account.dart
@@ -1,5 +1,6 @@
 import 'dart:convert';
 
+import 'package:cookie_jar/cookie_jar.dart';
 import 'package:crypto/crypto.dart';
 import 'package:http/http.dart';
 import 'package:json_annotation/json_annotation.dart';
--- a/packages/dynamite/dynamite_runtime/lib/http_client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/http_client.dart
@@ -1,6 +1,5 @@
 /// The dynamite client to handle http connections.
 library http_client;
 
-export 'package:cookie_jar/cookie_jar.dart';
 export 'src/dynamite_client.dart';
 export 'src/http_extensions.dart';
--- a/packages/dynamite/dynamite_runtime/test/client_test.dart
+++ b/packages/dynamite/dynamite_runtime/test/client_test.dart
@@ -1,3 +1,4 @@
+import 'package:cookie_jar/cookie_jar.dart';
 import 'package:dynamite_runtime/http_client.dart';
 import 'package:http/http.dart';
 import 'package:http/testing.dart';
--- a/packages/neon_framework/pubspec.yaml
+++ b/packages/neon_framework/pubspec.yaml
@@ -9,6 +9,7 @@ environment:
 dependencies:
   built_collection: ^5.0.0
   collection: ^1.0.0
+  cookie_jar: ^4.0.0
   crypto: ^3.0.0
   cupertino_icons: ^1.0.0
   dynamic_color: ^1.0.0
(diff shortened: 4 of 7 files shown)


packages/neon_framework/lib/src/models/account.dart
```diff
@@ -32,14 +33,18 @@ class Account implements Credentials, Findable {
...
-        );
+  }) : id = _buildId(username, serverURL) {
+    client = NextcloudClient(
+      serverURL,
+      loginName: username,
+      password: password,
+      appPassword: password,
+      userAgentOverride: userAgent,
+      cookieJar: PersistCookieJar(
+        storage: FileStorage('.cookies/$id'),
```
provokateurin, line 44: Shouldn't we point to the application cache dir? Or is that automatically prepended?

provokateurin (commented):
Formatting is not correct.

packages/neon_framework/lib/src/models/account.dart
Leptopoda: I just used the same approach the `PersistCookieJar` was defaulting to.
If you prefer we can keep the in memory version for now and just move to sqlite directly. I should have the time to do this in the coming days.

provokateurin: Yeah that would be better

provokateurin:
To get this merged already we can pass the memory cookie jar from neon_framework. That wouldn't change the behavior from how it is done right now.

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
```diff
@@ -383,9 +383,10 @@ class DynamiteClient {
     Uri baseURL, {
     this.baseHeaders,
     http.Client? httpClient,
-    this.cookieJar,
+    CookieJar? cookieJar,
     this.authentications,
   })  : httpClient = httpClient ?? http.Client(),
+        cookieJar = cookieJar ?? PersistCookieJar(),
```
provokateurin, line 389: I would prefer using a memory-only cookie jar by default. In the worst case something will not work between runs, but with a persistent cookie jar the worst case would be authenticating with the wrong user and executing (unintentionally) harmful requests.

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/dynamite_client.dart
Leptopoda: Why? We have to consider that this is going to be the default for all dynamite generated libraries.
I don't fully understand the reasoning why an in memory cache is less dangerous than a file backed implementation or how this would affect the general nextcloud package

provokateurin: Storing the cookies on disk makes it possible to read them easily so attackers could do that, but I was more thinking about leaking cookies where it shouldn't happen, like in our case with different accounts on the same server.

Leptopoda: Should we default to the `WebCookieStore` which does not store any cookies at all.

provokateurin: Instead of that I would make the cookie jar nullable like it was and provide no implementation.
This would be my preferred solution as we can't anything much wrong by default.

provokateurin: So just what we have on main minus re-exporting the package.

provokateurin (approved):
PR title is wrong.

PR #1540, opened 2024-01-27
refactor(neon_framework): Migrate image caching to RequestManager

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1125
I had to basically rewrite all the Neon*Image classes, but I think this made everything much cleaner and streamlined. Regardless of where the image is coming from, it will be cached correctly.


Leptopoda (changes requested):
fc5d7b185bea295967f980423880853920cb4818 (removing the slow cache test) is a big no go.
We had a real bug caused by this (the reason why I started the unit tests in the first place) and while adding these tests I also discovered two other inconsistencies.

Tests are there for a reason and unless functionality covered by a test is dropped we should never drop tests.

EDIT: after more time I see that you removed the potential race condition entirely. I'm not sure yet if we should do that because I have seen a few devices where a 300Km api request was faster than the cache. Yes some low end devices have questionable storage inside.

Leptopoda:
Full review will take some more time. Worst case delayed until mid next week.

provokateurin:
I removed the slow cache tests because we always wait for the cache now in order to know when we definitely don't need to execute the network request. I'm not sure how you would abort it async, but I also think it defeats the purpose since the network request will already be started.

provokateurin:
(just a rebase)

packages/neon_framework/lib/src/widgets/image.dart
```diff
@@ -336,70 +308,90 @@ class NeonUrlImage extends StatelessWidget {
...
-        }
+    super.dispose();
+  }
 
-        final completedUri = account.completeUri(uri);
+  Future<void> load() async {
+    if (widget.uri.data != null) {
+      var data = widget.uri.data!.contentAsBytes();
+      try {
+        data = utf8.encode(ImageUtils.rewriteSvgDimensions(utf8.decode(data)));
```
Leptopoda, line 354: We must cut down the parsing that is going on.
We retrieve the data from cache as base64, convert it to a String (utf8) convert it back to binary (utf8) just so the SvgAssetLoader can convert it back into a String by decoding it (utf8). When rendering we also utf8 decode it to check whether it even is a Svg or a PNG/JPEG

Maybe the svg dimension rewrite can happen in the asset loader? Or maybe we should use a String eliminating the utf8 decoding for checking whether it is a PNG.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -134,6 +135,87 @@ class RequestManager {
...
+        },
+        unwrap: (rawResponse) {
+          var data = rawResponse.response.body;
+          if (unwrap != null) {
+            data = unwrap(data);
+          }
+
+          return data;
+        },
+        serialize: (rawResponse) => base64.encode(rawResponse.response.body),
```
Leptopoda, line 164: As already mentioned there is way to much parsing going on.
Let's just save the binary data directly. IIRC the SQLite data type would be `BLOB`.

We would loose some performance as we would now need to utf8 en/decode normal api/webdav requests but they usually have a smaller payload anyways thus not being that bad.

Leptopoda (changes requested):
Just an initial feedback so you can get the main blocker.
Let's wait until the cache rework has landed so we can do this in one go.

provokateurin:
I agree with you, although I think that those problems should be solved separately. They already exist on main and are not really much worse than it is already.

Leptopoda:
I already have a local branch.

provokateurin:
Ok, if you create a follow up to this PR, what about this one?

Leptopoda (approved):

provokateurin:
I didn't expect this to go so clean :D

PR #1541, opened 2024-01-27
reuse webdav client for dav

vauvenal5:
So basically this is my idea how to reuse the webdav client for dav. Not sure if I am missing something or not. Let me know if you had something different in mind.


packages/nextcloud/lib/src/webdav/path_uri.dart
```diff
@@ -12,6 +12,12 @@ class PathUri {
     required this.pathSegments,
   });
 
+  /// Base path used for dav on the server
+  factory PathUri.dav() => PathUri.parse('/remote.php/dav');
+
+  /// Base path used for webdav on the server
+  factory PathUri.webdav() => PathUri.parse('/remote.php/webdav');
+
```
Leptopoda, line 20: The `PathUri` should be a `Uri` like class and not provide any paths itself.

Maybe create an enum with a getter that returns the uri.

packages/nextcloud/lib/webdav.dart
Leptopoda, line 1: The extension from `packages/nextcloud/lib/dav.dart` should also be moved into here.
The other one is not going to be removed automatically if the spec isn't there anymore.

Leptopoda (commented):
Some minor code style feedback for now.
tests are obviously still needed and I don't pretend to know the difference between the two dav endpoints. I also do not understand why we should remove the specs for the one.
@provokateurin can propably review this side better.

packages/nextcloud/lib/src/webdav/file.dart
```diff
@@ -28,7 +27,7 @@ class WebDavFile {
     return PathUri(
       isAbsolute: false,
       isDirectory: href.isDirectory,
-      pathSegments: href.pathSegments.sublist(webdavBase.pathSegments.length),
+      pathSegments: href.pathSegments.sublist(PathUri.webdav().pathSegments.length),
```
provokateurin, line 30: Shouldn't this still use the base?

packages/nextcloud/lib/src/api/dav.openapi.json
provokateurin, line 1: These files should not be removed.

packages/nextcloud/lib/dav.dart
provokateurin, line 1: This file should also not be changed.

packages/nextcloud/lib/src/webdav/path_uri.dart
provokateurin: IMO this can just be hardcoded in the constructors of the WebDavClient.

provokateurin (changes requested):
I'm not sure if this approach is the best.
The naming conflicts are definitely annoying here :/

packages/nextcloud/lib/dav.dart
vauvenal5: I believe I am misunderstanding something. What is the difference between the dav client and the webdav client? To my current understanding the generated clients are basically for OpenAPI endpoints. So why is for dav a client generated but not for webdav?

provokateurin: The generated client is from OpenAPI, but it is only for some REST endpoints of the [`dav`](https://github.com/nextcloud/server/tree/master/apps/dav) app.

packages/nextcloud/lib/src/webdav/file.dart
vauvenal5: I was looking into that too, it is not ideal, however getting the actual base path there is not trivial due to the XML handling. I will be thinking about this one more time but wanted general feedback first.

provokateurin:
@vauvenal5 please make sure that everything is covered with tests, otherwise we will have a hard time verifying that it works as intended.

vauvenal5:
I will close this one and re-open with the new changes latter on.

PR #1551, opened 2024-01-29
refactor(neon_framework): Align mock class names

provokateurin:
Follow up to https://github.com/nextcloud/neon/pull/1547


Leptopoda (approved):
I like the `Mock` being suffixed but I can see how others prefer it as a prefix

provokateurin:
I'm fine with either option, just want the all the names to follow the same pattern.

PR #1552, opened 2024-01-29
 feat(neon_framework): Support caching with ETag and Expires in RequestManager

provokateurin:
Split off from https://github.com/nextcloud/neon/pull/1540


packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -137,24 +146,44 @@ class RequestManager {
...
+              key,
+              cacheParameters.etag,
+              cacheParameters.expires != null ? _parseExpires(cacheParameters.expires!) : null,
+            ),
+          );
+          subject.add(subject.value.copyWith(isLoading: false));
+          return;
+        }
+      }
     }
```
Leptopoda, line 187: There is a lot of cache handling going on here.
Now that we only ever call `emitCache` once we should think about merging it into the main wrap method.

This would probably allow us easier logic cleanups and more readability.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -137,24 +146,44 @@ class RequestManager {
...
     final key = '${account.id}-$cacheKey';
 
-    Future<void>? cacheFuture;
     if (retries == 0) {
       // emit loading state with the current value if present
       final value = subject.valueOrNull?.asLoading() ?? Result.loading();
       subject.add(value);
 
-      cacheFuture = _emitCached(
+      final cachedValue = await _emitCached(
```
Leptopoda, line 161: Retrieving from cache can be expensive. Especially when it is large binary data (retrieving from sqlite and passing into the dart code via ffi).

I suggest a separate method to retrieve the cache parameters and only query the data if still valid.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -26,7 +28,10 @@ typedef SerializeCallback<T> = String Function(T);
 
 /// A callback to revive cached values [String] into their original type [T].
 @internal
-typedef DeserializeCallback<T> = T Function(String);
+typedef DeserializeCallback<T> = FutureOr<T> Function(String);
+
+/// A callback to get the ETag and Expires of the resource.
+typedef GetCacheParameters = FutureOr<({String? etag, String? expires})?> Function();
```
Leptopoda, line 34: Can we create a separate class with these two parameters?
The class should also have a getter `isExpired` wisch automatically checks the parameter against the current DateTime.

The typedef could be removed in favor of an `AsyncValueGetter<CacheParameters>`

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -137,24 +146,44 @@ class RequestManager {
...
         deserialize,
       );
+
+      if (cachedValue != null && cachedValue.value != null) {
+        if (cachedValue.expires?.isAfter(DateTime.now()) ?? false) {
+          subject.add(subject.value.copyWith(isLoading: false));
+          return;
+        }
+
+        final cacheParameters = await getCacheParameters?.call();
```
Leptopoda, line 174: needs to be timed out

packages/neon_framework/lib/src/utils/request_manager.dart
Leptopoda: It should also handle parsing the headers

Leptopoda (changes requested):
initial first feedback.

packages/neon_framework/lib/src/utils/request_manager.dart
provokateurin: Good point, I didn't think about that!
I guess then the problem with the slow cache is no more?

Leptopoda: I have not done a performance analysis but I think that we where really favoring the network request.
But even if the entire time will be a bit slower on low end device we must also consider the saved request on the server.

Let's at least try this and if needed I can do a performance check later on

packages/neon_framework/lib/src/utils/request_manager.dart
provokateurin: I agree, but I think we only called the method once before this already :thinking:

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -26,7 +28,7 @@ typedef SerializeCallback<T> = String Function(T);
 
 /// A callback to revive cached values [String] into their original type [T].
 @internal
-typedef DeserializeCallback<T> = T Function(String);
+typedef DeserializeCallback<T> = FutureOr<T> Function(String);
```
Leptopoda, line 31: Let's keep this synchronous
```suggestion
typedef DeserializeCallback<T> = T Function(String);
```

Leptopoda (approved):

PR #1560, opened 2024-01-31
fix(nextcloud): Stop exporting the crypton package

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1559


Leptopoda (approved):
thanks. Can you also double check that there is no other such case?
But having another person check is always good.

PR #1572, opened 2024-02-02
fix(tool): Add Husky support for Windows

0niel:
Hi guys! When using Husky on Windows, it runs through git bash and does not recognize the `fvm` command.

I tried to fix the problem with this commit


provokateurin:
You need to run the ./tool/setup.sh script which will install the right fvm version. You also need to make sure that the pub cache is in your path.

provokateurin:
@0niel can we close this? This is not a real fix and happened because your environment wasn't setup correctly.

PR #1573, opened 2024-02-02
fix(neon_framework): Add null check for location in `WeatherStatusBloc`

### LINKED ISSUE #1555: Do not try to load forecast without a location being set
Currently we request the forecast regardless of the location. If no location is currently set then we also don't need to try loading the forecast which will fail anyway.

0niel:
Possibly fixed #1555

### DIFF
--- a/packages/neon_framework/lib/src/blocs/weather_status.dart
+++ b/packages/neon_framework/lib/src/blocs/weather_status.dart
@@ -82,6 +82,10 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
       return;
     }
 
+    if (location.valueOrNull == null) {
+      return;
+    }
+
     await Future.wait([
       RequestManager.instance.wrapNextcloud(
         account: account,
--- a/packages/neon_framework/lib/src/blocs/weather_status.dart
+++ b/packages/neon_framework/lib/src/blocs/weather_status.dart
@@ -52,8 +52,12 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
     timer = TimerBloc().registerTimer(const Duration(minutes: 5), refresh);
 
     location.listen((locationResult) async {
-      if (!locationResult.isLoading && !locationResult.hasError) {
-        await refreshForecast();
+      if (!locationResult.isLoading && locationResult.hasData) {
+        if (locationResult.data?.address != null) {
+          await refreshForecast();
+        } else {
+          location.add(Result.success(null));
+        }
       }
     });
   }
--- a/packages/neon_framework/lib/src/blocs/weather_status.dart
+++ b/packages/neon_framework/lib/src/blocs/weather_status.dart
@@ -50,6 +50,12 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
     });
 
     timer = TimerBloc().registerTimer(const Duration(minutes: 5), refresh);
+
+    location.listen((locationResult) async {
+      if (!locationResult.isLoading && !locationResult.hasError) {
+        await refreshForecast();
+      }
+    });
   }
 
   final Account account;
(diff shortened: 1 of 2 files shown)


provokateurin:
This won't work as it also prevents loading the location in the first place.
This if check needs to move into the Future.wait to only load the forecast conditionally, but we also need to trigger the loading of the forecast when the location changes from null to something else or back to null.

provokateurin:
To be a bit clearer (I didn't think this through myself): The `refresh` method should only reload the location. Then you need to listen to the location subject and every time a `Result` is emitted that has `!isLoading && !hasError` you need to reload the forecast. This is to account for cached as well as for live data.

0niel:
@provokateurin thank you for your feedback! Based on your comments, I have tried to make the necessary corrections to solve the problem.

packages/neon_framework/lib/src/blocs/weather_status.dart
```diff
@@ -50,6 +50,12 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
     });
 
     timer = TimerBloc().registerTimer(const Duration(minutes: 5), refresh);
+
+    location.listen((locationResult) async {
+      if (!locationResult.isLoading && !locationResult.hasError) {
```
provokateurin, line 55:
```suggestion
      if (!locationResult.isLoading && locationResult.hasData) {
```
The error might be because the user is offline, but cached data might be available.

packages/neon_framework/lib/src/blocs/weather_status.dart
```diff
@@ -50,6 +50,12 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
...
+        await refreshForecast();
```
provokateurin, line 56: This is missing the condition to check if the location is actually set.
If it is set then the method needs to be executed, otherwise a `Result.success(null)` needs to be added to the subject to clear the old data (the type needs be nullable).

provokateurin (changes requested):

provokateurin (changes requested):
Please also add tests to ensure no regression will happen

packages/neon_framework/test/weather_status_bloc_test.dart
```diff
@@ -211,4 +211,11 @@ void main() {
       ..setLocation('Hamburg')
       ..setLocation('Berlin');
   });
+
+  test('Do not refresh if location has no data', () async {
+    bloc.location.add(Result.loading());
+    bloc.location.add(Result.success(null));
```
provokateurin, line 217: Do not modify the subject directly, you need to provide some mock responses that trigger this case.

packages/neon_framework/lib/src/blocs/weather_status.dart
```diff
@@ -50,6 +50,16 @@ class _WeatherStatusBloc extends InteractiveBloc implements WeatherStatusBloc {
     });
 
     timer = TimerBloc().registerTimer(const Duration(minutes: 5), refresh);
+
+    location.listen((locationResult) async {
+      if (!locationResult.isLoading && locationResult.hasData) {
+        if (locationResult.data?.address != null) {
+          await refreshForecast();
+        } else {
+          location.add(Result.success(null));
```
provokateurin, line 59: The forecast needs to be reset and not the location.

provokateurin (changes requested):
The test should be like this:
1. Set the location to something valid
2. Set the location to something invalid
3. Set the location to something valid

During those changes the forecast stream should emit in order:
1. A valid forecast
2. Null
3. A valid forecast

provokateurin:
Hi @0niel are you able to make the requested changes? Do you need some help or pointers what to do?

provokateurin:
I implemented this in https://github.com/nextcloud/neon/pull/1868 and did it in a better way than what I suggested here because I now realized how to properly implement this.

PR #1580, opened 2024-02-06
feat(dynamite_runtime): make DynamiteClient extend http.Client

Leptopoda:
My plan is to adapt the request manager to work with a generic `http.Client` and a `Request`.
From there the request manager can use streams, converters and whatever it needs.
This would unify/deduplicate the `RequestManager.wrapUri`, `RequestManager.wrapBinary` and  `RequestManager.wrapNextcloud` (leaving only the webdav methods separate for now).

By creating a custom request we are able to easily change the http method to a `HEAD` if we decide to use it for caching later on.

Please provide feedback on the dynamite_runtime changes (should we really have that many converters?
It also contains other corrections like respecting the `content-type` header that I can land in separate PRs if desired.


packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
provokateurin, line 1: Why not get rid of the deprecated things right away?

packages/dynamite/dynamite_runtime/lib/src/utils/byte_stream_extension.dart
```diff
@@ -0,0 +1,83 @@
...
+    return buffer.toBytes();
+  }
+
+  /// Collect the data of this stream in a json object, decoded according to
+  /// [encoding], which defaults to `UTF8`.
+  Future<Object?> bytesToJson([Encoding encoding = utf8]) =>
+      encoding.decoder.fuse(const JsonDecoder()).bind(this).first;
+
+  /// Collect the data of this stream in [XmlElement], decoded according to
+  /// [encoding], which defaults to `UTF8`.
```
provokateurin, line 37:
```suggestion
  /// Collect the data of this stream in a json object, decoded according to
  /// [encoding], which defaults to [utf8].
  Future<Object?> bytesToJson([Encoding encoding = utf8]) =>
      encoding.decoder.fuse(const JsonDecoder()).bind(this).first;

  /// Collect the data of this stream in [XmlElement], decoded according to
  /// [encoding], which defaults to [utf8].
```

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
```diff
@@ -130,6 +122,17 @@ class DynamiteClient with http.BaseClient {
       await cookieJar!.saveFromResponse(request.url, cookies);
     }
 
+    if (request case DynamiteRequest(:final validStatuses)
+        when validStatuses != null && !validStatuses.contains(response.statusCode)) {
```
provokateurin, line 126: First I've never seen this syntax :sweat_smile: and it also is not very readable to have it all in the if statement

provokateurin (commented):
The code looks good, but the tests are failing and I'm really seeing the direction this is going. Can you roughly explain why the changes are needed or how this will be used later on?

packages/dynamite/dynamite_runtime/lib/src/http_extensions.dart
Leptopoda: It's a breaking change in dynamite_runtime and thus also in the nextcloud package.

There is no maintenance bourdon in keeping it. We can remove it once other breaking changes have accumulated.

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
Leptopoda: Dart3 pattern matching :P

Do you really think the usual if is easier to read?
```dart
    if (request is DynamiteRequest) {
      final validStatuses = request.validStatuses;
      if (validStatuses != null && !validStatuses.contains(response.statusCode)) {
        
      }
    }
```
or as we do not need the validStatuses itself
```dart
    if (request is DynamiteRequest && request.validStatuses != null 
      && !request.validStatuses!.contains(response.statusCode)) {
```

`validStatuses` is a getter so we do not get null check elevation.
I agree that it might not be strictly necessary here but I would also not be against it in general. It can help reduce a lot of boilerplate code.

Leptopoda:
> but the tests are failing

I did not add the dynamite (generator) changes yet. Neither did I add the re generated nextcloud and petstore code.
I found it easier to understand (the dynamite changes themselves might need multiple PRs as there where bigger restructures involved).

>  Can you roughly explain why the changes are needed or how this will be used later on?

I think the first PR post sums it up nice.
The idea came when I tried to change the request handling and decoding of images. We currently require all of them to be `DynamiteRawResponse`s which creates some overhead. My broader goal was to make the request manager work with any type of request and response unifying the special handling we currently have for wrapUrl and wrapBinary.
I attached a patch where I realised that the current way is too complex.

[request_manager.patch.txt](https://github.com/nextcloud/neon/files/14220540/request_manager.patch.txt)

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
provokateurin: Either of those options are better than the current one imo.

Leptopoda:
Should be pretty ready for the next round of feedback.
@provokateurin I'll need your help with the regex magic

I didn't migrate all request manager methods yet as the PR is already enormous.
Should I split it apart?

provokateurin:
Please split everything you can, as long as it makes sense :sweat_smile: Then it can be reviewed and merged much faster.

Leptopoda:
Closing in favor of the separate PRs

Issue #1538, opened 2024-01-27
Refactor neon storages 


Leptopoda:
replaces: #1280, #793, #1536

Neon currently uses:
- memory based cookie persistence
- shared_preferences based options persistence
- sqlite based request caching

We should:
- [x] migrate all into sqlite (probably separate databases to keep the caches separate from actual data)
- [x] introduce separate interfaces to make switching in the future easier and keep the code separate (our options already use `SettingsStorage` and cookie_jar provides `Storage` so only needed for the cache)
<s> - [ ] use encryption on all storages </s>
<s>- [ ] add a storage version to the shared preference persistence for potential migrations.</s>

This probably requires major restructures and cleanups in the framework as we also expose `SingleValueStorage` a generic key value like storage. We should maybe rethink our key handling as we currently use enums to avoid duplication but we are also nesting this which doesn't scale well.

provokateurin:
We can store the cookies in the same database as the cache as they are only temporary information and can be easily recreated.

Leptopoda:
We should disable both the cookie and request cache for web as this should be handled by the browser.

Issue #1280, opened 2023-12-13
Encrypt the user credentials when stored


Leptopoda:
related to: #793

If we decide to stay with shared_prefs we should consider using sncrypted_shared_prefs (or encrypt them ourselves)

provokateurin:
All shared_preference usage should be encrypted as wight store other sensitive things as well. Won't really impact performance as we use enencrypted sqlite for caching.

Leptopoda:
I looked into this yesterday and it looks like most ecrypted shared prefs implementations are not well maintained. We might need to write a custom wrapper for this.

Leptopoda:
Closing as we decided that the added encryption is not worth the bundle size increase.
Adding it in is trivial so open an issue if you need this functionality.

PR #1585, opened 2024-02-09
feat(neon_framework)!: encrypt storages as sqlcypher databases

Leptopoda:
fixes: #1538, #1280

While sqlcipher supports migrating unencrypted databases I did not consider it in the scope of this PR.
We would need to create a ne db file and then migrate ...
Please make sure to remove the old cache as the DB might otherwise throw.
I also suggest cleaning your build cache with something like `melos exec flutter clean` as otherwise cmake might fail.

 I've tested it on both linux and android but only verified on linux that the db is actually encrypted.

### DIFF
--- a/.cspell/misc.txt
+++ b/.cspell/misc.txt
@@ -9,6 +9,7 @@ persistences
 playstore
 postmarket
 provokateurin
+sqlcipher
 subroutes
 uncategorized
 upsert
--- a/packages/app/linux/CMakeLists.txt
+++ b/packages/app/linux/CMakeLists.txt
@@ -2,6 +2,10 @@
 cmake_minimum_required(VERSION 3.10)
 project(runner LANGUAGES CXX)
 
+# SQLCipher do not bundle openssl
+# https://github.com/simolus3/sqlite3.dart/issues/186
+set(OPENSSL_USE_STATIC_LIBS OFF)
+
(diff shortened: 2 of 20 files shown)


Leptopoda:
Making a draft so I don't  forget to add unit/integration tests.

provokateurin (commented):
I didn't have a complete look at everything, but it looks pretty good to me (will do a proper review once the other PR is merged).

Will this affect the build size of the app?
Is there any way the request cache and sql persistence could share more logic? currently they are quite similar which is not great.

Leptopoda:
> Will this affect the build size of the app?

probably; tests coming
I'll also verify encryption is working on android as you have suggested.

> Is there any way the request cache and sql persistence could share more logic?

I don't think there is anything they can really share. It's only the init method that is similar while every other is implementation specific. I might reconsider moving the database creation outside of the init method so other persistences can use the same db (thinking of the cookies which should be stored with the cache).

Other than that I don't think we can do anything here.

Leptopoda:
Ok I was able to confirm that the DB is also encrypted on android.
The apk size increase is arround 20MB for the universal build (78.9MB => 100MB). I don't think there is anything we can do about this.

Leptopoda:
Closing as we decided that the added encryption is not worth the bundle size increase.
Adding it in is trivial so open an issue if you need this functionality.

PR #1598, opened 2024-02-13
refactor(neon_framework): Make AppImplementations a BuiltSet

provokateurin:
We only accept a Set to begin with, so we can make it a Set everywhere.


Leptopoda:
What about BuiltSet?

provokateurin:
I only wanted to align the types, but I guess it makes sense to make it immutable :+1:

Leptopoda (approved):

PR #1599, opened 2024-02-13
refactor: Remove unnecessary generic type arguments

provokateurin:
Much cleaner code by omitting all the unnecessary generic type arguments.
In the Blocs I found a bunch of places where the subjects were not final, I fixed that too (I could extract it if really necessary).
I used the following regex to find all the potential places: ` [A-Z]+[A-Za-z]*<.*>(\(|\.)` (notice the leading space; this also finds a bunch of false-positives and in a few places the generics are still needed).

### DIFF
--- a/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
+++ b/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
@@ -47,7 +47,7 @@ class _DashboardBloc extends InteractiveBloc implements DashboardBloc {
         }
       }
 
-      await Future.wait<void>([
+      await Future.wait([
         if (v1WidgetIDs.isNotEmpty)
           RequestManager.instance.wrapNextcloud(
             account: account,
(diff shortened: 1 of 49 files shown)


packages/neon_framework/lib/neon.dart
```diff
@@ -68,15 +68,15 @@ Future<void> runNeon({
...
+        NeonProvider.value(value: globalOptions),
+        NeonProvider.value(value: accountsBloc),
+        NeonProvider.value(value: firstLaunchBloc),
+        NeonProvider.value(value: nextPushBloc),
+        Provider(
           create: (_) => appImplementations,
           dispose: (_, appImplementations) => appImplementations.disposeAll(),
         ),
-        Provider<PackageInfo>.value(value: packageInfo),
+        Provider.value(value: packageInfo),
```
Leptopoda, line 79: Please leave the generics here.
Changing the provided value could otherwise have unwanted side effects down the road.

provokateurin: I also added them back to the SelectOption as I feel like one could easily make a mistake there.

packages/neon_framework/lib/neon.dart
```diff
@@ -77,7 +77,7 @@ Future<void> runNeon({
...
```
Leptopoda, line 80: Please add back the generic.

Leptopoda (approved):
Other than this LGTM

PR #1602, opened 2024-02-14
refactor(neon_framework): use a separate persistence layer backing th…

Leptopoda:
…e storages

The general idea is to decouple the storage layer, the app is talking to, from the layer actually persisting the data.
A `Persistence` is just a basic CRUD like (maybe asynchronous) interface for the storage layers to use. Other implementations like the `CachedPersistence` cache the data in memory (HashMap) so reads can be synchronous while writes are asynchronous in the background.

In turn we can build a complete `SharedPreferences` implementation on to of the cached persistence. Currently the `SharedPreferencesPersistence` is just a `CachedPeristence` warper around the shared preferences (with some key handling magic to allow the current data to seamlessly be used without any migration).

I understand that the current default implementation of our storages (`SettingsStore` and `SingleValueStore`) are already wrappers themselves but with the up cumming migration to a SQLite persistence we will no longer have the double wrappers. This may seem like a lot of unnecessary boilerplate interfaces but as shown in #1585 it will allow us to make changes with minimal effort.

We can later consider changing some of these patterns but it was the easiest way to tackle it in a reviewable small patch-set.

### DIFF
--- a/packages/neon_framework/lib/src/storage/persistence.dart
+++ b/packages/neon_framework/lib/src/storage/persistence.dart
@@ -1,0 +1,67 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart' show protected;
+
+/// A persistent key value storage.
+abstract interface class Persistence<T extends Object> {
+  /// Whether a value exists at the given [key].
+  FutureOr<bool> containsKey(String key);
+
+  /// Clears all values from persistent storage.
+  FutureOr<bool> clear();
+
+  /// Removes an entry from persistent storage.
+  FutureOr<bool> remove(String key);
+
+  /// Saves a [value] to persistent storage.
+  FutureOr<bool> setValue(String key, T value);
+
+  /// Fetches the value persisted at the given [key] from the persistent
+  /// storage.
+  FutureOr<T?> getValue(String key);
+}
+
+/// A key value persistence that caches read values to be accessed
+/// synchronously.
+///
+/// Mutating values is asynchronous.
+abstract class CachedPersistence<T extends Object> implements Persistence<T> {
+  /// Fetches the latest values from the host platform.
+  ///
+  /// Use this method to observe modifications that were made in the background
+  /// like another isolate or native code while the app is already running.
+  Future<void> reload();
+
+  /// The cache that holds all values.
+  ///
+  /// It is instantiated to the current state of the backing database and then
+  /// kept in sync via setter methods in this class.
+  ///
+  /// It is NOT guaranteed that this cache and the backing database will remain
+  /// in sync since the setter method might fail for any reason.
+  @protected
+  final Map<String, T> cache = {};
+
+  @override
+  T? getValue(String key) => cache[key];
+
+  /// Saves a [value] to the cached storage.
+  ///
+  /// Use this method to cache type conversions of the value that do not change
+  /// the meaning of the actual value like a `BuiltList` to `List` conversion.
+  /// Changes will not be persisted to the backing storage and will be lost
+  /// when the app is restarted.
+  void setCache(String key, T value) => cache[key] = value;
+
+  @override
+  bool containsKey(String key) => cache.containsKey(key);
+
+  @override
+  Future<bool> clear();
+
+  @override
+  Future<bool> remove(String key);
+
+  @override
+  Future<bool> setValue(String key, T value);
+}
--- a/packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
+++ b/packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
@@ -1,0 +1,133 @@
+// ignore_for_file: cascade_invocations
+
+import 'package:meta/meta.dart';
+import 'package:neon_framework/src/storage/persistence.dart';
+import 'package:shared_preferences_platform_interface/shared_preferences_platform_interface.dart';
+import 'package:shared_preferences_platform_interface/types.dart';
+
+/// The version of the [SharedPreferencesPersistence].
+///
+/// Needed to make potential migrations in the future.
+@visibleForTesting
+const int kSharedPreferenceVersion = 1;
+
+/// The default prefix used by `SharedPreferences`.
+///
+/// Used for legacy reasons to allow seamless upgrades.
+// TODO: replace with our packageID
+const String _defaultPrefix = 'flutter.';
+
+/// An SharedPreferences backed cached persistence for preferences.
+///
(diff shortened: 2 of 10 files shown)


packages/neon_framework/lib/src/storage/persistence.dart
```diff
@@ -0,0 +1,67 @@
+import 'dart:async';
+
+import 'package:meta/meta.dart' show protected;
+
+/// A persistent key value storage.
+abstract interface class Persistence<T extends Object> {
+  /// Whether a value exists at the given [key].
+  FutureOr<bool> containsKey(String key);
+
+  /// Completes with true once all values have been cleared.
```
provokateurin, line 10:
```suggestion
  /// Clears all values.
```

packages/neon_framework/lib/src/storage/persistence.dart
```diff
@@ -0,0 +1,67 @@
...
+  FutureOr<bool> clear();
+
+  /// Removes an entry from persistent storage.
+  FutureOr<bool> remove(String key);
+
+  /// Saves a [value] to persistent storage.
+  FutureOr<bool> setValue(String key, T value);
+
+  /// Fetches a the value persisted at the given [key] from the persistent
```
provokateurin, line 19:
```suggestion
  /// Fetches a the value persisted with the given [key] from the persistent
```

packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
```diff
@@ -0,0 +1,110 @@
...
+    final keys = _globalCache.keys;
+    for (final prefixedKey in keys) {
+      if (prefixedKey.startsWith(prefix)) {
+        final key = prefixedKey.replaceFirst(prefix, '');
+        cache[key] = _globalCache[prefixedKey]!;
+      }
+    }
+
+    _globalCache[prefix] = cache;
+    return cache;
```
provokateurin, line 50: Can you explain why this is needed?

This could end up returning values from storages with a prefix, that clashes with another prefix (e.g. `abc` would also return the values of `abc-def`). Is this intended?

packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  static final Map<String, Object> _globalCache = {};
+
+  /// Initializes all persistences by setting up the backing SharedPreferences
+  /// storage and priming the global cache.
+  ///
+  /// This must be called and completed before any sqlite persistence.
```
provokateurin, line 58: Why is sqlite mentioned here?

packages/neon_framework/lib/src/storage/storage_manager.dart
```diff
@@ -68,17 +65,32 @@ class NeonStorage {
...
+  /// with the [SingleValueStore].
   SettingsStore settingsStore(StorageKeys groupKey, [String? suffix]) {
     _assertInitialized();
 
-    return DefaultSettingsStore(_sharedPreferences, groupKey, suffix);
+    var key = groupKey.value;
+
+    if (key.isEmpty) {
+      throw ArgumentError('The group key must not be empty to avoid conflicts with the `SingleValueStore`.');
+    }
```
provokateurin, line 78: Use assert

provokateurin (changes requested):

packages/neon_framework/lib/src/storage/persistence.dart
Leptopoda: This does not explain what the returned value does.

packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
Leptopoda: I initially wrote the persistence with https://github.com/nextcloud/neon/pull/1585 in mind and just later backported it to the current shared_prefs. 
So basically copy pasta.

packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
Leptopoda: I will take a closer look but that's just a limitation of our current shared_prefs implementation.
We can not distinguish between both cases. It doesn't really matter for our current keys but I agree that it is not a good design).
If we only use this persistence for this PR (as a migration) I'd like to just move on with it as it will be deleted in the next one anyways (we can squash them if needed).

Sadly as we already discussed on matrix we might end up with using it on web platforms to avoid longer load times and worse performance for effectively no benefit. 
I'll re think this in the coming days

packages/neon_framework/lib/src/storage/persistence.dart
Leptopoda: I think the at is better because a value is stored at a key (location) and not with a key but I don't really care.

Reading this doc comment I find the start more interesting ` Fetches a the value` :upside_down_face:

packages/neon_framework/lib/src/storage/persistence.dart
provokateurin: But this is also not explained for all the other methods.

packages/neon_framework/lib/src/storage/shared_preferences_persistence.dart
Leptopoda: I think that splitting the values into separate maps does not make any sense if we plan to continue using it on web.
I'll use the old approach of concatenating the two keys and storing it in the single  (global) cache. SharedPrefs just isn't made to handle this kind of data separation.

packages/neon_framework/lib/src/storage/persistence.dart
```diff
@@ -0,0 +1,67 @@
...
+  /// Fetches the latest values from the host platform.
+  ///
+  /// Use this method to observe modifications that were made in the background
+  /// like another isolate or native code while the app is already running.
+  Future<void> reload();
+
+  /// The cache that holds all values.
+  ///
+  /// It is instantiated to the current state of the backing SQLite database
+  /// and then kept in sync via setter methods in this class.
```
provokateurin, line 38: Should this really mention SQLite? It's generic enough that at some point a different database could back it.

provokateurin (approved):

packages/neon_framework/lib/src/storage/persistence.dart
Leptopoda: You're right. I thought I had already fixed this after the first review.

PR #1608, opened 2024-02-17
refactor(neon_framework): Make platform code a plugin

provokateurin:
Will be used with https://github.com/nextcloud/neon/pull/372 where we can then directly rely on package:web instead of going through package:universal_html to be compilable.


Leptopoda (approved):
Did not test but LGTM

PR #1612, opened 2024-02-18
refactor(neon_framework): Use loop for retries in RequestManager

provokateurin:
Easier to read and understand and also harder to trigger an infinite loop.


Leptopoda:
I guess this is mainly to save the roundtrips to the cache?

provokateurin:
No that doesn't change as before we only loaded and emitted the cache value on the first try, so subsequent tries didn't cause any additional load.

Leptopoda (approved):

PR #1613, opened 2024-02-18
refactor(neon_framework): Only emit loading event in RequestManager if no cached data is available

provokateurin:


Leptopoda:
What exactly is the motivation behind this?
I fear that this degrades the ux on slower/older devices.

provokateurin:
Avoid rerendering in the UI even if nothing has changed

provokateurin:
>  I fear that this degrades the ux on slower/older devices.

We could still emit the loading event in case the subject has no value yet. Then it will still show on the initial loading. Once we have any data to show it should be ok to not emit the loading event.

Leptopoda:
Sounds like a good idea.
Let's try it.

provokateurin:
@Leptopoda please review again. I also removed the _emitError method as we can be sure the subject always has a value so we don't need to use valueOrNull (this was already the case before this PR too).

Leptopoda:
So far no issues on any slow real world devices.
Artificially making the cache slower (inserting a `Future.wait` into each RequestCache oeration) is a bit problematic.
The UI is "hanging" and not telling the user to just wait.

While a 1sec wait for every operation is hepefully not going to happen in real world but I'll do another round of testing with on top of the sqlcipher branch (it might legitimately add overhead).

provokateurin:
I'm sorry, but I don't see anywhere I added another await? I only changed how events are emitted and nothing about reading the cache has changed.

Leptopoda:
Sorry for the misunderstanding.

I meant I did this for testing and it turned out to be an issue. That's why I'll test it later with sqlcipher to see ho realistic my test even is.

Leptopoda (approved):

PR #1616, opened 2024-02-19
refactor(neon_dashboard): Stop using collection

provokateurin:
No need to pull in a dependency for such a simple thing.

### DIFF
--- a/packages/neon/neon_dashboard/pubspec.yaml
+++ b/packages/neon/neon_dashboard/pubspec.yaml
@@ -8,7 +8,6 @@ environment:
 
 dependencies:
   built_collection: ^5.0.0
-  collection: ^1.0.0
   flutter:
     sdk: flutter
   flutter_localizations:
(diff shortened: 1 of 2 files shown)


Leptopoda (approved):
What is the benefit of this cahange?
The collection package is used throughout the framework so it's going to be pulled in anyways.
It also brings further helper extensions that might come in handy in the future.
It's not like we are on js where every dependency has a huge impact on the project.

I agree that making this multi line would help in readability.

Leptopoda (changes requested):
Wrong button. I wanted to leave a comment so we can discuss this. Sorry.

provokateurin:
Yes it is used already so it will be included anyway, but in this package it should be only a dev dependency and for this single usage it feels a bit too much to have an additional dependency. In the end it doesn't really change anything, I just noticed this and found it cleaner this way.

Leptopoda (approved):

PR #1618, opened 2024-02-19
fix(nextcloud): Set correct intl version constraint to allow all 0.19.x versions

provokateurin:


Leptopoda (approved):
Do we have to do this manually every time or can renovate help us with it?

provokateurin:
Renovate should correctly update this version range, it was just me who did it wrong.

PR #1623, opened 2024-02-19
feat(neon_framework): Add ResultListBuilder widget

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/890

This reduces a lot of flickering because the data, error and loading states updates are rendered separately.
If the data is updated it will still cause images to reload as they are not optimized yet.

While submitting this PR I realized that now the images (only seems to affect Image.memory) stay incorrectly if an image in between is removed. I remember we already had this problem at some point, but I forgot what the fix was. I will try to dig up the fix...


provokateurin:
@Leptopoda can you review the implementation? I haven't yet found a solution to the problem with the "stuck" images.

Leptopoda (commented):
The approach looks good (I still think this is just trying to patch our loadingbar issue and does not tackle the root cause).

PR #1628, opened 2024-02-21
refactor(neon_framework): restructure NeonAppBar

Leptopoda:
I initially wanted to make the SearchBar adaptive for iOS but for various reasons this would've required further changes.

### DIFF
--- a/packages/neon_framework/lib/src/widgets/app_bar.dart
+++ b/packages/neon_framework/lib/src/widgets/app_bar.dart
@@ -19,7 +19,7 @@ import 'package:rxdart/rxdart.dart';
 
 /// Global app bar for the Neon app.
 @internal
-class NeonAppBar extends StatefulWidget implements PreferredSizeWidget {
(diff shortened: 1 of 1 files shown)


packages/neon_framework/lib/src/widgets/app_bar.dart
```diff
@@ -19,136 +19,162 @@ import 'package:rxdart/rxdart.dart';
...
-                    ),
-                  ] else ...[
-                    const SearchIconButton(),
-                  ],
-                  const NotificationIconButton(),
-                  const AccountSwitcherButton(),
-                ],
-              );
+  Widget build(BuildContext context) {
+    assert(unifiedSearchBloc.enabled.value, 'The UnifiedSearchBloc must be enabled.');
```
provokateurin, line 154: I don't think this is really necessary to check, even if it was not enabled it wouldn't do any harm.

provokateurin (approved):

Issue #1626, opened 2024-02-20
NeonUserStatusDialog broken on non material platforms (iOS, macOS)


Leptopoda:
`NeonUserStatusDialog` makes use of `ListTile` and `DropDownButton` from `flutter:material` which required the WIdget to be wrapped in a `Material` one. For the first we may use our `AdaptiveListTile` but we do not have a replacement for the later.
Another way would be to just wrap it in a `Material` widget like we already do in other dialogs.

As our current dialog is quite material we should consider just writing a separate one for cupertino.
Sadly the current iOS design from upstream is not really usable for us :/

PR #1629, opened 2024-02-21
feat(neon_framework): use adaptive icons

Leptopoda:
Just a small step towards a proper iOS design.
There are places where this change looks really good (e.g. the files app) and other ones where it currently looks a bit out of place (the notes bottom nav bar now is material design with cupertino icons).
I hope we can still land it :)

### DIFF
--- a/packages/neon_framework/lib/src/pages/login.dart
+++ b/packages/neon_framework/lib/src/pages/login.dart
@@ -5,6 +5,7 @@ import 'package:neon_framework/src/platform/platform.dart';
 import 'package:neon_framework/src/router.dart';
 import 'package:neon_framework/src/theme/branding.dart';
 import 'package:neon_framework/src/theme/dialog.dart';
+import 'package:neon_framework/src/theme/icons.dart';
 import 'package:neon_framework/src/utils/validators.dart';
 import 'package:neon_framework/src/widgets/nextcloud_logo.dart';
 
@@ -92,7 +93,7 @@ class _LoginPageState extends State<LoginPage> {
                         hintText: 'https://...',
                         labelText: NeonLocalizations.of(context).loginUsingServerAddress,
                         suffixIcon: IconButton(
-                          icon: const Icon(Icons.arrow_forward),
+                          icon: Icon(AdaptiveIcons.arrow_forward),
                           onPressed: () {
                             login(_controller.text);
                           },
@@ -110,8 +111,8 @@ class _LoginPageState extends State<LoginPage> {
                     ),
                     IconButton(
                       tooltip: NeonLocalizations.of(context).loginUsingQRcode,
-                      icon: const Icon(
-                        Icons.qr_code_scanner_rounded,
+                      icon: Icon(
+                        AdaptiveIcons.qr_code_scanner,
                         size: 60,
                       ),
                       onPressed: () => const LoginQRcodeRoute().go(context),
--- a/packages/neon_framework/lib/src/theme/icons.dart
+++ b/packages/neon_framework/lib/src/theme/icons.dart
@@ -1,0 +1,167 @@
+// ignore_for_file: non_constant_identifier_names
+
+import 'package:flutter/cupertino.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_material_design_icons/flutter_material_design_icons.dart';
+import 'package:neon_framework/src/utils/adaptive.dart';
+
+// Adding new Icons to this set:
+// Make sure to use the common doc comment style and add the correct preview.
+// If in doubt copy the previews from the icons you are redirecting to.
+
+/// A set of platform-adaptive icons.
+abstract final class AdaptiveIcons {
+  const AdaptiveIcons._();
+
+  /// Platform-adaptive icon for
+  /// <i class="material-icons md-36">search</i> &#x2014; material icon named "search"
+  /// and <i class='cupertino-icons md-36'>search</i> &#x2014; cupertino icon named "settings".
+  static IconData get search => !isCupertinoPlatform() ? Icons.search : CupertinoIcons.search;
+
+  /// Platform-adaptive icon for
+  /// <i class="material-icons md-36">settings</i> &#x2014; material icon named "settings"
+  /// and <i class='cupertino-icons md-36'>settings</i> &#x2014; cupertino icon named "settings".
+  static IconData get settings => !isCupertinoPlatform() ? Icons.settings : CupertinoIcons.settings;
+
+  /// Platform-adaptive icon for
+  /// ![Icon preview](https://fafre.github.io/flutter_material_design_icons/icons/account-plus/1.5.54/64.png "account-plus") material design icon named "accountPlus"
+  /// and <i class='cupertino-icons md-36'>person_add_solid</i> &#x2014; cupertino icon named "person_add_solid".
+  static IconData get add_account => !isCupertinoPlatform() ? MdiIcons.accountPlus : CupertinoIcons.person_add_solid;
+
+  /// Platform-adaptive icon for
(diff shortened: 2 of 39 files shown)


packages/neon/neon_notifications/lib/l10n/localizations.dart
provokateurin, line 1: This should fail the CI, but it wasn't caught

provokateurin (commented):
Should we make a script that prevents us using `Icons.` and `MdiIcons.` directly? I'm sure we and others will forget it otherwise

packages/neon/neon_notifications/lib/l10n/localizations.dart
provokateurin: https://github.com/nextcloud/neon/issues/1632

Leptopoda:
I also thought about it but it might be more complex.
1. what about MdiIcons?
2. some icons do not have an apple equivalent (double checkmark to clear all notifications)
3. some icons are only ever shown on material platforms (the file picker dialog on iOS does not show the camera, folder and other icons). In general iOS dialogs do not use icons.

I think there would be a lot of false positives

provokateurin:
Ok, but then we have to be careful when adding new icons anywhere.

Leptopoda:
For now missing one should not be an issue as we don't yet take the iOS design seriously.
But once we support it we also have to be careful with the entire UI (thinking of #1626).

packages/neon_framework/lib/src/theme/icons.dart
```diff
@@ -0,0 +1,130 @@
...
+import 'package:flutter_material_design_icons/flutter_material_design_icons.dart';
+import 'package:neon_framework/src/utils/adaptive.dart';
+
+/// A set of platform-adaptive icons.
+abstract final class AdaptiveIcons {
+  const AdaptiveIcons._();
+
+  /// Platform-adaptive icon for
+  /// <i class="material-icons md-36">search</i> &#x2014; material icon named "search"
+  /// and <i class='cupertino-icons md-36'>search</i> &#x2014; cupertino icon named "settings".;
```
provokateurin, line 14: So this is special html to make the icons appear in the docs or how does that work?

Leptopoda: copy paste from the upstream docs (the CupertinoIcons and Icons one).
I also have no idea how it works :P

provokateurin: Haha ok, I hope it is not too hard to find for new icons, maybe you can explain in a comment at the top of the file how to get this html.

Leptopoda: it's fairly simple `<i class="material-icons md-36">$iconName</i>`
I'll add a comment nevertheless.

packages/neon_framework/lib/src/pages/login.dart
```diff
@@ -92,7 +93,7 @@ class _LoginPageState extends State<LoginPage> {
                         hintText: 'https://...',
                         labelText: NeonLocalizations.of(context).loginUsingServerAddress,
                         suffixIcon: IconButton(
-                          icon: const Icon(Icons.arrow_forward),
+                          icon: Icon(Icons.adaptive.arrow_forward),
```
provokateurin, line 96: Should we still move this into the AdaptiveIcons class so we always have the same syntax? That could also avoid using different icons for the same thing in different places.

packages/neon_framework/lib/src/theme/icons.dart
```diff
@@ -0,0 +1,134 @@
...
+// If in doubt copy the previews from the icons you are redirecting to.
+
+/// A set of platform-adaptive icons.
+abstract final class AdaptiveIcons {
+  const AdaptiveIcons._();
+
+  /// Platform-adaptive icon for
+  /// <i class="material-icons md-36">search</i> &#x2014; material icon named "search"
+  /// and <i class='cupertino-icons md-36'>search</i> &#x2014; cupertino icon named "settings".
+  static IconData get search => !isCupertinoPlatform() ? Icons.search : CupertinoIcons.search;
```
provokateurin, line 19: Can you invert all the conditions? I think that is easier to understand and also indicates better that the material icons are a fallback.

packages/neon_framework/lib/src/theme/theme.dart
```diff
@@ -20,7 +20,7 @@ class AppTheme {
     this.useNextcloudTheme = false,
     this.oledAsDark = false,
     this.appThemes,
-  }) : platform = null;
+  }) : platform = TargetPlatform.iOS;
```
provokateurin, line 23: Left over from debugging?

packages/neon_framework/lib/src/utils/adaptive.dart
```diff
@@ -17,3 +20,21 @@ bool isCupertino(BuildContext context) {
       return true;
   }
 }
+
+/// Returns whether the current platform on which the framework is currently
+/// executing is a Cupertino one.
+///
+/// This is true for both `TargetPlatform.iOS` and `TargetPlatform.macOS`.
+/// Use [isCupertino] to check against the platform of the current theme.
+bool isCupertinoPlatform() {
```
provokateurin, line 29: Do we really need a separate method? Just getting this from the theme should be enough?

Leptopoda: Not really.
The theme could have `TargetPlatform.iOS` while not running on an actual iOS platform. In that case we wouldn't show any icon in release mode because the font isn't bundled (licensing thingy again).

Also how would the syntax be `AdaptiveIcons(context).search` or `AdaptiveIcons.search(context)` or `NeonIcons.adaptive(context).search`?
I find these proposed apis ugly and we would need to see how tree shaking handles this (the first example would be problematic as it isn't using static attributes).

packages/neon_framework/lib/src/pages/login.dart
Leptopoda: Sure we could add a redirecting method.

packages/neon_framework/lib/src/utils/adaptive.dart
provokateurin: I agree that adding the context is ugly, how about only using this method and removing the other one?

Leptopoda: I also thought about this (maybe even giving us some performance improvement as the platform will not change at runtime).
But as seen by my debugging leftovers above it's quite easy to do some testing through the theme. Especially in our unit tests where we rely on the ability to `mock the platform`.

packages/neon_framework/lib/src/theme/icons.dart
Leptopoda: I'm not 100% sure about that.
The flutter `PlatformAdaptiveIcon` has it exactly the same: https://api.flutter.dev/flutter/material/PlatformAdaptiveIcons/arrow_back.html

While developing I switched between the material, cupertino and adaptive and NeonIcons multiple times to find a fitting one and having them aligned is easier.
But I can see your point.

provokateurin: I see, it's your choice if you want to change it or not :woman_shrugging:

Leptopoda: That's a clear nope

provokateurin (commented):
LGTM, but what about neon_talk?

Leptopoda:
Just forgot it.
This is a broader question but should we add all the app specific icons to the framework or just have a few common ones with each app defining their own set (similar to what we do with the dialogs)?

provokateurin:
Some additional icons won't hurt much and it will make it much easier for developers, so I'd say we should put the icons into the framework whenever possible.

provokateurin (approved):

PR #1634, opened 2024-02-21
fix(neon_framework): package_info_plus overrides needed for every pac…

Leptopoda:
…kage

for whatever reason we have to override it in every package importing `neon_framework`.
This is horrific but nothing else we con do.

At least it is now tested to work ontop of #1609


provokateurin (approved):
We love duplicate code :grimacing:

PR #1639, opened 2024-02-22
refactor(neon_framework): Make wrapAction properly async

provokateurin:
This is useful to be able to change some state after the action has finished regardless of the success.
Will help a lot with https://github.com/nextcloud/neon/pull/366 which has quite complex state management and currently relies on manual try catch in many places.

### DIFF
--- a/packages/neon/neon_files/lib/src/blocs/browser.dart
+++ b/packages/neon/neon_files/lib/src/blocs/browser.dart
@@ -150,7 +150,7 @@ class _FilesBrowserBloc extends InteractiveBloc implements FilesBrowserBloc {
   }
 
   @override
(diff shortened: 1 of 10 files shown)


Leptopoda (approved):
LGTM
General question:
What should we do with the error handling? we currently catch everything but different actions might need to catch different errors. I was thinking of just catching timeout exceptions in the `wrapAction` and everything else (most of the times this would be `http.ClientException`) where wrapAction is actually called.

provokateurin:
I think just catching all is fine. If special handling for different exceptions is needed then a manual try-catch can be used. This method is just used to provide an easy way to have automatic error handling.

Issue #608, opened 2023-08-26
se a logging framework


Leptopoda:
we could then also enable the `avoid_print` linting rule.

provokateurin:
avoid_print is already enabled. To still log you need to use `debugPrint` instead.

Leptopoda:
ah yes it's only disabled for dart projects.

while `debugPrint` is an option a logging package like https://pub.dev/packages/logging would give us different log levels.

Leptopoda:
Another benefit of using a proper logging framework would be that we could silence logs in unit tests which would make them a lot cleaner.

provokateurin:
Closed in https://github.com/nextcloud/neon/pull/1642

PR #1642, opened 2024-02-23
feat(neon_framework): use logging framework instead of print

Leptopoda:
implements: #608


provokateurin:
I guess we should have a script that checks that print and debugPrint are never used in any of the neon_* packages?

packages/neon/neon_notes/lib/src/blocs/note.dart
```diff
@@ -63,10 +66,14 @@ class _NotesNoteBloc extends InteractiveBloc implements NotesNoteBloc {
         final response = await call(etag);
         emitNote(response.body);
         await notesBloc.refresh();
-      } catch (e, s) {
-        debugPrint(e.toString());
-        debugPrint(s.toString());
-        addError(e);
+      } on http.ClientException catch (error, stackTrace) {
```
provokateurin, line 69: I think we should always catch all exceptions. If a different exception is triggered then the user will get no feedback about it which is not the point of error handling.

Leptopoda: Where should a different exception come from?
`notesBloc.refresh` should handle the errors itself (in this case through the request manager) while emitNote shouldn't ever throw.
This leaves us with `call` which indicated by the return type is a dynamite request. Dynamite requests only throw `http.ClientException` (it does the error handling for us and propagates them as a ClientException).

Shure we could have a callback that does some custom stuff, throws custom objects and returns a DynamiteResponse but then the code would be awful anyways.

provokateurin: I mean it could also be a TimeoutException or SocketException. There are many different errors that can appear.

Leptopoda: Assuming that the call is always a normal nextcloud request (`account.client.news.getSomething()`) where would such an exception come from?

provokateurin: Ok I had a closer look and it seems like http package wraps all Exceptions using it's ClientException.
Then this will not be an issue, but I still think it is better to just handle all Exceptions. We gain nothing by only catching some (only worse UX) and it is much more likely that we introduce bugs when refactoring, in case a different Exception could be thrown at some point.

Leptopoda: Just so we are on the same page. I changed many catch statements where no on clause was specified to only catch the appropriate object.

Do you want to change it everywhere to catch arbitrary `Exception`s and `Error`s or just in this particular case?

provokateurin: Everywhere please. So we just catch `Exception` everywhere, right?

Leptopoda: I disagree. We should only catch the exceptions we expect to happen so unexpected issues do not get unnoticed. Error handling should happen where the error is expected and in this case only an ClientException.

imagine a scenario like:
```dart
void main() async {
  final subject = BehaviorSubject.seeded('seedValue');

  try {
    await subject.close();
    subject.add('event');
  } catch (_) {
    print('nothing wrong here');
  }
}
```

We wouldn't notice the `Bad state: Cannot add new events after calling close` (I know that this example actually throws a StateError which is not a subtype of Exception but you get my point).

My long term goal which we are fairly close to after this PR is to enable the [avoid_catches_without_on_clauses](https://dart.dev/tools/linter-rules/avoid_catches_without_on_clauses) lint rule. 
It basically says the same thing:
> Using catch clauses without on clauses make your code prone to encountering unexpected errors that won't be thrown (and thus will go unnoticed).

Also the docs on the Exception class itself (https://api.dart.dev/stable/3.3.0/dart-core/Exception-class.html) mention:
> Creating instances of Exception directly with Exception("message") is discouraged in library code since it doesn't give users a precise type they can catch.


In most literature about error handling one would find the advice "do not catch generic errors" and I'm hesitant in ignoring this advice.

provokateurin: I get the point of doing that, but I still think the UX will suffer. We can go this route, but I'm sure at some point we will have bug reports saying that something doesn't work but there is also no error displayed (would only appear in the logs).

Leptopoda: for this I'd like to use `FlutterError.onError` and the `PlatformDispatcher.instance.onError` callbacks (the same used by firebase_crashlytics, sentry and probably other similar services).
They are built for that exact usecase:
https://docs.flutter.dev/testing/errors
https://docs.flutter.dev/cookbook/maintenance/error-reporting (probably not so important for our usecase)

Leptopoda:
Not sure if you already had a look at the latest commit but I've tried to address the suggestion from your other https://github.com/nextcloud/neon/pull/1642#issuecomment-1962100116 I've opted to use the [custom_lint](https://pub.dev/packages/custom_lint) package which provides an abstraction to write custom lint rules using the analyzer api (similar to what the analyzer itself uses).

I had to bump the project sdk version to 3.2.0 (not the for the packages though) and the analyzer plugin api is still experimental. You can read more about it here https://dart.dev/tools/analysis#plugins.
There are some limitations like recommended 16GB of RAM and and not more than 10 packages in a monorepo. I'd say let's see how far we can go with it and if desired we can still switch it to regexes.

A benefit of this approach is that we can write custom quickFixes for things like `dart:io`=> `universal_io`, higher accuracy as we can use the analyzer directly and direct feedback while developing (I never run these scripts before submit :sweat_smile:)

provokateurin:
Yeah I really liked that you added this, it looks like a pretty cool feature to use. I don't like that we have to use a custom script to do the analyzing, but I guess that's a trade off worth it.

packages/neon/neon_notes/lib/src/blocs/note.dart
provokateurin: Right, let's do that

Leptopoda: I'll open an issue and add it to the release milestone. Is that ok?

provokateurin: I feel like that needs to be solved before we can merge this, otherwise we will miss some of the exceptions (in case there is a programming error somewhere).

provokateurin: Ok so we will do the global error handling later.

packages/neon_lints/lib/neon_lints.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:neon_lints/src/amend_model_suffix.dart';
+
+/// Registers the neon lints plugin.
+///
+/// Do **not** use this dierectly.
```
provokateurin, line 6:
```suggestion
/// Do **not** use this directly.
```

packages/neon_lints/lib/neon_lints.dart
```diff
@@ -0,0 +1,16 @@
...
+/// This method is applied used when enabling the `custom_lint` in the
```
provokateurin, line 7:
```suggestion
/// This method is used when enabling the `custom_lint` in the
```

packages/neon_lints/lib/src/amend_model_suffix.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+
+/// Lint rule checking that no print methods from `flutter:framework` are used.
+///
+/// Similar to the analyzer rule `avoid_print`.
+/// Instead use a a logging framework like `https://pub.dev/packages/logging`.
```
provokateurin, line 8:
```suggestion
/// Use `https://pub.dev/packages/logging` instead.
```

packages/neon_lints/lib/src/amend_model_suffix.dart
```diff
@@ -0,0 +1,41 @@
...
+///   debugPrintSynchronously('message'); // lint
+///   debugPrintThrottled('message'); // lint
+/// ```
+final class AvoidDebugPrint extends DartLintRule {
+  /// @nodoc
+  const AvoidDebugPrint() : super(code: _code);
+
+  static const _code = LintCode(
+    name: 'avoid_debug_print',
+    problemMessage: "Don't invoke 'debugPrint' and 'debugPrintStack' in production code.",
```
provokateurin, line 22:
```suggestion
    problemMessage: "Do not invoke 'debugPrint' and 'debugPrintStack' in production code.",
```

packages/neon_lints/lib/src/amend_model_suffix.dart
```diff
@@ -0,0 +1,41 @@
...
+    correctionMessage: 'Try using a logging framework.',
```
provokateurin, line 23:
```suggestion
    correctionMessage: "Use 'package:logging' instead.",
```

packages/neon_lints/lib/src/amend_model_suffix.dart
provokateurin, line 1: The file name is a bit odd, is this a mistake?

packages/neon_framework/lib/src/blocs/next_push.dart
```diff
@@ -59,9 +62,13 @@ class _NextPushBloc extends Bloc implements NextPushBloc {
           try {
             final response = await account.client.uppush.check();
             supported[account] = response.body.success;
-          } catch (e, s) {
-            debugPrint(e.toString());
-            debugPrint(s.toString());
+          } on http.ClientException catch (error, stackTrace) {
+            _log.warning(
+              'Error checking unified push availability.',
```
provokateurin, line 67:
```suggestion
              'Error checking NextPush availability.',
```

packages/neon/neon_files/lib/src/blocs/files.dart
```diff
@@ -216,7 +219,7 @@ class _FilesBloc extends InteractiveBloc implements FilesBloc {
     final file = File(p.join(cacheDir.path, 'files', etag.replaceAll('"', ''), uri.name));
 
     if (!file.existsSync()) {
-      debugPrint('Downloading $uri since it does not exist');
+      _log.info('Downloading $uri since it does not exist');
```
provokateurin, line 222:
```suggestion
      _log.fine('Downloading $uri since it does not exist');
```

packages/neon_framework/lib/src/blocs/user_status.dart
```diff
@@ -149,10 +152,14 @@ class _UserStatusBloc extends InteractiveBloc implements UserStatusBloc {
       }
 
       updateStatus(username, Result.success(correctStatus(data)));
-    } on DynamiteStatusCodeException catch (e, s) {
-      debugPrint(e.toString());
-      debugPrint(s.toString());
-      updateStatus(username, Result.error(e));
+    } on http.ClientException catch (error, stackTrace) {
+      _log.warning(
+        'Error loading the user statuses.',
```
provokateurin, line 157:
```suggestion
        'Error loading the user status.',
```

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -280,34 +283,43 @@ class RequestManager {
 
         await _cache?.set(key, serialized, cacheParameters);
         break;
-      } on TimeoutException catch (e, s) {
-        debugPrint('Request timed out ...');
-        debugPrint(e.toString());
-        debugPrintStack(stackTrace: s, maxFrames: 5);
+      } on TimeoutException catch (error, stackTrace) {
+        _log.info(
+          'Request timed out ...',
```
provokateurin, line 288:
```suggestion
          'Request timeout',
```

provokateurin (changes requested):
Two things:
1. In a lot of places where errors are caught only the info or warning levels are used, but they should be severe instead. I know that some places can stay info or warning, but most should really be severe.
2. My IntelliJ doesn't pick up the custom linting rule, do I need to change something so it works? (Yes I restarted the analyzer and the whole IDE).

packages/neon_lints/lib/src/amend_model_suffix.dart
Leptopoda: copied from the example docs. Thanks for catching.

Leptopoda:
I don't get any linting in my IDEA (not set up correctly) but in android studio it works.
Have you also activated the custom_lint package globally? Maybe run the setup script again.

Leptopoda:
> In a lot of places where errors are caught only the info or warning levels are used, but they should be severe instead. I know that some places can stay info or warning, but most should really be severe.

This is intentional. We did handle the error in the catch clause so just a warning that a process went wrong. I'd only use severe in the global error handler for uncaught errors.
I quite like the graphic in this post https://stackoverflow.com/a/64806781.

provokateurin:
>  Have you also activated the custom_lint package globally? Maybe run the setup script again.

I'm not sure what I did before, but now I ran the setup script, restarted IntelliJ and still nothing.

packages/neon_framework/lib/src/bloc/bloc.dart
```diff
@@ -63,9 +66,13 @@ abstract class InteractiveBloc extends Bloc {
...
       await (refresh ?? this.refresh)();
-    } catch (error, stacktrace) {
-      debugPrint(error.toString());
-      debugPrint(stacktrace.toString());
+    } on Exception catch (error, stackTrace) {
+      _log.info(
+        'An Exception has occurred executing an action.',
+        error,
+        stackTrace,
+      );
```
provokateurin, line 74: Instead of using it's own logger the method should get the logger passed. Otherwise the messages are really not helpful and don't make a lot of sense.

provokateurin:
> I quite like the graphic in this post [stackoverflow.com/a/64806781](https://stackoverflow.com/a/64806781).

Can you add this to our docs? I like the graphic as well.

Leptopoda:
where would this fit?
Maybe a "how to write a client" where we explain package choices and such conventions? Or just a `code_style.md`?

packages/neon_framework/lib/src/bloc/bloc.dart
provokateurin: Not resolved?

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -280,34 +283,43 @@ class RequestManager {
...
-              error: e,
+              error: error,
               isLoading: false,
             ),
           );
           break;
         }
       }
-
-      debugPrint('Retrying...');
```
provokateurin, line 310: This message should stay here because it is the only place we reliably know that we are going to retry without logging in multiple places.

packages/neon_framework/lib/src/bloc/bloc.dart
provokateurin: Oh it looks like you fixed it, but added it to the wrong commit.

packages/neon_lints/lib/src/avoid_debug_print.dart
provokateurin, line 1: Would it be possible to write a test for this to ensure it doesn't break?

provokateurin (changes requested):
LGTM, just some minor things and you added the requested changes to the wrong commit.

packages/neon_lints/lib/src/avoid_debug_print.dart
Leptopoda: yes but only in a package separate from neon_lints.
Should I add it to the framework?

For the record it is possible to write tests in the lints package but this is quite hard to do.

packages/neon_framework/lib/src/bloc/bloc.dart
Leptopoda: :woman_facepalming:

packages/neon_lints/lib/src/avoid_debug_print.dart
provokateurin: I think we will be fine for now, if it breaks we will hopefully notice it in the code reviews and even if that fails then it's not very dramatic.

Leptopoda: it is dead simple to test if not in the same package. Just add a ignore style comment `// expect_lint: avoid_debug_print` which suppresses the lint but complains if the code does not produce the expected lint.

Sadly I was not able to make it work in the defining package.

provokateurin: Ok if it is so simple then please add a simple test to the framework

provokateurin (approved):
Thanks a lot :heart:

Can you add the simple test for the lint?

PR #1647, opened 2024-02-23
fix(nextcloud): Add missing share types

provokateurin:
Aligned with https://github.com/nextcloud/server/blob/master/lib/public/Share/IShare.php
I had to use the value backed enum because it seems like 14 was just skipped. In the future we might also want to just remove the deprecated values, so this prepares us for that as well.

### DIFF
--- a/packages/nextcloud/lib/src/helpers/core.dart
+++ b/packages/nextcloud/lib/src/helpers/core.dart
@@ -37,6 +37,7 @@ extension CoreStatusVersionCheck on core.Status {
       );
 }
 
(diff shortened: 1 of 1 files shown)


Leptopoda (commented):
Can we do this in a non breaking way?
Maybe by adding a `_notUsed` element that isn't visible due to the underscore?

We should change it in the next major release but I don't see it warranted just for this.

Leptopoda (approved):
Can you open an issue and add it to the nc milestone so we don't forget to change this for the next relese?

provokateurin:
TBH I would just keep it like this and not make any further changes.

PR #1657, opened 2024-02-25
feat(neon_talk): Add room list

provokateurin:
Extracted/rewritten from https://github.com/nextcloud/neon/pull/366
Towards https://github.com/nextcloud/neon/issues/356


packages/neon/neon_talk/lib/src/widgets/unread_indicator.dart
```diff
@@ -0,0 +1,54 @@
...
+    final Color borderColor;
+
+    if (room.unreadMentionDirect || spreed.RoomType.fromValue(room.type).isSingleUser) {
+      textColor = Theme.of(context).colorScheme.onPrimary;
+      backgroundColor = Theme.of(context).colorScheme.primary;
+      borderColor = Theme.of(context).colorScheme.onBackground;
+    } else {
+      textColor = Theme.of(context).colorScheme.onBackground;
+      backgroundColor = null;
+      borderColor = room.unreadMention ? Theme.of(context).colorScheme.primary : Colors.grey;
```
Leptopoda, line 32: Onto the styling.
I think that using the primary color is a bit to much. I think we should use `primaryContainer`. It is currently not used by the nc theme but when the user uses the native m3 design it makes a big difference.

---

Also imo the different colors don't make clear that one involves a mention and one doesn't. We should maybe make mentions clearer by using an `@` sign.

This is how signal handles it (if only one unread message exists and it is a mention only the `@` symbol is shown). 
signal-2024-02-25-201324

Leptopoda (changes requested):

Leptopoda:
Mostly UI stuff and some minor code style.
I know that you are trying to get things done one step at a time but let's try to also do the UI right from the beginning.

packages/neon/neon_talk/lib/src/pages/main.dart
```diff
@@ -1,10 +1,65 @@
...
+      );
+
+  Widget buildMessagePreview(spreed.ChatMessage? chatMessage) {
+    final buffer = StringBuffer();
+
+    if (chatMessage != null) {
+      buffer.write('${chatMessage.actorDisplayName}: ${chatMessage.message}');
+    }
+
+    return Text(
```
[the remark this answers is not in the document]
provokateurin: This was on purpose to keep this PR small and reviewable. I will improve this in follow up PRs.

packages/neon/neon_talk/lib/src/pages/main.dart
```diff
@@ -1,10 +1,65 @@
...
+      buffer.toString(),
+      maxLines: 1,
+    );
+  }
```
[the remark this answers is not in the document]
provokateurin: You mean the message preview or the list tile as well?

packages/neon/neon_talk/lib/src/pages/main.dart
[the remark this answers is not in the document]
provokateurin: Maybe it is better then to remove the message preview from this PR and only add it properly later.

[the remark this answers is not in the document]
provokateurin: This was on purpose to keep the PR small, but I'll implement it now in a separate commit.

packages/neon/neon_talk/lib/src/widgets/unread_indicator.dart
provokateurin: I really like the `@`, I never noticed that Signal does this :sweat_smile: I just copied the behaviour of the Talk web client, but this is definitely much better.

packages/neon/neon_talk/lib/src/pages/main.dart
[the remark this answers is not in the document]
Leptopoda: Yes the entire list tile.
But if we go the signal approach as mentioned below the new `RoomListTile` should also handle the row layout of the multiple chips. This layout should not happen in the UnreadIndicator itself.

I'm not sure if we still need :thinking:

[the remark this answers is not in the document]
provokateurin: I think it makes sense to have it in the indicator widget. We can move it later on if needed.

provokateurin:
I implemented the unread indicator a bit different than you suggested, but I think it works pretty well this way.
Please note that the actual rendering of the message is not implemented yet (e.g. mentions or files won't work). That will be done in a different PR when the actual chat layout will be implemented.

provokateurin:
Hm should the indirect mention also show the @? I'm not sure how this is best handled, as `@room` can sometimes be very important and sometimes be used for no good reason. Maybe this should be implemented as an option to be let the user chose what they prefer.

packages/neon/neon_talk/lib/src/widgets/message_preview.dart
```diff
@@ -0,0 +1,60 @@
...
+              children: [
+                TextSpan(
+                  text: actorName,
+                  style: const TextStyle(
+                    fontWeight: FontWeight.bold,
+                  ),
+                ),
+                const TextSpan(
+                  text: ': ',
+                ),
```
Leptopoda, line 50: The `:` should also be bold

packages/neon/neon_talk/lib/src/widgets/unread_indicator.dart
```diff
@@ -0,0 +1,56 @@
...
+
+    final colorScheme = Theme.of(context).colorScheme;
+
+    final highlight = room.unreadMention || spreed.RoomType.fromValue(room.type).isSingleUser;
+    final backgroundColor = highlight ? colorScheme.primaryContainer : colorScheme.background;
+    final textColor = highlight ? colorScheme.onPrimaryContainer : colorScheme.onBackground;
+
+    return Row(
+      mainAxisSize: MainAxisSize.min,
+      children: [
```
Leptopoda, line 27: Leftovers

packages/neon/neon_talk/lib/src/widgets/unread_indicator.dart
```diff
@@ -0,0 +1,56 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    assert(room.unreadMessages > 0, 'Need at least on unread message');
+
+    final colorScheme = Theme.of(context).colorScheme;
+
+    final highlight = room.unreadMention || spreed.RoomType.fromValue(room.type).isSingleUser;
+    final backgroundColor = highlight ? colorScheme.primaryContainer : colorScheme.background;
+    final textColor = highlight ? colorScheme.onPrimaryContainer : colorScheme.onBackground;
```
Leptopoda, line 23: back to the design :sweat_smile: 
1. `primaryContainer` looks good in light mode but I really feal shouted at when looking at it in dark mode. I think `secondaryContainer` would be better. It looks the same in light mode but better when in dark.
2. I'm not sure why we should distinguish between highlighted and not highlighted messaged. For me all messages are equally important. Does this make more sense in a corporate environment where nc talk is most often used?
3. regarding the `@` in room mentions. I agree that this should probably be configurable. But let's do it in a separate PR. I think that this (and probably other stuff) should be stored on a per room basis but I don't know how our current storage would handle this.
Maybe we should somehow expose one persistence per app that can be used for such purposes?

packages/neon/neon_talk/lib/src/blocs/talk.dart
```diff
@@ -1,19 +1,71 @@
...
+    await RequestManager.instance.wrapNextcloud(
+      account: account,
+      cacheKey: 'talk-rooms',
+      subject: rooms,
+      rawResponse: account.client.spreed.room.getRoomsRaw(),
+      unwrap: (response) => BuiltList(
+        response.body.ocs.data.sorted(
+          (a, b) => b.lastActivity.compareTo(a.lastActivity),
+        ),
+      ),
```
Leptopoda, line 68: Let's sort the BuiltList directly.
In general the more operations we can do on the builder the more efficient it should be.

```suggestion
   unwrap: (response) => response.body.ocs.data.rebuild(
        (b) => b.sort((a, b) => b.lastActivity.compareTo(a.lastActivity)),
      ),
```

packages/neon/neon_talk/lib/src/widgets/unread_indicator.dart
provokateurin: About 2: Yeah I think the reasoning is that you can get quite a lot of messages when you are in a lot of groups and therefore messages that are considered more important are highlighted. We can also chose to not follow this, but I don't see why we shouldn't adopt it too.

provokateurin: About 1: I see your point, but I think `secondaryContainer` is very ugly in dark mode. I would like to just stick with `primaryContainer`.

provokateurin: About 3: Yeah we'll need something, but we can do it in a later PR.

packages/neon/neon_talk/lib/src/pages/main.dart
```diff
@@ -1,10 +1,81 @@
...
+  late TalkBloc bloc;
+  late StreamSubscription<Object> errorsSubscription;
+
+  @override
+  void initState() {
+    super.initState();
+
+    actorId = NeonProvider.of<AccountsBloc>(context).activeAccount.value!.username;
+    bloc = NeonProvider.of<TalkBloc>(context);
+    bloc.errors.listen((error) {
```
Leptopoda, line 32:
```suggestion
    errorsSubscription = bloc.errors.listen((error) {
```

Leptopoda (approved):
You forgot to assign the subscription.
Other than that let's get it done.

PR #1664, opened 2024-02-26
fix(neon_framework): User status DND takes precedence over emoji

provokateurin:


Leptopoda:
Shouldn't be that hard to add unit tests or am I mistaken?
If so please do so.

provokateurin:
Yeah, but it will require adding tests for the whole widget which I didn't have the time for yet.

provokateurin:
@Leptopoda I added tests (and split up the logic so it is easier to test).

Leptopoda (approved):

PR #1668, opened 2024-02-27
refactor(neon_framework): Allow using NeonUserAvatar without specifying the account

provokateurin:
Will help with https://github.com/nextcloud/neon/pull/1661.
NeonApiImage and NeonUriImage already provide this kind of API, so this just makes it easier to use all of them when the used account is the currently active one.


Leptopoda (approved):
Nice cleanup.
Didn't test but code looks good.

PR #1670, opened 2024-02-27
feat(neon_framework): Implement deleting accounts on the server

provokateurin:
Closes https://github.com/nextcloud/neon/issues/732
We can not directly use the endpoint because of https://github.com/nextcloud/neon/issues/1492, so I opted for implementing it like the mobile clients which just open the settings page in the browser. For that reason I also patched the spec to remove the endpoint as it is not possible to use it (yet).


Leptopoda:
I think another action in the appbar is confusing for users.
I think we should steal the desing of the official client https://github.com/nextcloud/android/pull/12015.

A dialog with two options.

tcitworld:
Nice !
Also, the UI could benefit from the information exposed by the capacities:
- the `delay` field saying whether the user data will be deleted straight away or after a while, which allows the accounts to be restored
- the `details` field saying why the account can't be deleted if applicable

provokateurin:
The details field is already used, but the delay isn't yet. I'll rewrite the UI to use the dialog and then those can also be displayed in a nice way.

Leptopoda:
@provokateurin maybe don't make the dialogue adaptive.

The RadioListTile will require a material ancestor and even if we wrap it in a material widget it'll look out of place on iOS anyways.

provokateurin:
The CI failure is due to https://github.com/nextcloud/server/pull/43877 and I hope I can get it into 28.0.3, otherwise we can just manually patch it for the time being (or update to a mid-release 28 snapshot).

Leptopoda:
So I'll need to manually install the app into the docker container to test it?
Looks like it isn't on the store (or am i missing something).

packages/neon_framework/lib/src/pages/account_settings.dart
```diff
@@ -45,32 +46,37 @@ class AccountSettingsPage extends StatelessWidget {
...
-              if (isActive) {
-                const HomeRoute().go(context);
-              } else {
-                Navigator.of(context).pop();
-              }
+                if (isActive) {
+                  const HomeRoute().go(context);
+                } else {
+                  Navigator.of(context).pop();
+                }
```
Leptopoda, line 79: Does this mean the local account is not deleted when the user selects a remote deletion?

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -516,6 +518,131 @@ class NeonAccountSelectionDialog extends StatelessWidget {
   }
 }
 
+@internal
+enum AccountDeletion {
```
Leptopoda, line 522: missing doc comments

packages/neon_framework/lib/src/widgets/dialog.dart
```diff
@@ -516,6 +518,131 @@ class NeonAccountSelectionDialog extends StatelessWidget {
...
+  local,
+  remote,
+}
+
+@internal
+class NeonAccountDeletionDialog extends StatefulWidget {
```
Leptopoda, line 528: missing doc comments

provokateurin:
> So I'll need to manually install the app into the docker container to test it?

Just rebuild the docker images and delete your volumes.

>  Looks like it isn't on the store (or am i missing something).

https://apps.nextcloud.com/apps/drop_account it is, but you can't search for app IDs in the store...

packages/neon_framework/lib/src/pages/account_settings.dart
provokateurin: Yes, because the user could still cancel the account deletion in the browser.
We could change the dialog to only have a checkbox (in case the app is available) and the user would be logged out in any case, but taken to the web page after being logged out. I'm not sure if that is a better UX, but it would be a viable alternative.

Leptopoda: Once we support in app deletion requests we should also use log the user out.
I think it might also be expected now but I'll leave that decision to you.

provokateurin: Ok then I'll keep it this way for now and once we can directly delete the user from within the app we should log them out as well.

Leptopoda (approved):

provokateurin:
Slow CI + a lot of random failure :smiling_face_with_tear:

PR #1677, opened 2024-02-28
refactor(dynamite_runtime): split dynamite_client into separate libra…

Leptopoda:
…ries

Extracted from: #1580
There shouldn't be any breaking changes but please review this and the following dynamite_runtime changes carefully.


packages/dynamite/dynamite_runtime/lib/src/client/request.dart
provokateurin, line 1: Empty file?

provokateurin (approved):
For testing I tried replicating the diff and it looks good, just some changed comments and a method that is only visible for testing anyway :+1:

PR #1678, opened 2024-02-29
feat(dynamite_runtime): make DynamiteClient extend http.Client

Leptopoda:

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/client/client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/client/client.dart
@@ -6,6 +6,7 @@ import 'package:dynamite_runtime/http_client.dart';
 import 'package:dynamite_runtime/src/utils/debug_mode.dart';
 import 'package:dynamite_runtime/src/utils/uri.dart';
 import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
 
 /// A client for making network requests.
 ///
@@ -14,7 +15,7 @@ import 'package:http/http.dart' as http;
 ///   * [DynamiteRawResponse] as the raw response that can be serialized.
 ///   * [DynamiteApiException] as the exception that can be thrown in operations
 ///   * [DynamiteAuthentication] for providing authentication methods.
-class DynamiteClient {
+class DynamiteClient with http.BaseClient {
   /// Creates a new dynamite network client.
   ///
   /// If [httpClient] is not provided a default one will be created.
@@ -42,6 +43,7 @@ class DynamiteClient {
   final Map<String, String>? baseHeaders;
 
   /// The base http client.
+  @protected
   final http.Client httpClient;
 
   /// The optional cookie jar to persist the response cookies.
@@ -85,10 +87,6 @@ class DynamiteClient {
   }) async {
     final request = http.Request(method, uri);
 
-    if (baseHeaders != null) {
-      request.headers.addAll(baseHeaders!);
-    }
-
     if (headers != null) {
       request.headers.addAll(headers);
     }
@@ -97,20 +95,7 @@ class DynamiteClient {
       request.bodyBytes = body;
     }
 
-    if (cookieJar != null) {
-      final cookies = await cookieJar!.loadForRequest(uri);
-      if (cookies.isNotEmpty) {
-        request.headers['cookie'] = cookies.join('; ');
-      }
-    }
-
-    final response = await httpClient.send(request);
-
-    final cookieHeader = response.headersSplitValues['set-cookie'];
-    if (cookieHeader != null && cookieJar != null) {
-      final cookies = cookieHeader.map(Cookie.fromSetCookieValue).toList();
-      await cookieJar!.saveFromResponse(uri, cookies);
-    }
+    final response = await sendWithCookies(request);
 
     if (validStatuses?.contains(response.statusCode) ?? true) {
       return response;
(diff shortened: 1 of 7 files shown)


packages/dynamite/dynamite_runtime/lib/src/client/client.dart
```diff
@@ -97,8 +95,28 @@ class DynamiteClient {
...
+      } else {
+        throw DynamiteStatusCodeException(response.statusCode);
+      }
+    }
+  }
+
+  @override
+  Future<http.StreamedResponse> send(http.BaseRequest request) async {
+    baseHeaders?.forEach((key, value) {
+      request.headers.putIfAbsent(key, () => value);
```
provokateurin, line 115: Shouldn't the base headers always override the request headers? At least that is how it worked so far. I'm not sure which one is usually the expected behavior.

packages/dynamite/dynamite_runtime/lib/src/utils/byte_stream_extension.dart
provokateurin, line 1: This extension isn't used yet, right? At least I don't see any changes that seem to use it.

provokateurin (commented):
LGTM, just two questions

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
Leptopoda: 1. so far the request headers did overwrite the base ones as we added them later on, so no change here:
```dart

    if (baseHeaders != null) {
      request.headers.addAll(baseHeaders!);
    }

    if (headers != null) {
      request.headers.addAll(headers);
    }
```
> If a key of other is already in this map, its value is overwritten.

https://api.dart.dev/stable/3.3.0/dart-core/Map/addAll.html

2. I think that this is the right thing to do as a user might add arbitrary headers to the baseClient that are not anticipated by openapi spec potentially leading in invalid requests.
Therefore a request should always set the headers it requires (I think we did already discuss this once :thinking:)

packages/dynamite/dynamite_runtime/lib/src/utils/byte_stream_extension.dart
Leptopoda: Oh that commit somehow slipped in.
Should I remove it and add it to the next PR or can we leave it?

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
provokateurin: Ok makes sense and yeah we probably had this discussion already :D
Maybe you can add a comment explaining it, so I won't ask again next time.

packages/dynamite/dynamite_runtime/lib/src/utils/byte_stream_extension.dart
provokateurin: I think it is better to remove it and only add it when it will be actually used.

Leptopoda:
We'll need to enable the CSRF hack for all platforms as this PR will send the cookies on every request.

packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -75,35 +72,35 @@ class WebDavClient {
...
+      request.headers.addAll(
+        authentication.headers,
+      );
+    }
 
     // On web we need to send a CSRF token because we also send the cookies.  In theory this should not be required as
     // long as we send the OCS-APIRequest header, but the server has a bug that only triggers when you also send the
     // cookies. On non-web platforms we don't send the cookies so we are fine, but on web the browser always does it
     // and therefore we need this workaround.
     // TODO: Fix this bug in server.
```
provokateurin, line 86:
```suggestion
    // We need to send a CSRF token because we also send the cookies.  In theory this should not be required as
    // long as we send the OCS-APIRequest header, but the server has a bug that only triggers when you also send the cookies.
    // TODO: Fix this bug in server.
```

provokateurin (approved):
Only the comment needs to be adjusted

provokateurin:
Fixtures need to be adjusted for the new headers

provokateurin:
I think it would be better to use `+` instead of `*` for the requesttoken regex because that will ensure that there is at least one character.

Leptopoda:
To be correct we could also check that it is two base64 characters separated by a ':' but tbh this is more readable.

What do you think?

provokateurin:
That would be too much, we should only verify that it is not empty.

Leptopoda:
For whatever reason webdav failed on 26 when using a fresh `http.Client` for the token request while it worked with the testing client from `nextcloud_test` (the rootClient directly).
Using the rootClient directly was also not an option as it would randomly add the token request to the fixture of the first run test.

For now I just added a separate method to do a request without cookies.

provokateurin:
But that would indicate it's not working on web for 26 :thinking: This doesn't seem right to me and don't like the fix.

provokateurin (changes requested):
Blocking for now

Leptopoda:
it should work on web now as it is using the rootClient directly which did work in the tests.
It was just not feasible for our fixtures.

Feel free to debug this further but maybe we should do this in a separate issue. There is no behavior change in this PR so I don't think we should t block it.

Leptopoda:
The fixtures where a problem as the request made to obtain the token 
https://github.com/nextcloud/neon/blob/b9e837a3855eb16dda149bba1a651d6e120d7377/packages/nextcloud/lib/src/webdav/client.dart#L90
would also create an entry in a fixture.
Because the token is cached for all tests within a preset we would only add it to the fixture of the first test of a preset.

This would make the fixtures dependent on the execution order of our our tests which is bad.

provokateurin:
OK I see the problem, but since this is only a problem of our testing code it should be fixed there and not in the actual library. So the fixture stuff needs to be adjusted to exclude the request to get the token.

Leptopoda:
Sure we can do this as a followup where we enable cookies on all platforms.
For now I would rather not add the cookies for every request as we also allow the client to make arbitrary requests to third party servers (RequestManager.wrapUri and RequestManager.wrapBinary).
I'd rather not persist cookies of such third party sited as they might be used for tracking.

What do you think?

provokateurin:
Yeah that makes sense.

provokateurin (approved):

PR #1681, opened 2024-02-29
feat(dynamite)!: Show field names in oneOf/anyOf validation errors

provokateurin:
An error message like `At least one of [null, null, null, null, null, null, null, null, null, null, null, null, null, null, null] must not be null` for anyOf is really not helpful and we should show the field names instead.


Leptopoda (approved):
LGTM

PR #1689, opened 2024-03-01
Refactor/dynamite/prepare_httpclient_request_flow

Leptopoda:
refactors extracted from: #1580
Motivation:
This decouples the resolving and generation of our parameter code.
The new code first resolves everything into a collections that can later be used to generate code.
```dart
      final operationParameters = <Parameter>[];
      final pathParameters = <openapi.Parameter, TypeResult>{};
      final headerParameters = <openapi.Parameter, TypeResult>{};
```
This is needed to keep further behavior changing changes easy to review.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/client.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/client.dart
@@ -305,7 +305,7 @@ Iterable<Method> buildTags(
           ),
         );
 
-        code.writeln(buildParameterSerialization(result, parameter, state));
+        code.writeln(buildParameterSerialization(result, parameter, state, state.emitter.allocator.allocate));
       }
       resolveMimeTypeEncode(operation, spec, state, operationName, operationParameters).forEach(code.writeln);
 
@@ -487,6 +487,7 @@ String buildParameterSerialization(
   TypeResult result,
   openapi.Parameter parameter,
   State state,
+  String Function(Reference) allocate,
 ) {
   final $default = parameter.schema?.$default;
   var defaultValueCode = $default?.value;
@@ -515,18 +516,9 @@ String buildParameterSerialization(
   }
 
   if (parameter.$in == openapi.ParameterType.header) {
-    final assignment = refer('_headers')
-        .index(literalString(parameter.pctEncodedName))
-        .assign(
-          refer('HeaderEncoder', 'package:dynamite_runtime/utils.dart')
-              .constInstance(const [], {
-                'explode': literalBool(parameter.explode),
-              })
-              .property('convert')
-              .call([refer(serializedName)]),
-        )
-        .statement
-        .accept(state.emitter);
+    final encoderRef = refer('HeaderEncoder', 'package:dynamite_runtime/utils.dart');
+    final assignment =
+        "_headers['${parameter.pctEncodedName}'] = ${allocate(encoderRef)}(explode: ${parameter.explode}).convert($serializedName);";
 
     if ($default == null) {
       buffer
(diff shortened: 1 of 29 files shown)


packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -556,62 +555,66 @@ bool needsAuthCheck(
   return securityRequirements.isNotEmpty;
 }
 
-Iterable<String> buildAuthCheck(
+StringSink buildAuthCheck(
```
provokateurin, line 633: The return value makes no sense because it is never used. You just write to the output directly.

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -515,18 +516,9 @@ String buildParameterSerialization(
...
-                'explode': literalBool(parameter.explode),
-              })
-              .property('convert')
-              .call([refer(serializedName)]),
-        )
-        .statement
-        .accept(state.emitter);
+    final encoderRef = refer('HeaderEncoder', 'package:dynamite_runtime/utils.dart');
+    final assignment =
+        "_headers['${parameter.pctEncodedName}'] = ${allocate(encoderRef)}(explode: ${parameter.explode}).convert($serializedName);";
```
provokateurin, line 604: I don't understand why this is better than before, especially because it doesn't use code_builder anymore?

packages/dynamite/dynamite/lib/src/builder/client.dart
```diff
@@ -430,17 +335,108 @@ return rawResponse.future;
...
+                    'validStatuses': literalConstSet(statusCodes),
+                }),
+                'bodyType': refer(bodyType.fullType ?? 'null'),
+                'headersType': refer(headersType.fullType ?? 'null'),
+                'serializers': refer(r'_$jsonSerializers'),
+              });
+
+              code.writeln(returnValue.returned.statement.accept(state.emitter));
+              return code.toString();
+            });
```
provokateurin, line 448: Would it be possible to avoid changing the order of the import aliases this change introduces?

provokateurin (commented):
LGTM overall

Typo in the commit message: `refactor(dynamite): split mimetiype resolving and parameter generation` should be `refactor(dynamite): split mimetype resolving and parameter generation`

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: I have a local branch where ~60% of the client is generated with using code_builder but sadly this tanked the performance. 
I'd have to test it again but iirc the build time increased from ~1.5m to ~15m so I decided that this sadly isn't the way to go :(

I can keep it for now if you prefer.

packages/dynamite/dynamite/lib/src/builder/client.dart
Leptopoda: We can try and play with the generation order but if not now it'll probably come in the following PR.
Let me rebase and check.

Leptopoda: I just rebased and can confirm that the order is going to stay

packages/dynamite/dynamite/lib/src/builder/client.dart
provokateurin: Ugh, then we should keep this change. Should be remove all other code_builder usage as well? Maybe that will make everything much faster?

Leptopoda: I'm not 100% sure. I think we'd need to conduct some further testing to be sure.
I'll keep the removal here for now as it also makes the code more readable.

provokateurin (approved):

PR #1690, opened 2024-03-02
fix(neon_framework): Use correct adapative icons

provokateurin:
Some of the icons and the documentation were not correct. I should have paid closer attention when reviewing it originally :sweat_smile:


Leptopoda (approved):
:sweat_smile:

PR #1694, opened 2024-03-02
refactor: Only use TestApp in tests

provokateurin:


Leptopoda (approved):
I'M not a fan of changing unit test all the time.

provokateurin:
I agree, but this was a necessary change to align all tests and not use this weird stuff from before.

PR #1695, opened 2024-03-02
Feat/dynamite_runtime/dynamite_serializer

Leptopoda:
Next step from: #1580

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -77,28 +77,26 @@ class RequestManager {
     required Account account,
     required String cacheKey,
     required BehaviorSubject<Result<T>> subject,
-    required DynamiteRawResponse<B, H> rawResponse,
+    required FutureOr<DynamiteRawResponse<B, H>> rawResponse,
     required DynamiteSerializer<B, H> serializer,
     required UnwrapCallback<T, DynamiteResponse<B, H>> unwrap,
     bool disableTimeout = false,
   }) async =>
-      wrap<T, DynamiteRawResponse<B, H>>(
+      wrap<T, DynamiteResponse<B, H>>(
         account: account,
         cacheKey: cacheKey,
         subject: subject,
-        request: () async {
-          await rawResponse.future;
+        request: () async => rawResponse,
+        unwrap: (rawResponse) => unwrap(rawResponse),
+        serialize: (data) {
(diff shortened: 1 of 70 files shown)


packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -77,6 +78,7 @@ class RequestManager {
     required String cacheKey,
     required BehaviorSubject<Result<T>> subject,
     required DynamiteRawResponse<B, H> rawResponse,
+    required DynamiteSerializer<B, H> serializer,
```
provokateurin, line 81: I don't like that we have to pass the serializer every time. Could we get it somehow from the raw response?

provokateurin: I guess not, after the raw response has been turned into a future in a later commit.

Leptopoda: This is going to stay (also thinking of https://github.com/nextcloud/neon/issues/878)
I can also push the remaining PRs so you get the Idea of what I was trying to achieve.

provokateurin: Just explain it in a few sentences please, opening many dependent PRs is not great.

Leptopoda: The idea is that the request manager will do what it's name says.
Use `http.BaseRequest` and the httpclient of the account to execute the request itself.
By executing the request itself it can directly work with the `http.StreamedResponse` to efficiently convert it into the desired `DynamiteResponse` or json for the cache.
No more need for the hard to understand/debug future juggling inside the DynamiteRawResponse. 

This requires the request manager to have access  to serializer to both deserialize the response and cache.

provokateurin: Sounds good.
There is absolutely no way to avoid passing the serializer manually?

Leptopoda: We could later try to pass it together with the request in a record but that's kinda ugly.

Leptopoda: FYI this will also allow us to unify all most of the login in the different RequestManager methods.
`wrapUri` and `wrapBinary` are easily migrated to provide a `http.Request` and appropriate serializer.

provokateurin (approved):
Nicely done!

PR #1705, opened 2024-03-04
fix(neon_framework): Fix NeonUserAvatar status sizes

provokateurin:
https://github.com/nextcloud/neon/pull/1664 introduced the problem that the scaled size was selected based on the emoji which wasn't accounting for the DND status. Just removing the SizedBox fixed the problem entirely.
The emoji was also a bit big compared to the other icons, so decreased it's size a bit.


Leptopoda (commented):
I think this is the perfect scenario for a golden test :)

provokateurin:
Mh I tried adding some, but the results didn't seem very useful to me. I can add some anyway.

Leptopoda (approved):
I'll leave it up to you. The code looks good so if tests are not useful feel free to go ahead

PR #1707, opened 2024-03-04
feat(dynamite): use http.Request in generated code

Leptopoda:
The last remaining piece of #1580

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -77,7 +77,7 @@ class RequestManager {
     required Account account,
     required String cacheKey,
     required BehaviorSubject<Result<T>> subject,
-    required FutureOr<DynamiteRawResponse<B, H>> rawResponse,
+    required http.BaseRequest request,
     required DynamiteSerializer<B, H> serializer,
     required UnwrapCallback<T, DynamiteResponse<B, H>> unwrap,
     bool disableTimeout = false,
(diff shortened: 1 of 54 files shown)


provokateurin:
Shouldn't it be possible now to put the serializer into the request? Or do you think that mixes the layers too much?

Leptopoda:
Do you mean similar to the validStatuses?

I think this is to much mixing. Although I have to admit that the validStatuses shouldn't be there either

provokateurin:
Yeah, and I somewhat agree that the validation should not be in the request.

Leptopoda:
So should we move the status code matching into the serializer?
Let it throw if the encountered status code is not valid?

Leptopoda:
This might also make sense long term if we want to allow different response schemas per status code.
I don't know how this would look in detail but it's more fitting there.

provokateurin:
That sounds good and should help with supporting multiple status codes :+1:

packages/dynamite/dynamite_runtime/lib/src/client/request.dart
```diff
@@ -0,0 +1,28 @@
...
+  /// It is up to the caller to validate the response.
+  ///
+  /// This set should only be set, not modified in place.
+  Set<int>? get validStatuses => _validStatuses;
+  Set<int>? _validStatuses;
+
+  set validStatuses(Set<int>? value) {
+    _checkFinalized();
+    _validStatuses = value;
+  }
```
provokateurin, line 19: This is now part of the serializer, please remove

provokateurin (changes requested):

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -86,7 +86,10 @@ class RequestManager {
         account: account,
         cacheKey: cacheKey,
         subject: subject,
-        request: () async => rawResponse,
+        request: () async {
+          final response = await account.client.send(request);
```
provokateurin, line 90: Has to be sendWithCookies

provokateurin (changes requested):

provokateurin (approved):

PR #1713, opened 2024-03-06
refactor(dynamite_runtime)!: validate status code in the dynamite ser…

Leptopoda:
…ializer

As discussed in #1707
As we already landed breaking changes in NC I also took the time to remove the deserialization in the `DynamiteStatusCodeException`.


packages/dynamite/dynamite_runtime/lib/src/client/client.dart
```diff
@@ -100,12 +99,7 @@ class DynamiteClient with http.BaseClient {
     if (validStatuses?.contains(response.statusCode) ?? true) {
       return response;
     } else {
```
provokateurin, line 101: This should go away, right? Otherwise we have validation in two places :thinking:

Leptopoda: We still need it for the raw requests.
The last step once #1707 has landed is going to remove the entire executeRequest and executeaRawRequest methods.

Leptopoda: My bad. I just double checked and we already use the DynaMiteSerializer in the RequestManager so this could probably be removed.

We'd only need to bring back the functionality for other callers like the RequestManager.wrapUri method.

Should we just leave it for now until it's removed or should I find a solution for the wrapURI case?

provokateurin: Please remove it here and fix it in wrapUri

provokateurin (approved):

PR #1717, opened 2024-03-07
fix(neon_framework): Set theme primaryColor to fix markdown rendering

provokateurin:
The flutter_markdown package still relies on Theme.primaryColor, so we have to set it.
Previously list items were not readable in dark mode because the primaryColor was set to dark grey on a black background.


provokateurin:
We will implement a new editor soon, so this doesn't need to be fixed now (especially because the fix itself isn't really good).

PR #1748, opened 2024-03-11
refactor: Use DateTimeUtils to handle unix epochs

provokateurin:


packages/neon/neon_files/lib/src/sort/files.dart
```diff
@@ -1,11 +1,12 @@
 import 'package:neon_files/src/options.dart';
 import 'package:neon_framework/sort_box.dart';
+import 'package:nextcloud/utils.dart';
 import 'package:nextcloud/webdav.dart';
 
 final filesSortBox = SortBox<FilesSortProperty, WebDavFile>(
   properties: {
     FilesSortProperty.name: (file) => file.name.toLowerCase(),
-    FilesSortProperty.modifiedDate: (file) => file.lastModified?.millisecondsSinceEpoch ?? 0,
+    FilesSortProperty.modifiedDate: (file) => file.lastModified?.secondsSinceEpoch ?? 0,
```
Leptopoda, line 9: Is this change intended?
I'm fine with it but wanted to double check with you

Leptopoda (approved):

packages/neon/neon_files/lib/src/sort/files.dart
provokateurin: It's not really relevant because it is only used for sorting, but the last modified date only has second precision so we don't gain anything by using the milliseconds.

PR #1754, opened 2024-03-11
refactor(nextcloud): use package:timezone TZDateTime for dates

Leptopoda:

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -48,6 +48,12 @@
       "enabled": false
     },
     {
+      "matchManagers": ["pub"],
+      "matchPackageNames": ["timezone"],
+      "rangeStrategy": "bump",
+      "enabled": true
+    },
+    {
       "groupName": "built_value",
       "matchPackagePatterns": "^built"
     },
--- a/packages/neon/neon_files/lib/src/utils/task.dart
+++ b/packages/neon/neon_files/lib/src/utils/task.dart
@@ -3,6 +3,7 @@ import 'dart:typed_data';
 
 import 'package:meta/meta.dart';
 import 'package:nextcloud/nextcloud.dart';
+import 'package:timezone/timezone.dart' as tz;
 import 'package:universal_io/io.dart';
 
 sealed class FilesTask {
@@ -34,7 +35,7 @@ sealed class FilesUploadTask extends FilesTask {
 
   final int? size;
 
-  final DateTime? lastModified;
+  final tz.TZDateTime? lastModified;
 }
 
 sealed class FilesTaskIO extends FilesTask {
@@ -83,10 +84,10 @@ class FilesUploadTaskIO extends FilesTaskIO implements FilesUploadTask {
   late final FileStat _stat = file.statSync();
 
   @override
-  late int? size = _stat.size;
+  late int size = _stat.size;
 
   @override
-  late DateTime? lastModified = _stat.modified;
+  late tz.TZDateTime lastModified = tz.TZDateTime.from(_stat.modified, tz.UTC);
 
   Future<void> execute(NextcloudClient client) async {
     await client.webdav.putFile(
@@ -129,7 +130,7 @@ class FilesUploadTaskMemory extends FilesTaskMemory implements FilesUploadTask {
   final int? size;
 
   @override
-  final DateTime? lastModified;
+  final tz.TZDateTime? lastModified;
 
   Future<void> execute(NextcloudClient client) async {
     await client.webdav.putStream(
--- a/packages/neon/neon_files/pubspec.yaml
+++ b/packages/neon/neon_files/pubspec.yaml
@@ -36,6 +36,7 @@ dependencies:
   queue: ^3.0.0
   rxdart: ^0.27.0
   share_plus: ^7.0.0
+  timezone: ^0.9.2
   universal_io: ^2.0.0
 
 dev_dependencies:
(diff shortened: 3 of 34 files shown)


packages/neon/neon_files/pubspec.yaml
```diff
@@ -36,6 +36,7 @@ dependencies:
   queue: ^3.0.0
   rxdart: ^0.27.0
   share_plus: ^7.0.0
+  timezone: ^0.9.2
```
provokateurin, line 39: 0.9.0 should work as well

packages/nextcloud/lib/src/utils/date_time.dart
```diff
@@ -1,22 +1,25 @@
...
   /// ```dart
   /// final newYearsDay =
-  ///     DateTimeUtils.fromSecondsSinceEpoch(1641031200, isUtc:true);
+  ///     DateTimeUtils.fromSecondsSinceEpoch(1641031200, location:tz.UTC);
   /// print(newYearsDay); // 2022-01-01 10:00:00.000Z
   /// ```
-  static DateTime fromSecondsSinceEpoch(int secondsSinceEpoch, {bool isUtc = false}) {
-    final millisecondsSinceEpoch = secondsSinceEpoch * Duration.millisecondsPerSecond;
+  static tz.TZDateTime fromSecondsSinceEpoch(int secondsSinceEpoch, {tz.Location? location}) {
+    location ??= tz.UTC;
```
provokateurin, line 19: The package has the location as the first parameter and as required: https://github.com/srawlins/timezone/blob/dd7fef499aec4d3f851eb394fdefb9aa095ad911/lib/src/date_time.dart#L181
We should use the same API.

packages/neon_framework/test/request_manager_test.dart
```diff
@@ -575,7 +576,10 @@ void main() {
 
       test('cached ETag', () async {
         final callback = MockCallbackFunction<CacheParameters>();
-        final newExpires = DateTime.timestamp().copyWith(microsecond: 0, millisecond: 0).add(const Duration(hours: 1));
+        final newExpires = tz.TZDateTime.from(
+          DateTime.timestamp().copyWith(millisecond: 0, microsecond: 0).add(const Duration(hours: 1)),
+          tz.UTC,
+        );
```
provokateurin, line 582: Use TZDateTime and do the copyWith+add on that instead

provokateurin (changes requested):
The dependency constraints need to be fixed and all the copyWith+add in tests need to be executed on TZDateTime directly.

packages/neon/neon_files/pubspec.yaml
Leptopoda: I'd rather stay with the up to date timezone data.
I want to go with the default timezone db with no historic data and 5years worth of future data.
0.9.0 has been released in 2022-08 so we'd only have another 2.5 years worth of future data.

0.9.2 was released 20 months ago so I don't think it will restrain any reasonably up to date dependencies.

packages/neon_framework/test/request_manager_test.dart
Leptopoda: apparently there is no copyWith extension for TZDateTime so calling copy with on it always returns a normal `DateTime` object.
We can't add such a method to our own utils as it would conflict the default one (unless we shadow it on dart:core everywhere we use it).

As a next step (after fixing the http date parsing) I want to add some sort of normalization to our date time extensions where a `TZDateTime` object is returned

.github/renovate.json
```diff
@@ -46,6 +46,10 @@
       ],
       "matchCurrentVersion": "/^0/",
       "enabled": false
+    },    {
+      "matchManagers": ["pub"],
+      "matchPackageNames": ["timezone"],
+      "rangeStrategy": "bump"
```
provokateurin, line 52:
```suggestion
      "rangeStrategy": "bump",
      "enabled": true
```
I ran renovate locally to confirm this rule works and you need to add this line to override the rules above that disable updating for certain dependencies. The `rangeStrategy` isn't even necessary, but you can keep it so it is clearer what is supposed to happen.

provokateurin (changes requested):

packages/neon/neon_files/lib/src/utils/task.dart
```diff
@@ -83,10 +84,11 @@ class FilesUploadTaskIO extends FilesTaskIO implements FilesUploadTask {
   late final FileStat _stat = file.statSync();
 
   @override
-  late int? size = _stat.size;
+  late int size = _stat.size;
 
+  // tz.local should be enough but to be sure we'll add the offset manually
   @override
-  late DateTime? lastModified = _stat.modified;
+  late tz.TZDateTime lastModified = tz.TZDateTime.from(_stat.modified, tz.UTC).add(_stat.modified.timeZoneOffset);
```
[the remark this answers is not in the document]
provokateurin: The comment above the line is wrong now.

.github/renovate.json
provokateurin: The enabled key is still missing from the rule.

provokateurin (changes requested):
I'll quickly check if the renovate rule stills works if it is at the top

provokateurin:
Yeah I is broken as I expected. You need to put it as the last rule that changes the enabled flag (so exactly where it was before).

Leptopoda:
So like this?

provokateurin (approved):
Yes

PR #1763, opened 2024-03-13
feat(nextcloud): add custom http date parser to support rfc822 time zone offsets

Leptopoda:

Fixes the issue of parsing an RFC822 date with a timezone offset.
The code is pulled in from the upstream http_parser package and adjusted to also handle our use case.
I disabled failing lint rules instead of fixing them to keep the code similar to upstream.

As a bonus I also converted the code to directly use a tz.TZDateTime so we can avoid parsing it later on.

### DIFF
--- a/packages/nextcloud/test/http_date_test.dart
+++ b/packages/nextcloud/test/http_date_test.dart
@@ -1,0 +1,387 @@
+// Copyright (c) 2014, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: require_trailing_commas
+
+import 'package:nextcloud/src/utils/http_date_parser.dart';
+import 'package:test/test.dart';
+
+void main() {
+  group('format', () {
+    test('many values with 9', () {
+      final date = DateTime.utc(2014, 9, 9, 9, 9, 9);
+      final formatted = formatHttpDate(date);
+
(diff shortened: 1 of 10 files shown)


packages/nextcloud/lib/src/utils/http_parser.dart
provokateurin, line 1: How easy is it to update this if there is a change upstream?

packages/nextcloud/test/http_date_test.dart
provokateurin, line 1: Are there some more real world examples from the RFCs we could test against? The tests look good, but I'd like to make sure it won't break again.

provokateurin (changes requested):
PR title is wrong

packages/nextcloud/lib/src/utils/http_parser.dart
Leptopoda: I only added the `_parseLocation` method while the rest of the code is untouched.
Updating it should be trivial.

packages/nextcloud/test/http_date_test.dart
Leptopoda: I copied the tests from upstream and added a group for the timezone offset case.
I did not find any example date in RFC822 but there might be some scattered throughout the other RFCs.

TBH I don't feel like going through all of them just to find a different test case.

packages/nextcloud/lib/src/utils/http_parser.dart
provokateurin, line 1: The file should probably be called `http_date_parser.dart` since it only does that part.

provokateurin (approved):

PR #1767, opened 2024-03-13
fix(neon_framework): Supress subject closed errors in RequestManager

provokateurin:
This error is triggered when a bloc or widget is disposed while the request is still running. Usually only happens in the NeonImage widgets.

Best reviewed with https://github.com/nextcloud/neon/pull/1767/files?w=1

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -214,6 +214,9 @@ class RequestManager {
   }) async {
     final key = '${account.id}-$cacheKey';
 
+    if (subject.isClosed) {
+      return;
+    }
     if (!subject.hasValue) {
       subject.add(Result.loading());
     }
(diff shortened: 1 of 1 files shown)


Leptopoda:
Why did you go the route of catching the error instead of checking whether the stream/subject is closed?

provokateurin:
I guess that is possible as well, but makes the code a lot less readable and we would have to add a return statement everywhere so that the code that comes after it doesn't continue to run.

Leptopoda:
I'm really not a fan of catching errors or using try/catch for control flow.
How ugly do you expect it to get? I don't think the 3 or so checks we'd need to add are too bad.

provokateurin:
OK I will try your suggestion. We just have to be careful later to always check the subject and properly return

provokateurin:
IMO this is a lot less readable, but if you prefer this I'm ok :woman_shrugging:

Leptopoda (approved):
Yeah it really is but it's also programmatically more correct.
I'd say let's go with this and maybe with the coming changes to the request manager we can reduce some of the number of async suspensions.

Issue #910, opened 2023-10-03
Setup CI coverage service


provokateurin:
Something like Codecov should work, so we track our coverage and its changes.

Leptopoda:
For documentation:
Codecov suggests flags and carryforward-flags for use inside a monorepo.
https://docs.codecov.com/docs/flags
https://docs.codecov.com/docs/carryforward-flags

PR #1774, opened 2024-03-14
feat(tool): setup codecov flags configuration

Leptopoda:
finally closes: #910 
Sadly we need to add every package to the codecov config.
I've tested it in my fork and it works (coverage is correctly displayed in the codecov panel.
Coverage will only be added over time when each package first runs its tests.

@provokateurin did you already add the CI token to the github secrets? I don't have access to that so I couldn't check myself.

### DIFF
--- a/tool/update-codecov-config.sh
+++ b/tool/update-codecov-config.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -euxo pipefail
+cd "$(dirname "$0")/.."
+
+yq -iy "del(.flags)" .github/codecov.yaml
+
+for package in $(melos list --dir-exists=test --json | jq -c '.[]'); do
+    name=$(echo $package | jq -r '.name')
+    location=$(echo $package | jq -r '.location')
+    location=${location##"$(pwd)/"}
+    echo "$name $location"
+
+    yq -iy ".flags.$name.paths[0] = \"$location\"" .github/codecov.yaml
+done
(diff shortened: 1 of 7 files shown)


provokateurin:
I didn't add the token yet

tool/update-codecov-config.sh
provokateurin, line 1: It would be better to always generate the full config every time. Right now if we would remove a package it wouldn't be removed from the config. Also you only update the path and never create the parent elements, so I think it will fail if we have a new package.

Leptopoda: It wouldn't fail.
I just added a call to jq dropping the flags before regenerating

tool/update-codecov-config.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+#set -euxo pipefail
```
provokateurin, line 2:
```suggestion
set -euxo pipefail
```

tool/update-codecov-config.sh
provokateurin, line 1: The workflow of adding a new package needs to be adjusted so people know they need to run this script.

provokateurin (changes requested):
Is it possible to trigger all tests in this PR to see that it is working?

provokateurin:
I just added the token to the secrets so it should work.

provokateurin (changes requested):
Ok looks like this works correctly. Please add the documentation and revert the test commit, then this can be merged

provokateurin (approved):

provokateurin:
@Leptopoda Merge?

PR #1784, opened 2024-03-16
feat(neon_talk): Show user status on room page

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1778

No extra check for the room type is necessary because the status will just be null for rooms that are not 1:1s.

### DIFF
--- a/packages/neon/neon_talk/lib/src/pages/room.dart
+++ b/packages/neon/neon_talk/lib/src/pages/room.dart
@@ -57,13 +57,29 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
     return ResultBuilder.behaviorSubject(
       subject: bloc.room,
       builder: (context, result) {
+        Widget title = Text(result.requireData.displayName);
+
+        final statusMessage = result.requireData.statusMessage;
+        if (statusMessage != null) {
+          title = Column(
+            crossAxisAlignment: CrossAxisAlignment.start,
+            children: [
+              title,
+              Text(
+                statusMessage,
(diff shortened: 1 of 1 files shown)


packages/neon/neon_talk/lib/src/pages/room.dart
```diff
@@ -123,4 +129,16 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
...
+  Widget buildUserStatus(spreed.Room room) {
+    final statusMessage = room.statusMessage;
+    if (statusMessage != null) {
+      return Text(
+        statusMessage,
+        style: Theme.of(context).textTheme.labelSmall,
+      );
+    }
+
+    return const SizedBox();
```
Leptopoda, line 142: Not a fan of returning the sized box

Leptopoda (commented):
How can I test this?
I was not able to find a room status setting in talk web.

provokateurin:
You need to create 1:1 with another user and then set the user status of the other user.

packages/neon/neon_talk/lib/src/pages/room.dart
provokateurin: I guess the whole thing could be inlined into the column or the whole column moved into a separate function :woman_shrugging:

Leptopoda: I'm open to both but please change this.

Leptopoda:
Design looks good :)

Leptopoda (approved):

PR #1786, opened 2024-03-17
feat(neon_framework): store cache expration date in seconds since epoch

Leptopoda:


packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -65,14 +66,14 @@ final class DefaultRequestCache implements RequestCache {
...
       onCreate: onCreate,
       onUpgrade: (db, oldVersion, newVersion) async {
         // We can safely drop the table as it only contains cached data.
         // Non breaking migrations should not drop the cache. The next
         // breaking change should remove all non breaking migrations before it.
         await db.transaction((txn) async {
-          if (oldVersion <= 2) {
+          if (oldVersion <= 3) {
             await txn.execute('DROP TABLE cache');
             await onCreate(txn);
```
provokateurin, line 78: Technically a DELETE would be enough since the schema doesn't change

packages/neon_framework/test/persistence_test.dart
```diff
@@ -48,16 +49,16 @@ void main() {
...
-        expires: tz.TZDateTime.fromMillisecondsSinceEpoch(
-          tz.UTC,
-          tz.TZDateTime.now(tz.UTC).millisecondsSinceEpoch,
-        ),
+        expires: now,
       );
       await cache.updateParameters(account, 'key', parameters);
       result = await cache.getParameters(account, 'key');
-      expect(result, equals(parameters));
+      result as CacheParameters;
```
provokateurin, line 59:
```suggestion
```

provokateurin (changes requested):

packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -65,15 +66,15 @@ final class DefaultRequestCache implements RequestCache {
...
-      version: 3,
+      version: 4,
       onCreate: onCreate,
       onUpgrade: (db, oldVersion, newVersion) async {
         // We can safely drop the table as it only contains cached data.
         // Non breaking migrations should not drop the cache. The next
         // breaking change should remove all non breaking migrations before it.
         await db.transaction((txn) async {
-          if (oldVersion <= 2) {
-            await txn.execute('DROP TABLE cache');
```
provokateurin, line 76: The old migration needs to stay because coming from an older version should still migrate to the new schema.
Only if we absolutely need to drop the table we can remove all older migrations (we also discussed this last weekend).

provokateurin (changes requested):

packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -76,6 +77,10 @@ final class DefaultRequestCache implements RequestCache {
             await txn.execute('DROP TABLE cache');
             await onCreate(txn);
           }
+
+          if (oldVersion == 3) {
```
provokateurin, line 81:
```suggestion
          if (oldVersion <= 3) {
```
I know that if the old version was 2 or less then the DROP will be executed so the DELETE isn't necessary, but I think this is easier to read especially with future migrations also having to use <= and not ==

provokateurin (changes requested):

Leptopoda:
Yes I hope to have a fix soon

PR #1787, opened 2024-03-17
feat(neon_framework): request cache return CacheParameters on get

Leptopoda:
No matter what the cache parameters are we will always also query the cached data.
Doing both in one go saves us a round trip to the DB and imo also makes the request manager a bit easier to read.

This is the result of my effort to validate the `etag` and `expires` field in DB which turned out to not be viable.

### DIFF
--- a/packages/neon_framework/lib/src/storage/request_cache.dart
+++ b/packages/neon_framework/lib/src/storage/request_cache.dart
@@ -73,7 +73,7 @@ final class DefaultRequestCache implements RequestCache {
         // Non breaking migrations should not drop the cache. The next
         // breaking change should remove all non breaking migrations before it.
         await db.transaction((txn) async {
-          if (oldVersion <= 2) {
+          if (oldVersion <= 3) {
             await txn.execute('DROP TABLE cache');
             await onCreate(txn);
           }
--- a/packages/neon_framework/lib/src/storage/request_cache.dart
+++ b/packages/neon_framework/lib/src/storage/request_cache.dart
@@ -15,19 +15,13 @@ final _log = Logger('RequestCache');
 
 /// A storage used to cache a key value pair.
 abstract interface class RequestCache {
-  /// Get's the cached value for the given [key].
-  ///
-  /// Use [getParameters] if you only need to check whether the cache is still
-  /// valid.
-  Future<String?> get(Account account, String key);
+  /// Get's the cached value and parameters for the given [key].
+  Future<({String value, CacheParameters? parameters})?> get(Account account, String key);
 
   /// Set's the cached [value] at the given [key].
   ///
   /// If a value is already present it will be updated with the new one.
   Future<void> set(Account account, String key, String value, CacheParameters? parameters);
-
-  /// Retrieves the cache parameters for the given [key].
-  Future<CacheParameters> getParameters(Account account, String key);
 
   /// Updates the cache [parameters] for a given [key] without modifying the `value`.
   Future<void> updateParameters(Account account, String key, CacheParameters? parameters);
@@ -112,15 +106,15 @@ CREATE TABLE "cache" (
   }
 
   @override
-  Future<String?> get(Account account, String key) async {
+  Future<({String value, CacheParameters? parameters})?> get(Account account, String key) async {
     List<Map<String, Object?>>? result;
     try {
       result = await _requireDatabase.rawQuery(
         '''
-SELECT value
+SELECT value, etag, expires
 FROM cache
 WHERE account = ? AND key = ?
-      ''',
+''',
         [account.id, key],
       );
     } on DatabaseException catch (error, stackTrace) {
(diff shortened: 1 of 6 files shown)


provokateurin:
Wrong PR title

Leptopoda:
I hate github :(

provokateurin (commented):
Previously you were against this. We might not use the cached data if it has expired already and this is still the case with this PR if I see it correctly. I'm not sure which way is the best, it depends on how often we encounter expired data and how big the data is.

Leptopoda:
Thanks for the reminder.
The current code always fetches the cached data from cache even if the parameters indicate that it is already expired.
The value is then displayed while the request for the new value is loading:
https://github.com/nextcloud/neon/blob/3feb73fa061a826dd38cdb2fbcde65a0623815a7/packages/neon_framework/lib/src/utils/request_manager.dart#L274-L277

provokateurin:
I guess that makes sense, even if the data is not up-to-date we should show it so the user doesn't see nothing while loading the new data.

packages/neon_framework/lib/src/utils/request_manager.dart
provokateurin, line 1: Please add some comments to this code to explain the overall idea of how it works. I found it very difficult to review this diff and it will certainly help us in the future in case we need to refactor it again.

packages/neon_framework/lib/src/storage/request_cache.dart
provokateurin, line 1: The getParameters method should not be used anymore and can be removed, right?

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -219,26 +219,46 @@ class RequestManager {
...
-        CacheParameters? cacheParameters;
+    if (cached != null) {
+      // If the cached data is not expired unwrap and emit it.
+      // Return as the value is up to date.
+      if (cached.parameters case CacheParameters(isExpired: false)) {
+        final unwrapped = unwrap(deserialize(cached.value));
+        subject.add(Result(unwrapped, null, isLoading: false, isCached: true));
+        return;
+
+        // If the cached data is expired and has an etag try to refresh the cache parameters.
```
provokateurin, line 236:
```suggestion
        // If the cached data expired and has an etag, try to refresh the cache parameters.
```

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -219,26 +219,46 @@ class RequestManager {
...
+        // If the etag did not change unwrap and emit the cached data.
```
provokateurin, line 237:
```suggestion
        // If the etag did not change, unwrap and emit the cached data.
```

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -251,38 +271,22 @@ class RequestManager {
...
-        }
       }
-    }
 
-    final cachedValue = await _cache?.get(account, cacheKey);
-    if (subject.isClosed) {
-      return;
-    }
-    if (cachedValue != null) {
+      // If the cached data did not satisfy the above `CacheParameters` checks emit it in a loading state.
```
provokateurin, line 276:
```suggestion
      // If the cached data did not satisfy the above `CacheParameters` checks, emit it in a loading state.
```

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -251,38 +271,22 @@ class RequestManager {
...
       subject.add(
         subject.value.copyWith(
-          data: unwrap(deserialize(cachedValue)),
+          data: unwrapped,
           isLoading: true,
           isCached: true,
         ),
       );
+
+      // When there was no value in the re emit the current value as loading.
```
provokateurin, line 288: In the what?

provokateurin (changes requested):

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -219,26 +219,46 @@ class RequestManager {
...
+        // If the etag did not change, unwrap and emit the cached data.
+        // Save the new cache parameters and return.
+      } else if (cached.parameters case CacheParameters(:final etag) when etag != null && getCacheParameters != null) {
         try {
-          cacheParameters = await timeout(
-            () async => getCacheParameters?.call(),
+          final newParameters = await timeout(
+            getCacheParameters.call,
             timeLimit: timeLimit,
           );
```
provokateurin, line 244: Shouldn't the cached data be emitted here before trying to get the new parameters? In case the new etag is the same as before we just need to return and no emit the cached data a second time.

provokateurin (commented):
Code looks good, just one logic question.

packages/neon_framework/lib/src/utils/request_manager.dart
Leptopoda: In what state should the cached data be emitted?
to be correct we would need to emit it as loading and if the etag still matches just re emit as done and return.

If you like this change I'll do it in a followup as this PR doesn't do any behavior changes to the current approach (and as mentioned by you is already hard to review on its own).

provokateurin: Yeah I think this makes more sense than the current logic. Follow up is fine by me :+1:

provokateurin (approved):

PR #1788, opened 2024-03-17
refactor(neon_talk): Use room of RoomBloc on RoomPage

provokateurin:
Instead of using the room that is passed to the widget the room of the bloc should be used because it will be updated on refresh. In case the room avatar would change it will be correctly reflected now while previously the room page had to be closed, the room list reloaded and the room page opened again.

### DIFF
--- a/packages/neon/neon_talk/lib/src/pages/room.dart
+++ b/packages/neon/neon_talk/lib/src/pages/room.dart
@@ -57,9 +57,11 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
     return ResultBuilder.behaviorSubject(
       subject: bloc.room,
       builder: (context, result) {
-        Widget title = Text(result.requireData.displayName);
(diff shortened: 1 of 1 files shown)


Leptopoda (approved):
Can we test this in a meaningful way?

provokateurin:
If we add tests to the page surely. Ideally we shouldn't only test individual widgets but whole pages.

Leptopoda (changes requested):
In that case please do so

provokateurin:
I'm not sure how to actually test it because I'd want to mock the bloc, but it gets created inside the page state so it can be disposed correctly. Either the bloc creation is moved outside the widget or we make the factory mockable somehow.

provokateurin:
@Leptopoda can we just merge this? I don't think I will have the time to fix this soon.

Leptopoda (approved):

PR #1792, opened 2024-03-19
chore(neon_talk): Ignore irrelevant coverage parts in room avatar

provokateurin:
Testing the other cases has no point, so we don't do it. The added ignores bump the coverage to 100%


Leptopoda:
why should we ignore it?
I know that this is currently what's missing for the 100% but it's just a number and ignoring it would make it harder to find uncovered places

provokateurin:
In my opinion this makes it easier to spot actual problems. There is no point in testing these other cases, so we should just ignore them. Now if there is an actual problem in the code (e.g. the case we do test) then we will spot it right away.

Leptopoda (approved):

PR #1805, opened 2024-03-22
fix(neon_framework): Initialize timezone database

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1800

This needs to go through the platform because web is different.


packages/neon_framework/lib/src/platform/web.dart
```diff
@@ -39,8 +40,10 @@ class WebNeonPlatform implements NeonPlatform {
   bool get canUsePaths => false;
 
   @override
-  FutureOr<void> init() {
+  FutureOr<void> init() async {
```
Leptopoda, line 43:
```suggestion
  Future<void> init() async {
```

There is no way this will complete synchronously

Leptopoda (approved):

PR #1819, opened 2024-03-27
fix(neon_framework): Catch FormatExceptions when parsing Http-dates in RequestManager

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1762

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -395,6 +395,8 @@ class CacheParameters {
   });
 
   /// Parse the cache parameters from HTTP response headers.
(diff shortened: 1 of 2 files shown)


Leptopoda (changes requested):
https://github.com/nextcloud/neon/blob/90106577d5db810ba26ebd634399db9a367beb66/packages/nextcloud/lib/src/webdav/file.dart#L68-L69

Also needs the treatment

provokateurin:
No I think in that case we shouldn't catch it. The users have to handle it themselves. If we just return null then one might think the file simply has no last modified date or the prop wasn't requested.

Leptopoda (approved):

PR #1820, opened 2024-03-27
feat(neon_framework): emit cached data before fetching new cache parameters

Leptopoda:

As discussed in: https://github.com/nextcloud/neon/pull/1787#discussion_r1540738969
This will result in faster loading of the cached data with an extra ui refresh when the loading is done.


provokateurin (approved):
Just needs a proper rebase

PR #1823, opened 2024-03-27
Improve WebDAV props generation

provokateurin:
Towards https://github.com/nextcloud/neon/issues/274

I strongly recommend to review commit-by-commit to see which changes actually belong together.


Leptopoda (approved):
LGTM
Can you please explain the reasoning behind the last commit to me?

provokateurin:
I did it because in the REST APIs the fileid is usually an integer as well, so it is easier to use it. Also if it is always an integer it should be typed as such.

PR #1826, opened 2024-03-27
feat(neon_talk): Display icon next to deleted messages

provokateurin:
Makes it a lot clearer that it is not a normal chat message.


packages/neon/neon_talk/lib/src/widgets/message.dart
```diff
@@ -303,6 +304,20 @@ class TalkCommentMessage extends StatelessWidget {
             : textTheme.bodyMedium,
       ),
     );
+    if (chatMessage.messageType == spreed.MessageType.commentDeleted) {
+      text = Row(
+        children: [
+          Icon(
+            AdaptiveIcons.cancel,
+            size: 12,
```
Leptopoda, line 312:
```suggestion
            size: textTheme.bodySmall?.fontSize,
```

Leptopoda (changes requested):

Leptopoda (approved):

PR #1829, opened 2024-03-28
Feat/nextcloud/webdav_requests

Leptopoda:

### DIFF
--- a/packages/nextcloud/lib/src/webdav/client.dart
+++ b/packages/nextcloud/lib/src/webdav/client.dart
@@ -12,7 +12,7 @@ import 'package:nextcloud/src/webdav/props.dart';
 import 'package:nextcloud/src/webdav/utils.dart';
 import 'package:nextcloud/src/webdav/webdav.dart';
 import 'package:nextcloud/utils.dart';
-import 'package:universal_io/io.dart' hide HttpClient;
+import 'package:universal_io/io.dart' show File, FileStat;
 
 // ignore: do_not_use_environment
 const bool _kIsWeb = bool.fromEnvironment('dart.library.js_util');
@@ -30,15 +30,6 @@ class WebDavClient {
   Future<http.StreamedResponse> _send(
     http.BaseRequest request,
   ) async {
-    final authentication = rootClient.authentications?.firstOrNull;
-    if (authentication != null) {
-      request.headers.addAll(
-        authentication.headers,
-      );
-    }
-
-    request.headers[HttpHeaders.contentTypeHeader] = 'application/xml';
-
     // On web we need to send a CSRF token because we also send the cookies.  In theory this should not be required as
     // long as we send the OCS-APIRequest header, but the server has a bug that only triggers when you also send the
     // cookies. On non-web platforms we don't send the cookies so we are fine, but on web the browser always does it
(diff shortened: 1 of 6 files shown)


packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -62,27 +28,17 @@ class WebDavClient {
...
+    http.BaseRequest request,
+  ) async {
     final authentication = rootClient.authentications?.firstOrNull;
     if (authentication != null) {
       request.headers.addAll(
         authentication.headers,
       );
     }
 
+    request.headers[HttpHeaders.contentTypeHeader] = 'application/xml';
```
provokateurin, line 40: Shouldn't the header be set in the requests directly? Otherwise we have to set it again in the RequestManager.

Leptopoda: Yes; in an ideal world the contentTypeHeader and authentication headers would be set for every request and if you want I can also do it in this PR.

Sadly this will not help us in any way with the RequestManager. The CSRF bug still requires some async computation before the request is finalized. 
I'm currently planning to allow the request manager `wrap` method to take a HttpClient as a parameter. This way the `wrapWebDav` method can provide a client that adds the CSRF workaround (probably the `WebDavClient` once it also extends HttpBaseClient). All other wrap implementations would provide the base nextcloud client form the account as it is already done.

This also allows us to further decouple the RequestManager from the neon_framework by reducing the reliance on the `Account` model allowing us to separate it into another package when needed.

Leptopoda:
The codecov report concerns me a bit.
I would expect it to already show the flags ...

provokateurin:
Yeah I also tried looking at the report but it showed nothing :/

packages/nextcloud/lib/src/webdav/client.dart
Leptopoda: Should I change the headers now or just leave them in the client as it will be needed anyways to set the CSRF token?

provokateurin: Since we will no longer need the CSRF hack at some point the whole async thing will go away, so I'd prefer if the headers are already set in the request because we will not need the custom client anymore.

provokateurin:
Wouldn't it make sense to have one common method that applies the content-type and authorization headers instead of copying the code into every method? At least that was what I envisioned (same for the upload headers method which you removed now).

Leptopoda:
Yeah I also thought about that. I went with this approach as it would reduce the number of maps created as the values are added to the request directly.
Otherwise we would:
- create a map for auth headers
- copy it into a new common headers map (content-type + auth)
- create a new map for the upload headers
- copy both resulting maps into the request.headers map

We could write a method that takes a request as a parameter and alter it but I'm not sure if this is better.
I'm open for suggestions and ideas :)

provokateurin:
> We could write a method that takes a request as a parameter and alter it

That was my idea, then there are no extra copies.

It is not the cleanest way to do this, but it eliminates the duplicate code effectively while still keeping it readable and performant, so it is good enough IMO.

Leptopoda:
I'm not that sure.
This way we rely on side-effects and I think that the code is less readable.
While it does duplicate some code it is immediately visible what headers are applied and under what condition :thinking:

Leptopoda:
@provokateurin ping
I guess you still prefer the method approach?

provokateurin:
Yeah I think it will just make it easier to work on the code in the future, but you can also keep it this way since it is unlikely that the code will really change a lot again :woman_shrugging: We just need to really make sure that all methods are updated if something changes (most likely the upload headers if ever).

provokateurin (approved):

PR #1833, opened 2024-03-28
feat(neon_lints): Add warnings to avoid dart:html and dart:io

provokateurin:


Leptopoda (changes requested):
Please also test the new rules:
https://github.com/nextcloud/neon/blob/main/packages/neon_framework/test/neon_lints.dart

Leptopoda (approved):

Issue #1764, opened 2024-03-13
More custom_lint rules


Leptopoda:
- [ ] fix custom_lint in IDEs
- [ ] disallowing the use of `@internal` and `export` in neon_* packages as they should not have a public API other than the AppImplementation
- [ ] Enforce tz.TZDateTime to avoid wrong timezones

PR #1836, opened 2024-03-29
feat(neon_lints): add avoid_exports lint rule

Leptopoda:
Towards: #1764


provokateurin (changes requested):
Please add a test and also check if the meta dependency can be removed from the packages.

Leptopoda:
How should I test this?
The lint rule is disabled in the framework package.

provokateurin:
Would it work to add an example subpackage to neon_lints and move all the tests to it?

provokateurin (approved):

PR #1849, opened 2024-04-02
fix(neon_talk): Improve message list constraints

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1740

This not only adds the padding so that the scroll bar never overlaps the message time, but also moves the constraint into the list itself so that the scrollbar is always at the most right position regardless of the list width and you can scroll everywhere.

I didn't find any way to use SliverConstrainedCrossAxis because it aligns everything to the left but the messages should be centered. Maybe @Leptopoda you know what would make this work.

### DIFF
--- a/packages/neon/neon_talk/lib/neon_talk.dart
+++ b/packages/neon/neon_talk/lib/neon_talk.dart
@@ -13,6 +13,7 @@ import 'package:neon_talk/src/blocs/talk.dart';
 import 'package:neon_talk/src/options.dart';
 import 'package:neon_talk/src/pages/main.dart';
 import 'package:neon_talk/src/routes.dart';
+import 'package:neon_talk/src/theme.dart';
 import 'package:nextcloud/core.dart' as core;
 import 'package:nextcloud/nextcloud.dart';
 import 'package:nextcloud/spreed.dart';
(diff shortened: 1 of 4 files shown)


Leptopoda:
> I didn't find any way to use SliverConstrainedCrossAxis because it aligns everything to the left but the messages should be centered. Maybe @Leptopoda you know what would make this work.

I don't know of any way to make it work with a SliverConstrain
This boils down to the question of what you actually what to constrain. The chat view or the messages themselves?
If we ever want to implement a chat bubble UI (like Signal/WhatsApp use) you do not actually want to constrain the chat list as the bubbles should be alligned to the edges of the list/screen.
In this case I'd rather just put the old `Center/ConstrainedBox` inside the `TalkMessage` widget.

provokateurin:
Only the messages should be constrained. Right now I just constrained the whole sliver, but I guess constraining the individual messages is also possible. Then the padding can be applied to the whole list and it's separate from the width constraint.

provokateurin:
I couldn't find any other way to make it work, using ConstrainedBox on the TalkMesssage widget has no effect (but I don't understand why).

Leptopoda:
[talk_constraints.patch.txt](https://github.com/nextcloud/neon/files/14970244/talk_constraints.patch.txt)

This ontop of this branch works for me. Or are you trying to achieve something different?

provokateurin:
I almost had the same patch as you sent, but it didn't work. Now it did :woman_shrugging:

Leptopoda (approved):

PR #1850, opened 2024-04-02
feat(nextcloud)!: deserialize webdav date fields

Leptopoda:


packages/nextcloud/generate_props.dart
```diff
@@ -36,18 +36,18 @@ void main() {
       'lock-owner-displayname': 'String',
       'lock-owner-editor': 'String',
       'lock-owner-type': 'int',
-      'lock-time': 'int',
+      'lock-time': 'unixEpoch',
       'lock-timeout': 'int',
```
provokateurin, line 40: This could be parse as a Duration

provokateurin (approved):
Cool stuff!

packages/nextcloud/generate_props.dart
Leptopoda: What do you mean?

It must be a DateTime according to the upstream docs: https://github.com/nextcloud/documentation/blob/master/developer_manual/client_apis/WebDAV/basic.rst#supported-properties
> Date when the lock was created as a timestamp.

provokateurin: No the timeout which is the TTL in seconds. The code preview shows too much context.

provokateurin (approved):

provokateurin (changes requested):
Wait you also changed the int/bool fields?

Leptopoda:
yep. Why?
I only changed it where the docs said it's either `1` or `0`

provokateurin:
But will it serialize correctly? Because the server might not accept a `true` instead of a `1` (I didn't test that though).

Leptopoda:
You're right.
I also only checked the code for deserialization.

provokateurin:
According to https://www.w3.org/TR/xmlschema-2/#boolean it should work because all those representations are possible. If this turns out to be true then the server docs should be changed to clearly state that the values are not integers but actually booleans.

provokateurin (changes requested):
oc:favorite should also be bool.
At least the favorites prop handles all representations correctly: https://github.com/nextcloud/server/blob/c0d21b44437d6ff0ef4ffa08096f18c1e4870fcc/apps/dav/lib/Connector/Sabre/TagsPlugin.php#L285
Other props might not implement it correctly but that needs investigation (let me know if I should help with that as I already have experience digging into the server DAV code).

provokateurin:
Ok I had a look at all the boolean fields except for oc:favorite they are all dead props meaning the client can't update them. oc:favorite has the correct handling, so everything looks good!

provokateurin:
I already had the changes for oc:favorite for local testing, so I just pushed them. Please squash and rebase and then we are good to go.

Leptopoda:
Thanks for taking a look

PR #1856, opened 2024-04-03
fix(dynamite_runtime): cookie header format

Leptopoda:
this does not fix our current 401 errors with webdav

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/client/client.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/client/client.dart
@@ -62,7 +62,22 @@ class DynamiteClient with http.BaseClient {
     if (cookieJar != null) {
       final cookies = await cookieJar!.loadForRequest(request.url);
       if (cookies.isNotEmpty) {
-        request.headers['cookie'] = cookies.join('; ');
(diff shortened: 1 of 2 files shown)


provokateurin:
Why was it wrong before? What was broken?

Leptopoda:
`Cookie.toString` as used before will generate a string as is used in the set-cookie header.
This means that all parameters (like HttpOnly, MaxAge ... ) where also wronly appended.

The change test case checked against `'a=b; HttpOnly'`

provokateurin:
Makes sense. Is it really necessary to use the raw charcodes?

Leptopoda:
probably not :)
I was just reading to much `rfc6265`.

provokateurin:
Then I'd prefer to not use them as it makes it easier to read and understand later.

packages/dynamite/dynamite_runtime/lib/src/client/client.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: constant_identifier_names
```
provokateurin, line 1: This ignore can be removed now

provokateurin (approved):

PR #1857, opened 2024-04-03
fix(nextcloud): Send cookies when acquiring CSRF token for WebDAV requests

provokateurin:
We need to send the cookies when getting the token because otherwise the token will not belong to the session we use to send the actual request.


Leptopoda:
I think it might make more sense to just delete the `sendWithCookies` method and always send the cookie in the send method.

AFAICT there is no need for the separation anymore.

provokateurin:
True, I'll make a new PR

PR #1858, opened 2024-04-03
Fix tests and WebDAV requests

provokateurin:
Many problems at once that need to be solved at the same time:
1. Green CI even though tests fail: https://github.com/nextcloud/neon/actions/runs/8541770039/job/23401897660?pr=1857
2. All requests should always send cookies (including the request that gets the CSRF token https://github.com/nextcloud/neon/pull/1857)
3. Fixtures were not updated in https://github.com/nextcloud/neon/pull/1853 but it wasn't noticed because the CI was green.


provokateurin:
It looks like this also fixes the coverage problems for dart-only packages.

melos.yaml
```diff
@@ -38,18 +38,6 @@ scripts:
   analyze: >
     dart analyze --fatal-infos . &&
     dart run custom_lint --fatal-infos .
-  test: >
-    melos run test:dart &&
-    melos run test:flutter
-  test:dart: >
```
Leptopoda, line 44: Please leave the separate flutter and dart test commands.

I find them very useful as most of the time I work on neon and and don't need to run the time consuming nextcloud tests locally

Leptopoda (changes requested):

Leptopoda (approved):

PR #1859, opened 2024-04-03
refactor(dynamite_runtime)!: Remove executeRequest and executeRawRequest methods

provokateurin:
I noticed these tests are no longer used by any generated code and can be replaced by just using the send method.

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -181,8 +181,7 @@ class RequestManager {
       account: account,
       cacheKey: uri.toString(),
       getCacheParameters: () async {
-        final response = await account.client.executeRawRequest(
-          'HEAD',
+        final response = await account.client.head(
           uri,
           headers: headers,
         );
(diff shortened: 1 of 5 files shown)


packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -181,11 +181,11 @@ class RequestManager {
...
-          headers: headers,
-        );
+        final request = http.Request('HEAD', uri);
+        if (headers != null) {
+          request.headers.addAll(headers);
+        }
+        final response = await account.client.send(request);
 
         return CacheParameters.parseHeaders(response.headers);
       },
```
Leptopoda, line 191:
```suggestion
        final response = await account.client.head(
          uri,
          headers: headers,
        );

        return CacheParameters.parseHeaders(response.headers);
      },
```

There is no benefit over your proposed changes but I find this more readable.

Leptopoda (commented):
This would've been my next PR anyways but thanks for the cleanup :)

Leptopoda (approved):

PR #1860, opened 2024-04-04
feat(dynamite): Remove need for double fix and format

provokateurin:
Towards https://github.com/nextcloud/neon/issues/1244


packages/dynamite/dynamite/lib/src/models/type_result/type_result.dart
```diff
@@ -107,9 +107,13 @@ sealed class TypeResult {
...
       ..write(serializerName ?? r'_$jsonSerializers')
-      ..write('.deserialize(')
+      ..write('.deserialize(\n')
       ..write(object)
-      ..write(', specifiedType: $fullType)');
+      ..write(',\n')
+      ..write('specifiedType: ')
+      ..write(fullType)
+      ..write(',\n')
+      ..write(')');
```
Leptopoda, line 116: Why not use `writeln` ?

PR #1866, opened 2024-04-04
feat(neon_framework): Only search in active app first

provokateurin:
Closes https://github.com/nextcloud/neon/issues/684

~~No tests added yet, but please already provide feedback @Leptopoda~~

### DIFF
--- a/packages/neon_framework/lib/src/widgets/unified_search_results.dart
+++ b/packages/neon_framework/lib/src/widgets/unified_search_results.dart
@@ -30,22 +30,35 @@ class NeonUnifiedSearchResults extends StatelessWidget {
   Widget build(BuildContext context) {
     final accountsBloc = NeonProvider.of<AccountsBloc>(context);
     final bloc = accountsBloc.activeUnifiedSearchBloc;
-    return ResultBuilder.behaviorSubject(
-      subject: bloc.providers,
-      builder: (context, providers) => NeonListView(
-        scrollKey: 'unified-search',
-        isLoading: providers.isLoading,
-        error: providers.error,
-        onRefresh: bloc.refresh,
-        itemCount: providers.data?.length ?? 0,
-        itemBuilder: (context, index) {
-          final provider = providers.data![index];
+    return StreamBuilder(
+      stream: bloc.isExtendedSearch,
+      builder: (context, isExtendedSearchSnapshot) => ResultBuilder.behaviorSubject(
+        subject: bloc.providers,
+        builder: (context, providers) => NeonListView(
+          scrollKey: 'unified-search',
+          isLoading: providers.isLoading,
+          error: providers.error,
+          onRefresh: bloc.refresh,
+          topFixedChildren: [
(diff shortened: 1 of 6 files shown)


packages/neon_framework/lib/src/widgets/unified_search_results.dart
```diff
@@ -30,20 +30,45 @@ class NeonUnifiedSearchResults extends StatelessWidget {
...
+            scrollKey: 'unified-search',
+            isLoading: providers.isLoading,
+            error: providers.error,
+            onRefresh: bloc.refresh,
+            itemCount: itemCount,
+            itemBuilder: (context, index) {
+              if (!isExtendedSearch && index == itemCount - 1) {
+                return Padding(
+                  padding: const EdgeInsets.only(top: 10),
+                  child: Align(
```
Leptopoda, line 55: Wht is the alignment doing?

packages/neon_framework/lib/src/widgets/unified_search_results.dart
```diff
@@ -30,20 +30,45 @@ class NeonUnifiedSearchResults extends StatelessWidget {
...
+
+          return NeonListView(
+            scrollKey: 'unified-search',
+            isLoading: providers.isLoading,
+            error: providers.error,
+            onRefresh: bloc.refresh,
+            itemCount: itemCount,
+            itemBuilder: (context, index) {
+              if (!isExtendedSearch && index == itemCount - 1) {
+                return Padding(
```
Leptopoda, line 53: To my knowledge search providers can return as many results as they want so I'd rather have the "search globally" button above the results. That way a user does not need to first scroll down before searching globally.

packages/neon_framework/lib/src/widgets/unified_search_results.dart
```diff
@@ -30,20 +30,45 @@ class NeonUnifiedSearchResults extends StatelessWidget {
...
+                    child: ElevatedButton.icon(
+                      onPressed: bloc.enableExtendedSearch,
+                      label: Text(NeonLocalizations.of(context).searchEverywhere),
+                      icon: Icon(AdaptiveIcons.search),
+                    ),
+                  ),
+                );
+              }
+
+              final provider = providers.data![index];
```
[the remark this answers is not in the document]
provokateurin: The problem is this does nothing because the type is always nullable and it would have to be `requireData!`

packages/neon_framework/lib/src/widgets/unified_search_results.dart
provokateurin: Forcing the button to not take up the full width.

packages/neon_framework/lib/src/widgets/unified_search_results.dart
provokateurin: Sounds good, having it in the scrollable content was annoying anyway (both UX and code wise).

Leptopoda (approved):

PR #1871, opened 2024-04-05
feat(neon_news): Show feeds without folder

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2


Leptopoda (commented):
the first commit must have tests and it would be great if the second one would also get some :)

provokateurin:
I'm not too interested in investing time to add tests right now. My plan is to rewrite the News app from scratch because it is the oldest part of the whole project and it really shows. There is so much horrible stuff there and I don't want to touch it.

Leptopoda (approved):

PR #1877, opened 2024-04-06
feat(tool,nextcloud)!: Generate rich object parameters

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1847

Not really useful at the moment because of the lack of discriminator support (https://github.com/nextcloud/neon/issues/1231) and a way to use exhaustive switch statements.


provokateurin:
I change the patch to only be a simple object instead of the discriminator so in theory it is usable now (and also everything works). I would like to first upstream the missing parameter type and missing fields. Additionally we have to wait until the next Talk releases which will include https://github.com/nextcloud/spreed/pull/12021 as the some parameters were incorrectly not typed as string before.

provokateurin:
Ok apparently the fix for Talk was only backport to 29. To keep supporting the broken values we'll have to change the type to `oneOf<int|string>` for now :/ Once 28 support is dropped we can remove the hack.

provokateurin:
PR for the fields that are missing in the current definitions is sent upstream: https://github.com/nextcloud/server/pull/45424

provokateurin:
Let's just get this merged and using the discriminator can be done later once dynamite properly supports it again.

packages/nextcloud/lib/src/api/spreed.openapi.json
```diff
@@ -900,97 +900,139 @@
...
-                                "type": "string"
-                            },
-                            {
-                                "type": "integer"
-                            }
-                        ]
+                        "type": "string",
+                        "description": "The file size in bytes",
+                        "example": "3145728"
                     },
```
Leptopoda, line 971: Why can't it be an integer anymore?
Is this an error or was it just false before?

Leptopoda (commented):
LGTM. Please verify the comment.

packages/nextcloud/lib/src/api/spreed.openapi.json
provokateurin: Ah yeah I forgot to patch that back in

Leptopoda (approved):

Issue #353, opened 2023-06-03
Implement Cookbook App


provokateurin:

provokateurin:
@Leptopoda I hope assigning you here is ok?

Issue #933, opened 2023-10-06
Support generating code for JSON schema files


provokateurin:
Dynamite is already able to generate JSON schemas, but not directly from a JSON schema file.
A second builder should be introduced to support this. Some code refactoring that separates the two features more is likely needed.

Leptopoda:
Having this code separate would also allow us to easily generate the api and the schema separately like:

```
core.openapi.dart
core.schema.dart
```

With the first one importing the later.

This might make larger specs both more readable and more performant to build (thinking about talk)

provokateurin:
We still can't output two files in a single builder, so that won't work

Leptopoda:
But if we have two builders we could chain them together in the build.yaml

provokateurin:
Then we would still need two separate inputs I think

Leptopoda:
I think this is worth testing before going ahead with the issue.

I think this is possible with some build.yaml magic.

provokateurin:
For the initial implementation I would just try to get this working. Afterwards we can improve our OpenAPI support by trying to separate the two steps.

Leptopoda:
I did have a look at this and the only limitation I found was our current code around reference resolving (`$ref`).
We always pass the entire openapi spec arround in our resolving code while the only use is for reference resolving.

Our current reference resolving is very limited and needs to be rewritten. I tried a few ways to fix them but the only one that worked was a bit ugly.

PR #1881, opened 2024-04-08
fix(dynamite): allow boolean values for Schema.additionalProperties

Leptopoda:
needed for #353

### DIFF
--- a/packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
+++ b/packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
@@ -107,3 +107,42 @@ class SchemaType extends EnumClass {
 
   static Serializer<SchemaType> get serializer => _$schemaTypeSerializer;
 }
+
+/// A Schema value can be either a json boolean or object.
+///
+/// https://json-schema.org/understanding-json-schema/basics
+class SchemaPlugin implements SerializerPlugin {
+  const SchemaPlugin();
+
+  @override
+  Object? afterDeserialize(Object? object, FullType specifiedType) => object;
+
+  @override
+  Object? afterSerialize(Object? object, FullType specifiedType) => object;
+
+  @override
+  Object? beforeDeserialize(Object? object, FullType specifiedType) {
+    if (specifiedType.root != Schema) {
+      return object;
+    }
+
+    switch (object) {
+      case null:
+        return null;
+
+      case bool _:
+        if (object) {
+          // An empty list in BuiltValue it equivalent to the empty json object.
+          return [];
+        } else {
+          throw UnsupportedError('The never matching schema is not yet supported.');
+        }
+
+      default:
+        return object;
+    }
+  }
+
+  @override
+  Object? beforeSerialize(Object? object, FullType specifiedType) => object;
+}
--- a/packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.dart
+++ b/packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.dart
@@ -70,6 +70,56 @@ abstract class Base implements $BaseInterface, Built<Base, BaseBuilder> {
   static Serializer<Base> get serializer => _$baseSerializer;
 }
 
+@BuiltValue(instantiable: false)
+abstract interface class $AdditionalPropertiesInterface {
+  @BuiltValueField(wireName: 'empty_schema_bool')
+  BuiltMap<String, JsonObject>? get emptySchemaBool;
+  @BuiltValueField(wireName: 'empty_schema')
+  BuiltMap<String, JsonObject>? get emptySchema;
+  BuiltMap<String, BuiltMap<String, JsonObject>>? get nested;
+  @BuiltValueField(wireName: 'Object')
+  BuiltMap<String, JsonObject>? get object;
+  @BuiltValueField(wireName: 'bool')
+  BuiltMap<String, bool>? get $bool;
+  BuiltMap<String, int>? get integer;
+  @BuiltValueField(wireName: 'double')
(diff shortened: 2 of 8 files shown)


packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
```diff
@@ -47,7 +49,36 @@ abstract class Schema implements Built<Schema, SchemaBuilder> {
...
+    return openapi.serializers.deserializeWith(
+      serializer,
+      properties.value,
+    )!;
+  }
+
+  /// Use [parsedAdditionalProperties] instead.
+  @visibleForOverriding
+  @BuiltValueField(wireName: 'additionalProperties')
+  JsonObject get additionalProperties;
```
provokateurin, line 81: Wire name is not needed. IMO it would be better to name this getter something like `rawAdditionalProperties` (and keep the wire name) and then call the `parsedAdditionalProperties` just `additionalProperties`.

packages/dynamite/dynamite/lib/src/builder/resolve_type.dart
```diff
@@ -120,7 +120,7 @@ TypeResult resolveType(
         );
       case openapi.SchemaType.object:
         if (schema.properties == null) {
-          if (schema.additionalProperties == null) {
+          if (schema.parsedAdditionalProperties == openapi.Schema()) {
             result = TypeResultBase(
               'JsonObject',
```
provokateurin, line 125: This should be a `BuiltMap<String, JsonObject>`. With additional properties we know we have a Map with Strings as keys and an unknown type as values. JsonObject could also be any other class that implements JsonObject which is not correct.

provokateurin (changes requested):

provokateurin:
The funny thing is I know that this was implemented at some point, but I don't know why it went missing :sweat_smile:

Leptopoda:
Once I fixed everything I'll make sure to have thorough dynamite_e2e tests.

I'm also looking into the posibility of a custom `Schema` builtvalue_serializer as the boolean true/false could occur everywhere we currently accept a Schema (although it doesn't make much sense in the someOf cases ...)

provokateurin:
I don't think `true` is accepted everywhere as a schema. Only `additionalProperties` is an exception iirc.

Leptopoda:
https://json-schema.org/understanding-json-schema/basics

nope. since draft6 every json-schema can be described by a true/false.
In practice this is probably pointless but thinking of #933 I'll probably add it the right way

packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
```diff
@@ -88,6 +88,11 @@ abstract class Schema implements Built<Schema, SchemaBuilder> {
         );
       }
     }
+
+/*     final additionalProperties = b.additionalProperties;
+    if (additionalProperties == null) {
+      b.additionalProperties = Schema();
+    } */
```
provokateurin, line 95: ?

provokateurin (changes requested):
Please also add an end to end test

packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.dart
```diff
@@ -70,6 +70,56 @@ abstract class Base implements $BaseInterface, Built<Base, BaseBuilder> {
   static Serializer<Base> get serializer => _$baseSerializer;
 }
 
+@BuiltValue(instantiable: false)
+abstract interface class $AdditionalPropertiesInterface {
+  @BuiltValueField(wireName: 'empty_schema_bool')
+  BuiltMap<String, JsonObject>? get emptySchemaBool;
+  @BuiltValueField(wireName: 'empty_schema')
+  BuiltMap<String, JsonObject>? get emptySchema;
+  BuiltMap<String, BuiltMap<String, JsonObject>>? get nested;
```
Leptopoda, line 79: I think that nested is the only broken one rn.
And iirc the nextcloud specs make use of that pattern

provokateurin: Yep there are some places this appears. Why is it broken though, this code looks fine to me?

provokateurin:
What we currently don't handle is an object that has both properties and additionalProperties. In theory this is also possible with specs coming from Nextcloud, but hasn't happened so far. Probably worth looking into at some point.

Leptopoda:
but this is something we'll need to implement in the serializer itself and I guess it will get ugly very fast

Leptopoda:
@provokateurin ping :)

packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.dart
provokateurin: What about this?

provokateurin:
Sorry I thought you were going to make the next move :D

packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.dart
Leptopoda: You where right :)
I could not find anything that indicated this is actually wrong.

packages/dynamite/dynamite_end_to_end_test/lib/types.openapi.json
```diff
@@ -55,6 +55,104 @@
             "Object": {
                 "type": "object"
             },
+            "AdditionalProperties": {
```
provokateurin, line 58: Please add some tests that make use of this code, especially the nested variant.

provokateurin (changes requested):

provokateurin (approved):

PR #1894, opened 2024-04-10
feat(nextcloud)!: Support server 29 and spreed 19

provokateurin:
~~Depends on https://github.com/nextcloud/server/pull/44732
This is not the final version (I hope for https://github.com/nextcloud/spreed/pull/12041 to land), but everything is working already when testing against the old versions.~~ Once the server releases are done and the docker images are available we can also test against those and get this merged.


provokateurin:
Everything merged upstream, this PR is ready :)

provokateurin:
@Leptopoda please review the maxItems fix. I think there is something fundamentally wrong about checks for types in someOfs because they don't get generated anywhere, but that was already the case before and is not in scope of this fix.

packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
```diff
@@ -0,0 +1,19 @@
+import 'package:built_collection/built_collection.dart';
+
+/// Checks the [input] length against [minItems].
+///
+/// Throws an `Exception` containing the [parameterName] if the `input` is too short.
+void checkMinItems(BuiltList<dynamic>? input, int minItems, String parameterName) {
```
Leptopoda, line 6: Why not pass the `ListBuilder`?
This would save us a build and thus a list copy that would get discarded immediately.

The builder also has a length property: https://pub.dev/documentation/built_collection/latest/built_collection/ListBuilder/length.html

packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
```diff
@@ -0,0 +1,19 @@
...
+    throw FormatException('Parameter "$input" has to be at least $minItems items long');
+  }
+}
+
+/// Checks the [input] length against [maxItems].
+///
+/// Throws an `Exception` containing the [parameterName] if the `input` is too long.
+void checkMaxItems(BuiltList<dynamic>? input, int maxItems, String parameterName) {
+  if (input != null && input.length > maxItems) {
+    throw FormatException('Parameter "$input" has to be at most $maxItems items long');
```
Leptopoda, line 17: This is also wrong on the string check but it should print the `parameterName` instead of the input.

packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
```diff
@@ -61,6 +61,10 @@ abstract class Schema implements Built<Schema, SchemaBuilder> {
 
   int? get maxLength;
 
+  int? get minItems;
+
+  int? get maxItems;
+
```
Leptopoda, line 67: This and other validators should be documented in the doc comment.
Probably something for a followup so please open an issue.

Leptopoda (changes requested):

Leptopoda:
>  I think there is something fundamentally wrong about checks for types in someOfs because they don't get generated anywhere

Can you show me where you'd expect one but they are not generated?

packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
provokateurin: The problem is that for parameters we get a BuiltList and for responses we get a ListBuilder. One of them has to be converted to the other, or we handle it special and make the checks different such that lists only pass the length property like you said (but that requires more work).

provokateurin:
For example https://github.com/nextcloud/neon/blob/007a96fa4334433ac35e00fc0d794fd015184f3b/packages/nextcloud/lib/src/api/spreed.openapi.dart#L16365C9-L16365C25 should have the check but I don't see it anywhere.

provokateurin:
Yeah, no point in spending time on fixing this right now.

packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
provokateurin: I guess it would be possible to do this without a lot of hacks by just calling `$value.length` which will work for both options.

Leptopoda: 1. do we make use of this anywhere? I can't find any reference so please add a test to e2e.
2. What about making the parameter of the `checkMinItems` and `checkMaxItems` methods dynamic and just do a typecheck? This shouldn't be as bad as it sounds because the methods are only used by generated code.

provokateurin: 1. I added e2e tests already :thinking:
2. I think passing the length to the method directly is better than doing a type check inside the method. We know that only BuiltList and ListBuilder will be passed, so we can safely call `.length` on them.

Leptopoda: I meant a test where a `BuiltList` is used.
I can only see e2e tests where a `ListBuilder` is used.

> It should be possible to do this without a lot of hacks by just calling $value.length which will work for both options (somehow I thought this would be more complex).

Yeah this might be the best thing for now. I just wanted to not have the extra complexity  in dynamite but there is probably nothing we can do for now :(

packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
provokateurin: Aren't you doing that in https://github.com/nextcloud/neon/pull/1931 already?

Leptopoda: nope

Leptopoda: I meant that the constraints of a parameter should be added to the doc comments of the generated method/attribute. i.e that the generated code has these documentations.

Not that the dynamite code has them. You are right that I did add thorough comments to them.

provokateurin: https://github.com/nextcloud/neon/issues/1944

provokateurin:
@Leptopoda can you review again?

Leptopoda:
looks good for now.
I have a lot of dynamite changes coming so I'll do a final review once it is ready to be merged

Leptopoda (approved):

Leptopoda (approved):

PR #1896, opened 2024-04-10
fix(neon_framework): Use broadcast stream in maintenance mode bloc

provokateurin:
When switching to another account and back will throw an error because the home page listens to the same stream again.


Leptopoda (commented):
please add a comment explaining that.
I can already see me wondering why we need it

Leptopoda (approved):

PR #1903, opened 2024-04-11
chore(dynamite,dynamite_runtime,nextcloud)!: generated clients no lon…

Leptopoda:
…ger implement http.Client

It was an error to let all generated clients implement http.Client. A user could get confused which client to use.
This also reduces our api surface and potetials for bugs.
I do not think that a `CookieJarClient` must be shipped with the runtime and moved it into the nextcloud pacakge where it is actually used (just copied the existing unit test).

 This change is not as breaking as one might think because the commit making the clients an http.Client has not landed in any release yet.


provokateurin:
1. This is a refactor not a chore
2. The cookie functionality is pretty useful in general, so I don't see why we shouldn't keep it in dynamite_runtime.

Leptopoda:
> The cookie functionality is pretty useful in general, so I don't see why we shouldn't keep it in dynamite_runtime.

I really, really dislike the current api the cookie_jar package provides and do not want to force other users to also import that package.
I also think that it is more than trivial to create a custom client that serves that functionality.

provokateurin (approved):

PR #1910, opened 2024-04-13
feat(neon_talk): Implement rich objects rendering

provokateurin:
The algorithm for splitting the message and replacing the rich objects might not be the best one, but I found this to be relatively simple and I couldn't think of a much better solution.

The first commit basically backports a change that will also come with https://github.com/nextcloud/neon/pull/1894, so we might just want to wait for 29 to land before merging this.

Not all parameter types are supported yet, because i wasn't able to test the maps integration for geo-location (I think it is broken) and displaying vcards can be done later as it is not that important imo. Polls are also not implemented, but that's because they are a whole different thing we need to support and it would bloat this PR which just gets the basics into place.


packages/neon/neon_talk/lib/src/widgets/rich_object/mention.dart
```diff
@@ -0,0 +1,51 @@
...
+    return Chip(
+      shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(50)),
+      padding: EdgeInsets.zero,
+      labelPadding: const EdgeInsetsDirectional.only(end: 8),
+      avatar: Padding(
+        padding: const EdgeInsets.all(4),
+        child: child,
+      ),
+      label: Text(parameter.name),
+    );
```
Leptopoda, line 49: The padding is the same as used in the `TalkRichObjectFallback` widget.
Please add them to a `TalkTheme` ThemeExtension that can be registered in the AppImplementation:
https://github.com/nextcloud/neon/blob/70eb554f1e9a9fa079e5365a2f0eb871296e9700/packages/neon_framework/lib/src/models/app_implementation.dart#L192

Leptopoda (changes requested):
> The algorithm for splitting the message and replacing the rich objects might not be the best one, but I found this to be relatively simple and I couldn't think of a much better solution.

I didn't try to 100% understand the algorithm but if it does the job let's go with it.

---

Screenshot_20240414_111508

I think that the `TalkRichObjectMention` has to respect the text scale. Optimally the entire chip should not be larger than the surrounding text. Now the text in the chip is already larger.

packages/neon/neon_talk/lib/src/widgets/rich_object/file.dart
```diff
@@ -0,0 +1,91 @@
...
+
+  /// The parameter to display.
+  final spreed.RichObjectParameter parameter;
+
+  @override
+  Widget build(BuildContext context) {
+    Widget child;
+
+    if (parameter.previewAvailable == spreed.RichObjectParameter_PreviewAvailable.yes) {
+      final maxHeight = MediaQuery.of(context).size.height ~/ 2;
```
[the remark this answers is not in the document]
provokateurin: Yes, I wanted to limit the max height an image can take up on the screen to avoid it filling out the whole screen at once.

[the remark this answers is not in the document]
provokateurin: Do you think there is a better way to do this? The image should not always just take the full original size.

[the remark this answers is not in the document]
Leptopoda: I just wanted to mention that `MediaQueryData.size` usually is the screen size of the device and not the widget size (depending on where the nearest MediaQuery ancestor is in the WidgetTree).
If the widget is constrained or other things on the screen take up room like an appBar they are not considered.
I _think_ that a LayoutBuilder is preffered in this scenario.

The algorithm itself is ok. Just the way you obtain the height should change.

packages/neon/neon_talk/lib/src/widgets/rich_object/mention.dart
provokateurin: I'll wait for https://github.com/nextcloud/neon/pull/1849

packages/neon/neon_talk/lib/src/widgets/rich_object/file.dart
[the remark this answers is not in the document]
provokateurin: LayoutBuilder won't work because it is inside a list view so the max height constraints is infinity. I think using the window size is fine (it's not the screen size btw).

packages/neon/neon_talk/lib/src/widgets/rich_object/mention.dart
provokateurin: The chips have diverged now to make the mentions in line with the text it would be embedded in and they don't share any common values anymore.

packages/neon/neon_talk/lib/src/widgets/rich_object/file.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  /// The TextStyle to applied to all text elements in this rich object.
+  final TextStyle? textStyle;
+
+  @override
+  Widget build(BuildContext context) {
+    Widget child;
+
+    if (parameter.previewAvailable == spreed.RichObjectParameter_PreviewAvailable.yes) {
+      final maxHeight = MediaQuery.sizeOf(context).height ~/ 2;
```
Leptopoda, line 31:
```suggestion
      final maxHeight = MediaQuery.sizeOf(context).height / 2;
```

We do not need to have it as an integer.

Leptopoda (commented):
Is it possible to propagate the textstyle through the BuildContext? That should allow future mistakes.

I suggest to wrap the entire systemMessage in a `DefaultTextStyle` widget and let the children build according to the current default style.

packages/neon/neon_talk/lib/src/widgets/rich_object/file.dart
provokateurin: Not at that point, but later on we only need the integer. It's easier and more performant to convert the double to integer as early as possible.

provokateurin:
> I suggest to wrap the entire systemMessage in a DefaultTextStyle widget and let the children build according to the current default style.

This doesn't work for the Chips because they rely on the ChipThemeData and do not consider the DefaultTextStyle. I think we will just have to live with it, even though I like your idea.

packages/neon/neon_talk/lib/src/widgets/rich_object/file.dart
Leptopoda: Why do we need the maxHeight as an integer?
In lines 45-46 a difference of op to `0.5` wouldn't be that significant.

And further down the maxHeight is used as a double

Leptopoda (approved):

PR #1913, opened 2024-04-14
feat(neon_talk): Implement sending messages

provokateurin:
https://github.com/nextcloud/neon/issues/356

The creation of the bloc is now moved outside the page so it can be properly tested, but the page is still responsible for disposing it. This also mitigates the problem of https://github.com/nextcloud/neon/pull/1788 where the potentially outdated room was used.

I also added tests for the other elements on the room page.


packages/neon/neon_talk/lib/src/pages/main.dart
```diff
@@ -106,7 +107,10 @@ class _TalkMainPageState extends State<TalkMainPage> {
...
           MaterialPageRoute<void>(
             builder: (context) => TalkRoomPage(
-              room: room,
+              bloc: TalkRoomBloc(
+                account: NeonProvider.of<AccountsBloc>(context).activeAccount.value!,
+                room: room,
+              ),
             ),
           ),
         );
```
Leptopoda, line 116: I can't make a suggestion but please just wrap the RoomPage in a NeonProvider.
It will automatically handle the disposal of the bloc when the widget is removed from the subtree:
```dart
        await Navigator.of(context).push(
          MaterialPageRoute<void>(
            builder: (_) => NeonProvider(
              create: (context) {
                final account = NeonProvider.of<AccountsBloc>(context).activeAccount.value!;

                return TalkRoomBloc(
                  account: account,
                  room: room,
                );
              },
              child: TalkRoomPage(),
            ),
          ),
        );
```

Ref: https://github.com/nextcloud/neon/blob/70eb554f1e9a9fa079e5365a2f0eb871296e9700/packages/neon_framework/lib/src/utils/provider.dart#L96-L101

Leptopoda (commented):
This is really cool :heart_eyes:

Leptopoda (approved):

PR #1923, opened 2024-04-14
fix(dynamite): build config to allow built_value_generator running inside dynamite

Leptopoda:
before that running build_runner inside dynamite also tried to execute dynamite itself.
Also formats the build.yaml to use single quotes. The same we have everywhere else in the project.


provokateurin (approved):
Great stuff, this was so annoying :sweat_smile:

PR #1931, opened 2024-04-16
refactor(dynamite): split JsonSchema into separate library

Leptopoda:
This reorganizes the `Schema` handling by splitting of the validators and general schema handling.
Schemas are now split per types where only the relvant attributes are taken into account.

This will allow easier integration of #1247 and broader cleanups by only passing a specific schema type arround.


packages/dynamite/dynamite/lib/src/builder/resolve_type.dart
```diff
@@ -145,7 +145,7 @@ TypeResult _resolveType(
 
     case openapi.Schema(type: openapi.SchemaType.array):
       final TypeResult subResult;
-      if (schema.maxLength == 0) {
+      if (schema.maxItems == 0) {
```
provokateurin, line 148: Please remove this whole commit for now. Patching all the specs is not feasible.
We will just have to wait until https://github.com/nextcloud/neon/pull/1894 and accept that both dynamite and the current specs have this problem.

packages/dynamite/dynamite/lib/src/models/json_schema/schema.dart
```diff
@@ -146,3 +339,63 @@ class SchemaPlugin implements SerializerPlugin {
   @override
   Object? beforeSerialize(Object? object, FullType specifiedType) => object;
 }
+
+class _JsonSchemaSerializer extends StructuredSerializer<JsonSchema> {
```
provokateurin, line 343: I see no handling for `$ref`, will all those go into EmptySchema because they have no `type` set?

provokateurin: Ah yes this seems to be the case according to the new tests. TBH I find that a bit confusing because I would expect EmptySchema to be equivalent to a `{}` schema and nothing else.

provokateurin (changes requested):

packages/dynamite/dynamite/lib/src/builder/resolve_type.dart
Leptopoda: Then this PR has to wait.

I will not mixin the `StringValidator` into the `ArraySchema` just to reproduce this issue with the latest commit.

packages/dynamite/dynamite/lib/src/models/json_schema/schema.dart
Leptopoda: Yeah I'm also not a fan.

What about renaming the `JsonSchema` into `BaseSchema` and `EmptySchema` to `JsonSchema`?

packages/dynamite/dynamite/lib/src/builder/resolve_type.dart
provokateurin: You're right. Another option would be to remove the first and the last commit and keep everything else as that can be merged easily.

packages/dynamite/dynamite/lib/src/models/json_schema/schema.dart
provokateurin: To me that also feels wrong as I'd expect all other schemas to then extend JsonSchema which they don't. The EmptySchema is only necessary to be able to distinguish it from the other schemas, right? Is that even necessary anywhere?
My idea: Rename EmptySchema to GenericSchema which at least somewhat indicates that it can also have some values but it is not exactly decided which ones they are.

packages/dynamite/dynamite/lib/src/builder/resolve_type.dart
Leptopoda: I'm not sure it is worth it.
We do not rely on these cleanups so they can just wait (unless you would like to have some of them pulled in already).

provokateurin: I mean cleanups are good, but if it blocks no other work and just creates additional work now, we can just wait.

Leptopoda: I just rebased onto #1894.
Should only be a couple of days

packages/dynamite/dynamite/lib/src/helpers/pattern_check.dart
```diff
@@ -1,14 +1,13 @@
...
-import 'package:dynamite/src/models/openapi/schema.dart';
 
 Iterable<Expression> buildPatternCheck(
   openapi.Schema schema,
   String value,
   String name,
 ) sync* {
   switch (schema.type) {
-    case SchemaType.string:
+    case openapi.SchemaType.string:
```
provokateurin: Shouldn't the SchemaType also move?

provokateurin (changes requested):
The second commit should also rename SchemaType to JsonSchemaType. Everything else LGTM

provokateurin (approved):

PR #1938, opened 2024-04-19
fix(neon_framework): Create new http.Request every time when retrying failed requests

provokateurin:
The same request can not be used multiple times so a new request has to be created every time.

### DIFF
--- a/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
+++ b/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
@@ -70,7 +70,7 @@ class _DashboardBloc extends InteractiveBloc implements DashboardBloc {
         account: account,
         cacheKey: 'dashboard-widgets',
         subject: widgets,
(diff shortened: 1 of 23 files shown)


Leptopoda (commented):
Please add a test to avoid future regressions

provokateurin:
I don't think this is testable in a way that makes sense.
We can not add a test for every call to the wrap methods that ensures a new test is created all the time.
We can also not test the RequestManager because it relies of the method signature to ensure new requests are created every time. Adding a test for that would literally test nothing except that there is a callback that gets called.
The only thing that would be testable is to check that the callback is called multiple times when the request fails and is retried, but I don't know how valuable that is (but I can add it if you want).

Leptopoda:
I'd like to have the "retry calls the request multiple times" test.

I was well aware of the limitation that a request can not be retried but still made that fundamental mistake (So I'd probably do it again with the next refractor)

Leptopoda (approved):

Issue #1905, opened 2024-04-11
Default values and patter/length checks are not inherrited


Leptopoda:
When generating a schema that inherits parameters from a super schema the builtvalue `_initialize` and `_finalize` hooks will not be included in the subschema.

```json
            "SuperObject": {
                "type": "object",
                "properties": {
                    "value": {
                        "type": "string",
                        "pattern": "^[0-9]*$",
                        "minLength": 3,
                        "maxLength": 20,
                        "default": "123"
                    }
                },
                "required": [
                    "value"
                ]
            },
            "SubObject": {
                "type": "object",
                "allOf": [
                    {
                        "$ref": "#/components/schemas/SuperObject"
                    }
                ]
            }
```

provokateurin:
https://github.com/google/built_value.dart/issues/1310 for reference

PR #1945, opened 2024-04-20
Fix/dynamite/inheritt_defaults_and_vaidation

Leptopoda:
fixes: #1905

With further refactors we may be able to remove the generation of the empty methods but for now this is the easiest we can do to fix this issue.

### DIFF
--- a/packages/dynamite/dynamite/lib/src/builder/resolve_interface.dart
+++ b/packages/dynamite/dynamite/lib/src/builder/resolve_interface.dart
@@ -42,6 +42,8 @@ Spec buildInterface(
   BuiltList<Method>? methods,
   Iterable<TypeResultObject>? interfaces,
   String? documentation,
+  Iterable<String>? defaults,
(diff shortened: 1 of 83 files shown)


packages/dynamite/dynamite/lib/src/builder/resolve_interface.dart
```diff
@@ -62,45 +30,18 @@ Spec buildInterface(
...
-              return generateProperty(
-                result,
-                propertyName,
-                propertySchema.formattedDescription,
-              );
-            }),
-          );
+      for (final schema in allOf) {
+        if (schema case openapi.Schema(properties: != null)) {
+          _generateProperties(schema, spec, state, identifier, b, defaults, validators, isHeader: isHeader);
```
provokateurin, line 38: Add a comma after the last argument to make it more readable

provokateurin (approved):
Nice work!

PR #1946, opened 2024-04-21
perf(dynamite): use memoization for for better performance

Leptopoda:


provokateurin (approved):
How much faster is this?

Leptopoda:
We are getting into the hard to measure teritory.
The print changes also yielded some drastic improvements and all together we get a reduction of ~10s.

PR #1947, opened 2024-04-21
fix(dynamite): content and contenttype requirements

Leptopoda:
> A map containing the representations for the parameter. The key is the media type and the value describes it. The map MUST only contain one entry.

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#parameter-object

> **REQUIRED.** The content of the request body. The key is a media type or media type range and the value describes it. For requests that match multiple keys, only the most specific key is applicable. e.g. text/plain overrides text/*

https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#request-body-object


packages/dynamite/dynamite/lib/src/models/openapi/parameter.dart
```diff
@@ -44,15 +44,14 @@ abstract class Parameter implements Built<Parameter, ParameterBuilder> {
       return $schema;
     }
 
-    if (content != null && content!.isNotEmpty) {
-      if (content!.length > 1) {
-        print('Can not work with multiple mime types right now. Using the first supported.');
-      }
+    if (content != null) {
+      final test = content!.entries.single;
```
provokateurin, line 48:
```suggestion
      final mediaType = content!.entries.single;
```

provokateurin (approved):

Issue #1263, opened 2023-12-09
Reduce amount of warnings in dynamite


provokateurin:
We have some warnings that are specific to certain things like only supporting one response schema which should always be logged, but others like the integer precision should only be printed once and not every time we encounter an integer.

PR #1948, opened 2024-04-21
refactor(dynamite): use logger to display warnings

Leptopoda:
fixes: #1263


provokateurin (changes requested):
The no_print rule should be enabled now.
Does this work across specs or do we get the log messages per spec?

Leptopoda:
It will work across specs. Sadly there is only one dynamite instance running at a given time.
I wanted to make the `State` a singleton and learned it the hard way :(

provokateurin (approved):

PR #1953, opened 2024-04-22
fix(dynamite): json schema reference resolving

Leptopoda:
I did not find a better solution yet.
In theory this already allows us to fetch schemas fro other sources like local filesystem or remote servers.
I did not observe any performance penalty.

Towards: #933

### DIFF
--- a/packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
+++ b/packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
@@ -5,7 +5,8 @@ import 'package:built_value/serializer.dart';
 import 'package:dynamite/src/helpers/docs.dart';
 import 'package:dynamite/src/helpers/logger.dart';
 import 'package:dynamite/src/models/exceptions.dart';
-import 'package:dynamite/src/models/openapi/discriminator.dart';
+import 'package:dynamite/src/models/openapi.dart';
+import 'package:rfc_6901/rfc_6901.dart';
 
 part 'schema.g.dart';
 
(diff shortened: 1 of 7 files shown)


packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
```diff
@@ -16,8 +17,37 @@ abstract class Schema implements Built<Schema, SchemaBuilder> {
...
+  @BuiltValueField(wireName: r'$id')
+  Uri? get id;
+
   @BuiltValueField(wireName: r'$ref')
-  String? get ref;
+  Uri? get ref;
+
+  Schema buildRef(Map<String, dynamic> json) {
+    if (ref == null) {
+      throw StateError('Referenced schema can only be built when a ref is present');
```
provokateurin, line 28:
```suggestion
      throw StateError(r'Referenced schema can only be built when a $ref is present');
```

packages/dynamite/dynamite/lib/src/models/openapi/schema.dart
```diff
@@ -16,8 +17,37 @@ abstract class Schema implements Built<Schema, SchemaBuilder> {
...
   static Serializer<Schema> get serializer => _$schemaSerializer;
 
+  @BuiltValueField(wireName: r'$id')
+  Uri? get id;
+
   @BuiltValueField(wireName: r'$ref')
-  String? get ref;
+  Uri? get ref;
+
+  Schema buildRef(Map<String, dynamic> json) {
```
provokateurin, line 26:
```suggestion
  Schema resolveRef(Map<String, dynamic> json) {
```
I think this better reflects what it does. The ref isn't built, the ref is resolved and then the schema for it is built

provokateurin (changes requested):
Could this have some tests? I see that it is not breaking anything because none of the generated code has changed, but that is no guarantee it won't break in the future.

Leptopoda:
We already had the e2e tests for this but I added some directly dynamite nevertheless

provokateurin (approved):

provokateurin:
Commit message has a typo and should be `fix(dynamite): json schema reference resolving`

PR #1964, opened 2024-04-24
feat(dynamite): do not relly on the openapi spec for type resolving

Leptopoda:
implements: #933


packages/dynamite/dynamite/build.yaml
```diff
@@ -1,8 +1,15 @@
 builders:
-  dynamite:
-    import: 'package:dynamite/builder.dart'
```
provokateurin, line 3: If the `dynamite` builder is removed then why do the other build.yaml files switch to `dynamite|dynamite_openapi`?

packages/dynamite/dynamite/lib/src/json_schema_builder.dart
```diff
@@ -0,0 +1,149 @@
...
+      }
+    }
+
+    final inputId = buildStep.inputId;
+    final outputId = inputId.changeExtension('.dart');
+
+    try {
+      final json = jsonDecode(await buildStep.readAsString(inputId)) as Map<String, dynamic>;
+
+      final schema = openapi.serializers.deserializeWith(
```
provokateurin, line 54: Why does this call into the openapi code?

packages/dynamite/dynamite/lib/src/json_schema_builder.dart
provokateurin, line 1: Quite a bit of this code is just copied from the existing builder and should be factored out into common methods.

packages/dynamite/dynamite_end_to_end_test/lib/json_schema/object.json_schema.json
provokateurin, line 1: This is not a schema definition: https://json-schema.org/learn/getting-started-step-by-step#create
We should really implement the whole thing and not just the schema part itself.

provokateurin (changes requested):
First commit message has a typo and should be `feat(dynamite): do not rely on the openapi spec for type resolving`

packages/dynamite/dynamite/build.yaml
provokateurin: Ah ok :+1:

packages/dynamite/dynamite_end_to_end_test/lib/json_schema/object.json_schema.json
Leptopoda: I don't understand? what is still missing?

provokateurin: Sorry my initial message was a bit confusing, I'd simply like to have proper support for JSON schema definitions.

Leptopoda: Can you please say what JsonSchema keywords you'd like to support?

It appears that the only user facing information is title and I don't think that it should be set as a library level doc comment. I'd rather add it to the generated class itself similar to how the dart doc style expects the documentation:

```dart
/// {title}.
///
/// {description}.
class MyObject {}
```

Leptopoda:
Making a draft for now until we have merged all the depending changes.

packages/dynamite/dynamite_end_to_end_test/lib/json_schema/object.json_schema.json
provokateurin: Yes the comment should be on the generated class. IMO the `$id` can also be relevant for users, so I'd like to have that linked as well.

provokateurin:
`$defs` should also be supported as it acts like the `components.schemas`

packages/dynamite/dynamite_end_to_end_test/lib/json_schema/object.json_schema.json
provokateurin: Since there is `$defs` it is possible to have multiple object inside a single schema. I don't know if the `title` and `description` should be interpreted as the title and description for the main object or the whole schema.

Leptopoda: In theory we already support `$defs`. our reference resolving is independent from the dart schema object and directly searches in the json data (see: #1953).

Therefore only the schemas referenced in the main schema will be generated. Or do you mean to generate all schemas defined in the definitions just in case :woman_shrugging:

provokateurin: No this was about the other fields, not specifically $defs. I was just saying that title and description might not be for the main schema but for the whole schema which would make them library level.

provokateurin:
@Leptopoda can we move this forward and include it in the next batch of releases or is there something holding it back?

Leptopoda:
I don't think this is anything we can immediately implement.

 > $defs should also be supported as it acts like the components.schemas
 
 This requires a complete rework of how our `$ref` resolving works and this will take quite some time ...

PR #1980, opened 2024-04-26
feat(dynamite_runtime): support all common schema validators

Leptopoda:
fixes: #1247


packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
provokateurin, line 1: These methods should only ever be used by generated code, so it's safe to just remove them right away without deprecation.

packages/dynamite/dynamite_runtime/lib/src/utils/json_schema_validation.dart
```diff
@@ -0,0 +1,96 @@
...
+  int? minLength,
+  RegExp? pattern,
+}) {
+  String value;
+  if (input is JsonObject && input.isString) {
+    value = input.asString;
+  } else if (input is String) {
+    value = input;
+  } else {
+    return;
```
provokateurin, line 23: Shouldn't this rather throw an argument error because the method was called with the wrong data?

packages/dynamite/dynamite_runtime/lib/src/utils/json_schema_validation.dart
```diff
@@ -0,0 +1,96 @@
...
+  if (minItems != null && length < minItems) {
+    throw FormatException('Parameter "$parameterName" has to be at least $minItems items long but was $length long');
+  }
+
+  if (maxItems != null && length > maxItems) {
+    throw FormatException('Parameter "$parameterName" has to be at most $maxItems items long but was $length long');
+  }
+
+  late final value = switch (input) {
+    BuiltSet() || SetBuilder() || Set() => null,
```
provokateurin, line 84: Shouldn't it use `input` for these cases?

provokateurin (changes requested):

packages/dynamite/dynamite_runtime/lib/src/utils/json_schema_validation.dart
Leptopoda: I don't think so; ArrayValidator.maxItems for example says:
>  An array instance is valid against "maxItems" if its size is less than, or equal to, the value of this keyword.

https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-01#section-6.4.1

To my understanding any non `Iterable` value is always valid. Dynamite shouldn't ever try to apply such a validation to a non `JsonObject`, `Iterable` like or `Null` object but why potentially break at runtime when it is not wrong to just return.

packages/dynamite/dynamite_runtime/lib/src/utils/json_schema_validation.dart
Leptopoda: We know that these types only hold unique items anyways so we can just skip the check altogether.
If you want I can change this.

packages/dynamite/dynamite_runtime/lib/src/utils/array_checker.dart
Leptopoda: Yes but it must be a separate commit as this is breaking to `dynamite_runtime` but if it was in the same commit melos would also count it as breaking in `nextcloud` so I just deprecated it for now.

I'll add another commit.

packages/dynamite/dynamite_runtime/lib/src/utils/json_schema_validation.dart
provokateurin: Ah you're absolutely right, I didn't think about that :sweat_smile: Maybe add a comment, otherwise I'll ask this myself again at some point.

provokateurin (approved):

PR #1984, opened 2024-04-27
refactor(ci): Only build docker images when needed

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1453


Leptopoda:
This means we have to manually build the images if we change the `build-dev-container.sh` script?

provokateurin:
Yes, but it also doesn't matter for the CI because in that case the images are not needed for running tests.

Leptopoda (approved):

PR #1987, opened 2024-04-28
fix(nextcloud)!: Require content-length for all WebDAV put requests

provokateurin:
Some specific setups do not like it when the content-length header is not present. The consumer of the package should always know the size of the data they are trying to upload, so enforcing the content-length is no problem.
For reference: https://github.com/saber-notes/saber/issues/945 https://github.com/saber-notes/saber/issues/1232

### DIFF
--- a/packages/nextcloud/lib/src/webdav/client.dart
+++ b/packages/nextcloud/lib/src/webdav/client.dart
@@ -109,9 +109,9 @@ class WebDavClient {
 
     _addUploadHeaders(
       request,
+      localData.length,
       lastModified: lastModified,
       created: created,
-      contentLength: localData.length,
     );
     _addBaseHeaders(request);
     return request;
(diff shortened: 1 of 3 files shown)


packages/nextcloud/lib/src/webdav/client.dart
```diff
@@ -146,23 +146,23 @@ class WebDavClient {
   ///   * [putStream] for a complete operation executing this request.
   http.BaseRequest putStream_Request(
     Stream<List<int>> localData,
+    int contentLength,
```
Leptopoda, line 149: I'd like to leave this as a named parameter.
1. easier to read code
2. less breaking for our users

Leptopoda (commented):
This means  another breaking release shortly?

provokateurin:
I don't like having a breaking release because this doesn't fix anything that is actually broken. Maybe it makes more sense to add an assert or print a log to inform the developers they are probably making a mistake? I'm not sure what the best way is... I'd also be open to not merge this patch at all and just add some doc comments that explain why it is important to set the parameter :woman_shrugging:

Leptopoda:
I'm fine with either way.

PR #1989, opened 2024-04-28
fix(nextcloud): Make WebDAV get operations work with chunked responses

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1988

I'm unsure how to actually test this, as the PHP internal server doesn't return chunked responses itself. The only way I see is by mocking the client and returning a chunked response manually, but that could still go wrong.

### DIFF
--- a/packages/nextcloud/lib/src/webdav/client.dart
+++ b/packages/nextcloud/lib/src/webdav/client.dart
@@ -294,25 +294,23 @@ class WebDavClient {
 
     unawaited(
       response.then(
-        (response) async {
+        (response) {
           final contentLength = response.contentLength;
-          if (contentLength == null || contentLength <= 0) {
-            onProgress?.call(1);
-          } else {
-            final completer = Completer<void>();
-            var downloaded = 0;
+          var downloaded = 0;
 
-            response.stream.listen((chunk) async {
+          response.stream.listen(
+            (chunk) async {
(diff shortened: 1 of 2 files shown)


provokateurin:
@Leptopoda any ideas how to make a test for this?

Leptopoda:
In theory this should be doable with a plain `MockClient` bypassing our server setup altogether:
```dart
  test('webdav get chunked response', () {
    final client = NextcloudClient(
      Uri(),
      httpClient: http.MockClient.streaming((request, bodyStream) {
        // Send chunked data but do not set the contentlength header
      }),
    );

    client.webdav.getStream(path);
  });
```

provokateurin:
Yeah, but right now the nextcloud package only tests against real servers and I'd like to keep it that way. Mocking HTTP responses doesn't test anything here.
Do you want to see a test for this or not? I'd be fine with merging this without a new test as I don't consider a test with mocks of high value.

Leptopoda:
Please add a test.

provokateurin:
The test was written with https://github.com/dart-lang/sdk/blob/main/tests/standalone/io/http_content_length_test.dart as reference because I couldn't get it working with a mocked client for some reason. For a test this is totally fine IMO and even better than a mock response.

Leptopoda (approved):

PR #1994, opened 2024-04-30
fix(neon_dashboard): Only allow single lines in items

provokateurin:
This can happen for example with formatted Talk messages that have more than one line which is used as the subtitle. Did the same for the title to prevent the same problem there if it would ever happen.


Leptopoda:
I guess a unit test is overkill here?

Leptopoda (approved):

Issue #1995, opened 2024-04-30
Interfaces should be used everywhere instead of types


provokateurin:
Currently it is not possible to fully utilize the type hierarchy because many places just assume the type instead of it's interface:

https://github.com/nextcloud/neon/blob/42551c66386da2740bf10b21b363f183bab55bd6/packages/nextcloud/lib/src/api/spreed.openapi.dart#L16655 should be `typedef Room_LastMessage = ({$BaseMessageInterface? baseMessage, BuiltList<Never>? builtListNever, $ChatMessageInterface? chatMessage});` instead.

https://github.com/nextcloud/neon/blob/42551c66386da2740bf10b21b363f183bab55bd6/packages/nextcloud/lib/src/api/spreed.openapi.dart#L16759 should be `$OCSMetaInterface get meta;` instead.

Leptopoda:
I recently thought about this issue when working on cookbook and some initial tests where sadly quite breaking and iirc also having issues with the serializer.

I also thought about hiding the default implementation similar to the blocs in the neon code but this would hide the `BuiltValue` type which was also undesirable:
```dart
class TestInterface{
factory TestInterface() = Test;
}

class Test implements TestInterface{
}
```

I'll take another look in the coming days.

Leptopoda:
Ok this is really hard to do with our current dynamite code.
I'm already working on changing this limitation but it will take some time.

Hiding the implementation and exposing it through a factory in the interface is not possible for various limitations in built_value_generator.
1. it can not generate serializers for hidden classes (due to potential naming conflicts)
2. it does not allow any constructor on an interface class (could be changed upstream but I doubt it will be accepted)

provokateurin:
A simple but dirty workaround for this is calling toJson/fromJson to convert between classes that use the same interface.

PR #2000, opened 2024-05-02
feat(dynamite)!: add `Built` interface to the generated interfaces

Leptopoda:
Allows calling `toBuilder` and `rebuild` on the interface making it a drop in replacement to the built class itself.
related to #1995 

As this is not needed yet I'd like to wait until other breaking changes accumulate.


provokateurin (approved):
2000 :tada:

provokateurin:
@Leptopoda I think this commit shouldn't be marked as breaking and we can merge it right away because you only add new methods and don't remove or change existing ones.

provokateurin:
Yep, nobody should implement these interfaces...

provokateurin:
After https://github.com/nextcloud/neon/pull/2190 this is no longer a breaking change as really nobody else can implement the interface now.

PR #2001, opened 2024-05-02
refactor(dynamite): handle AssetIds in the State

Leptopoda:
the remaining unprefixed imports are referenced in the `TypeResult` code.
Major refactors for that are already on the way.


packages/dynamite/dynamite/lib/src/openapi_builder.dart
```diff
@@ -78,7 +78,9 @@ class OpenAPIBuilder implements Builder {
             Directive.import('dart:typed_data'),
             Directive.import('package:built_collection/built_collection.dart'),
             Directive.import('package:built_value/json_object.dart'),
-            Directive.import('package:collection/collection.dart'),
+            // The allocator will always start at _i1.
+            // It is safe to use 0.
+            Directive.import('package:collection/collection.dart', as: '_i0'),
```
provokateurin, line 83: So this is manually imported and only used for extensions, right?

Leptopoda: yes.
Thinking more about it I we should probably switch it to: `_i0.IterableExtension(authentications)?.firstWhereOrNull`
This removes the special case and will only add the import if used (helping us with #1244)

provokateurin (approved):
Typo in the commit: `refactor(dynamite): use a custom allocator to hanld the part directive` should be `refactor(dynamite): use a custom allocator to handle the part directive`

PR #2020, opened 2024-05-08
refactor(neon_framework): Replace direct Blocs access with BehaviorSubjects

provokateurin:
This makes testing easier because no mock blocs need to be created just to pass the subjects along. Also reduces chance of mixing up business logic or making mistakes.


Leptopoda (approved):

provokateurin:
Oops something in this change causes a stackoverflow...

PR #2024, opened 2024-05-10
fix(neon_framework): Fix cyclic dependency between AccountOptions and AppsBloc

provokateurin:
Fixes https://github.com/nextcloud/neon/pull/2020
I had to revert some of the changes of that PR because the two methods were calling each other causing an infinite loop resulting in a stackoverflow.


Leptopoda (approved):
makes way more sense :+1:

PR #2027, opened 2024-05-10
refactor(neon_framework): Cleanup version checks

provokateurin:
Exposes all checks which will be necessary for apps to check what version is actually used.

### DIFF
--- a/packages/neon_framework/lib/src/blocs/apps.dart
+++ b/packages/neon_framework/lib/src/blocs/apps.dart
@@ -12,7 +12,6 @@ import 'package:neon_framework/src/models/notifications_interface.dart';
 import 'package:neon_framework/src/utils/account_options.dart';
 import 'package:neon_framework/src/utils/findable.dart';
 import 'package:neon_framework/src/utils/request_manager.dart';
-import 'package:neon_framework/src/utils/server_version.dart';
 import 'package:nextcloud/core.dart' as core;
 import 'package:nextcloud/nextcloud.dart';
 import 'package:provider/provider.dart';
@@ -50,8 +49,11 @@ abstract class AppsBloc implements InteractiveBloc {
   /// A subject emitting an event when the notifications page should be opened.
   BehaviorSubject<void> get openNotifications;
 
-  /// A collection of unsupported apps and their minimum required version.
+  /// A collection of apps and their version checks.
   BehaviorSubject<BuiltMap<String, VersionCheck>> get appVersionChecks;
+
+  /// A collection of unsupported apps and their minimum required version.
+  BehaviorSubject<BuiltMap<String, VersionCheck>> get unsupportedApps;
 
   /// Returns the active [Bloc] for the given [appImplementation].
   ///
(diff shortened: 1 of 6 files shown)


packages/neon_framework/lib/src/blocs/apps.dart
```diff
@@ -50,7 +49,7 @@ abstract class AppsBloc implements InteractiveBloc {
   /// A subject emitting an event when the notifications page should be opened.
   BehaviorSubject<void> get openNotifications;
 
-  /// A collection of unsupported apps and their minimum required version.
+  /// A collection of apps and their version checks.
   BehaviorSubject<BuiltMap<String, VersionCheck>> get appVersionChecks;
```
Leptopoda, line 53: Is there a reason we expose all version checks?
I think we should have a map with only the unsupported apps.
This would move the logic from https://github.com/nextcloud/neon/pull/2027/files#diff-543b9d4d76180fb9419fc101e9ba86d855ec85948bcb43bfd6fc9685433785bbR57 into the bloc where it's supposed to be

packages/nextcloud/lib/src/helpers/common.dart
```diff
@@ -11,7 +11,9 @@ class VersionCheck {
     required this.versions,
     required this.minimumVersion,
     required int? maximumMajor,
-  }) : maximumMajor = maximumMajor ?? minimumVersion.major;
+    bool? isSupportedOverride,
```
Leptopoda, line 14: I don't see any place where this is `false` and I can't imagine in what use case we'd like to do that.
What about making this non nullable with a default value of `false`.

Leptopoda (changes requested):

packages/neon_framework/lib/src/blocs/apps.dart
provokateurin: I did this because we will have to check in a few places which features are supported. Talk does it already really well by exposing the features through a list in the capabilities, but basically any other app needs to rely on the version. For example the version check will be important at runtime for https://github.com/nextcloud/neon/issues/2010 or endpoints only available through https://github.com/nextcloud/neon/issues/1492. Of course this will go away at some point once those features are supported by all versions we support, but that can take a while (minimum 1 year for server).

packages/nextcloud/lib/src/helpers/common.dart
provokateurin: It would be possible to exclude some versions that are known to be bad (e.g. 10.0 works, 10.1 is broken, but 10.2 is fine again and 10.1 could be blocked this way). I'm not planning to implement this at the moment, so I can also change it like you suggested.

packages/neon_framework/lib/src/blocs/apps.dart
Leptopoda: That makes sense.
Please add the unupported apps as a getter to the app nevertheless.

Leptopoda (approved):

Issue #2003, opened 2024-05-03
Improve server icons


provokateurin:
https://github.com/nextcloud/server/blob/master/core/src/icons.js generates the CSS rules for the server icons. It is more complex than our naive approach, so not everything works (almost everything does though).
We should follow the code in the script and map our icons accordingly. Instead of parsing that file it should be easier to just parse https://github.com/nextcloud/server/blob/master/dist/icons.css and extract the icon names and SVGs from there.

devesh-2002:
May I work upon this issue?

provokateurin:
Sure! Let me know if you need any help. The existing script that generates the icons is at https://github.com/nextcloud/neon/blob/main/tool/generate-assets.sh

provokateurin:
@devesh-2002 how is it going? Do you have questions or need help implementing it?

devesh-2002:
Hello. I am very sorry I was not able to work upon this PR due to certain reasons. I will work upon it this week and update regarding it.

provokateurin:
No problem, I was just wondering if you needed and pointers.

provokateurin:
I had a closer look at this and it is quite a lot harder than I thought. Might not even be worth to put in the time to fully implement it unless we know something is actually broken.

provokateurin:
The easier way might even be to implement a new API in the server that allows resolving these icons (which could also be used in the CSS to avoid embedding all the SVGs).

PR #2029, opened 2024-05-11
feat(neon_framework): Improved server icons

devesh-2002:
Issue #2003, 
I have tried to parse the css file and extract icons from there.
Please let me know, if there are any changes needed.

### DIFF
--- a/tool/generate-assets.sh
+++ b/tool/generate-assets.sh
@@ -7,6 +7,28 @@ color="#f37736"
 rm -rf /tmp/nextcloud-neon
 mkdir -p /tmp/nextcloud-neon
 
+function parse_icons_css() {
+    icons_css_url="https://raw.githubusercontent.com/nextcloud/server/master/dist/icons.css"
+    icons_css_file="/tmp/icons.css"
+    wget "$icons_css_url" -O "$icons_css_file"
+    
+    icon_mapping=$(awk -F'[()]' '/background-image/ {print $2,$4}' "$icons_css_file")
+    
+    echo "$icon_mapping" > "/tmp/icon_mapping.txt"
+}
+function get_svg_url() {
+    icon_name="$1"
+    icon_svg_url=$(grep "$icon_name" /tmp/icon_mapping.txt | awk '{print $2}')
+    echo "$icon_svg_url"
+}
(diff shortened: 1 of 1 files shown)


tool/generate-assets.sh
```diff
@@ -7,6 +7,28 @@ color="#f37736"
...
+    icon_name="$1"
+    icon_svg_url=$(grep "$icon_name" /tmp/icon_mapping.txt | awk '{print $2}')
+    echo "$icon_svg_url"
+}
+
+function download_icon_svg() {
+    icon_name="$1"
+    icon_svg_url=$(get_svg_url "$icon_name")
+    wget "$icon_svg_url" -O "/tmp/$icon_name.svg"
+}
```
provokateurin, line 29: Unused

tool/generate-assets.sh
```diff
@@ -7,6 +7,28 @@ color="#f37736"
 rm -rf /tmp/nextcloud-neon
 mkdir -p /tmp/nextcloud-neon
 
+function parse_icons_css() {
+    icons_css_url="https://raw.githubusercontent.com/nextcloud/server/master/dist/icons.css"
```
provokateurin, line 11: This file should be read from the git submodule at external/nextcloud-server instead

provokateurin (changes requested):
The new code should replace the behavior of https://github.com/devesh-2002/neon/blob/843306c3ddf34e00ba60eb204703db2d6587900d/tool/generate-assets.sh#L88-L94, but instead of just reading the files from the paths it should extract the svgs contained in the css file as the icon files with their respective names.

tool/generate-assets.sh
devesh-2002: How would I do that?

provokateurin:
Given my comments in https://github.com/nextcloud/neon/issues/2003 this PR should probably be closed as this problem is a lot harder than I assumed and it's not really clear that the effort is even necessary as there are currently no known problems with the icons. @devesh-2002 would you be fine with that?

devesh-2002:
Yes, You can go ahead and close this PR. Thanks a lot for all the help!

PR #2046, opened 2024-05-15
feat(docs): mention the use of git submodules

Leptopoda:
See: https://github.com/nextcloud/neon/pull/2029#discussion_r1600925440


provokateurin (approved):
Using `--recursive` during `git clone` would be even easier, but this is also fine

PR #2053, opened 2024-05-20
refactor(neon_framework): Remove direct AccountsBloc usage where possible

provokateurin:
Unfortunately I was not able to split this refactor, but I think the result is pretty clean. There are some places left that still use the AccountsBloc because they actually need access to it, but everywhere else the Blocs and Account are accessed through NeonProvider.


packages/neon_framework/lib/src/app.dart
```diff
@@ -236,6 +240,49 @@ class _NeonAppState extends State<NeonApp> with WidgetsBindingObserver, WindowLi
...
+                      NeonProvider<WeatherStatusBloc>.value(
+                        value: _accountsBloc.getWeatherStatusBlocFor(account),
+                      ),
+                      NeonProvider<MaintenanceModeBloc>.value(
+                        value: _accountsBloc.getMaintenanceModeBlocFor(account),
+                      ),
+                    ],
+                    child: app,
+                  );
+                }
```
Leptopoda, line 283: I don't like having to specify this.
Can't we do some more magic 🧙‍♂️ by resolving the required bloc based on it's type in the NeonProvider itself?
https://github.com/nextcloud/neon/blob/6e9cbfbcd586fb4f819b4aee6afc98323af9e0ae/packages/neon_framework/lib/src/utils/provider.dart#L60-L61
This should also revert to the old way where only the AccountsBloc needs to be injected into the context while still being able to resolve other blocs (reverting quite a few changes of this PR)

packages/neon_framework/lib/src/widgets/account_tile.dart
```diff
@@ -29,27 +28,24 @@ class NeonAccountTile extends StatelessWidget {
   /// {@endtemplate}
   final Account account;
 
+  final UserDetailsBloc userDetailsBloc;
+
+  /// {@macro neon_framework.UserStatus.userStatusBloc}
+  final UserStatusBloc? userStatusBloc;
```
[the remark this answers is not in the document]
provokateurin: No, because `NeonProvider.of<UserStatusBloc>(context)` will always return the bloc of the active account and not necessarily of the account that is displayed here.

packages/neon_framework/lib/src/app.dart
provokateurin: I don't understand what you mean. Sure we could somehow create some matching logic, but it would still have to call a specific function based on the type which would end up with almost the same code and just a switch statement instead?

provokateurin:
(Just a rebase to solve the many conflicts)

packages/neon_framework/lib/src/app.dart
Leptopoda: You are right. My idea was to just have a function that dynamically loads the type needed instead of always attaching it to a widget. The current solution will instantiate all blocs at once while my idea was to only create them on the fly when accessed.

  I was not able to find any noticable performance regression. Let's keep it this way for now.

Leptopoda (approved):

packages/neon_framework/lib/src/app.dart
provokateurin: Ok now I see your point, but except for UnifiedSearchBloc and WeatherStatusBloc all of the blocs are always used and instantiated anyway.

PR #2063, opened 2024-05-22
feat(neon_talk): Support adding and removing reactions

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2037
The added patch is the equivalent of https://github.com/nextcloud/spreed/pull/12386
I know adding the reactions button to every message is a bit ugly, but I will fix that properly with https://github.com/nextcloud/neon/issues/2038 when a proper menu is introduced.


provokateurin:
Ah I forgot to add tests for the bloc part.

Leptopoda (approved):

PR #2067, opened 2024-05-22
feat(neon_framework): Respect the configured apps order

provokateurin:
Closes https://github.com/nextcloud/neon/issues/814
Unfortunately I can not add tests in this PR because the whole bloc is not tested yet. Just verifying manually that this works should be enough though.

### DIFF
--- a/packages/neon_framework/lib/src/blocs/apps.dart
+++ b/packages/neon_framework/lib/src/blocs/apps.dart
@@ -74,7 +74,24 @@ class _AppsBloc extends InteractiveBloc implements AppsBloc {
     required this.allAppImplementations,
   }) {
     apps.listen((result) {
-      appImplementations.add(result.transform((data) => filteredAppImplementations(data.map((a) => a.id))));
+      appImplementations.add(
+        result.transform(
+          (data) {
+            final apps = SetBuilder<AppImplementation>();
+
+            for (final entry in data) {
+              for (final app in allAppImplementations) {
+                if (app.id == entry.id || (app.additionalMatchingIDs?.contains(entry.id) ?? false)) {
+                  apps.add(app);
+                  break;
+                }
+              }
+            }
+
+            return apps.build();
+          },
+        ),
+      );
 
       if (result.hasData) {
         unawaited(updateApps());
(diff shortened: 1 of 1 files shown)


packages/neon_framework/lib/src/blocs/apps.dart
```diff
@@ -268,7 +266,9 @@ class _AppsBloc extends InteractiveBloc implements AppsBloc {
       subject: apps,
       getRequest: account.client.core.navigation.$getAppsNavigation_Request,
       serializer: account.client.core.navigation.$getAppsNavigation_Serializer(),
-      unwrap: (response) => response.body.ocs.data,
+      unwrap: (response) => response.body.ocs.data.rebuild(
+        (b) => b..sort((a, b) => (a.getOrder() ?? 100).compareTo(b.getOrder() ?? 100)),
```
Leptopoda, line 270: 1. why is `100` the number used?
2. please extract this into a separate constant
3. let `getOrder()` return the default order.

packages/neon_framework/lib/src/blocs/apps.dart
```diff
@@ -74,7 +74,19 @@ class _AppsBloc extends InteractiveBloc implements AppsBloc {
...
+        result.transform(
+          (data) => BuiltSet<AppImplementation>(
+            data
+                .map(
+                  (app) => allAppImplementations.firstWhereOrNull(
+                    (a) => a.id == app.id || (a.additionalMatchingIDs?.contains(app.id) ?? false),
+                  ),
+                )
+                .whereNotNull(),
+          ),
```
Leptopoda, line 87: Please manually loop over `data` and conditionally add the AppImplementation to the set.

packages/neon_framework/lib/src/blocs/apps.dart
provokateurin: The order is actually not nullable, because it is a oneOf (but I just randomly chose 100 as there is no value that naturally makes sense if the order were missing).

provokateurin:
Ok, I even found out the order is always an integer and the oneOf is wrong (though I remember I once had a broken response). I looked at the server code and only int is valid, but it is currently typed wrongly. I'll add a patch to drop the order oneOf and then this is even cleaner.

provokateurin:
I decided to not patch the specs for now as it would be a breaking change and will go away with https://github.com/nextcloud/server/pull/45455 (Nextcloud 30) while nothing is really broken here.

provokateurin:
Ok forget my comments, the order is actually sometimes and int and sometimes a string (but int encoded as string). I'll try to fix this upstream in https://github.com/nextcloud/server/pull/45455.

Leptopoda (approved):

PR #2069, opened 2024-05-22
feat(neon_notifications): Add buttons for notification actions

provokateurin:
Towards https://github.com/nextcloud/neon/issues/212


provokateurin:
Yeah I also realized testing the whole app is very easy, so I opened https://github.com/nextcloud/neon/issues/2068. I think I'll tackle that one soon so we don't need to add the test here.

PR #2070, opened 2024-05-22
fix(tool): search for bash in path

Leptopoda:
Not all linux distributions have bash installed at `/bin/bash`. Using the current path to search for the bash executable.
Another idea would be to directly use `/bin/sh` although I did not want to check whether all our scripts are compatible with plain posix shell (I doubt it).

I did succesfully test that this also works on OSX


provokateurin (approved):
Yeah not all scripts are POSIX compliant, so we can't use `sh` directly.

PR #2076, opened 2024-05-23
feat(neon_talk): Add mention autocompletion

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1919
Flutters Autocomplete widget is just utterly broken and also not really made for this use-case. Getting this to work with the flutter_typeahead package was pretty easy though.


packages/neon/neon_talk/test/message_input_test.dart
```diff
@@ -0,0 +1,162 @@
...
+    await tester.tap(find.text('😀'));
+    await tester.pumpAndSettle();
+
+    await expectLater(find.byType(TextField), matchesGoldenFile('goldens/message_input_emoji.png'));
+
+    await tester.testTextInput.receiveAction(TextInputAction.send);
+    verify(() => bloc.sendMessage('123😀456')).called(1);
+  });
+
+  testWidgets('Mention suggestions', (tester) async {
```
provokateurin, line 119: This is missing a golden image

packages/neon/neon_talk/lib/src/widgets/message_input.dart
```diff
@@ -0,0 +1,197 @@
...
+          if (skippedChars <= cursor && (skippedChars + part.length + 1) >= cursor) {
+            matchingPart = part;
+            break;
+          }
+          skippedChars += part.length + 1;
+        }
+
+        if (matchingPart == null || !matchingPart.startsWith('@') || matchingPart.length <= 1) {
+          return [];
+        }
```
Leptopoda, line 104: Quite a few iterations and plits
```suggestion
        String? matchingPart = controller.text.substring(0, cursor);
        final index = matchingPart.lastIndexOf(' ') + 1;
        matchingPart = matchingPart.substring(index);
        if (!matchingPart.startsWith('@') || matchingPart.length <= 1) {
          return [];
        }
```

In theory we can also do the condition without spliting the string again (character at index (if existing) is @ and the difference between the cursor and index is greater than 2 but that's hard to understand later)

Leptopoda (changes requested):

packages/neon/neon_talk/lib/src/widgets/message_input.dart
provokateurin: Thanks, this makes perfect sense. I always forget that there are so many useful methods in Dart :sweat:

Leptopoda (approved):

PR #2091, opened 2024-05-27
feat(neon_talk): Update messages from polled hidden messages

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2064


packages/neon/neon_talk/lib/src/utils/constants.dart
```diff
@@ -0,0 +1,11 @@
+/// `systemMessage` of chat messages that are hidden and only need to be processed internally.
+/// They modify the state of messages, but should not be displayed to the user.
+// Keep this in sync with https://github.com/nextcloud/spreed/blob/ada70487a306751accb280a475b1f123a56d4e74/src/store/conversationsStore.js#L689
+const hiddenMessages = {
```
Leptopoda, line 4: Please add a type annotation:
```dart
const Set<String> hidden ...
```
I've had occasions where the type inference was not doing the right thing so I consider it a best practice to add them to constants.

I'm open for other opinions

Leptopoda (approved):

PR #2098, opened 2024-05-29
fix(neon_talk): Ignore hidden messages for finding previous and last message

provokateurin:
These messages need to be ignored when finding the previous or last message since they won't be displayed in the end.


packages/neon/neon_talk/lib/src/blocs/room.dart
```diff
@@ -62,7 +62,14 @@ class _TalkRoomBloc extends InteractiveBloc implements TalkRoomBloc {
...
       }
 
-      final lastMessage = result.requireData.firstOrNull;
+      spreed.ChatMessageWithParent? lastMessage;
+      for (final message in result.requireData) {
+        if (!message.isHidden) {
+          lastMessage = message;
+          break;
+        }
+      }
```
Leptopoda, line 71: isn't this the same as 
```dart
final lastMessage = result.requireData.firstWhereOrNull((message) => !message.isHidden);
```

I find it easier to understand

packages/neon/neon_talk/lib/src/pages/room.dart
```diff
@@ -119,10 +120,17 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
...
 
                   spreed.ChatMessageWithParent? previousMessage;
-                  if (messagesResult.requireData.length > index + 1) {
-                    previousMessage = messagesResult.requireData[index + 1];
+                  for (var i = index + 1; i < messagesResult.requireData.length; i++) {
+                    final m = messagesResult.requireData[i];
+                    if (!m.isHidden) {
+                      previousMessage = m;
+                      break;
+                    }
```
Leptopoda, line 133: Again my preference would be:
```dart
final previousMessage =  messagesResult.requireData.skip(index).firstWhereOrNull((message) => !message.isHidden);
```

Leptopoda (commented):
I'm wondering why the UI has to skip hidden messages?
This should just be done in the bloc itself (the unwrap callback where the messages subject is populated).

provokateurin:
Ideally I'd like to add tests for the lastKnownMessageId now, but the value would only be used for testing and it should generally be avoided adding code that is only used for testing (and in this case the simple integer would have to be turned into a behaviorsubject which is totally unnecessary).

PR #2099, opened 2024-05-29
feat(neon_talk): Add date divider between messages of different days

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2081
Depends on https://github.com/nextcloud/neon/pull/2098


packages/neon/neon_talk/test/goldens/room_page_messages.png
provokateurin, line 1: This golden looks weird because the Chip isn't properly centered on the Divider, but when you test this manually you see that it is actually correctly centered. I don't know why this happens, but it must be some difference on how flutter renders in tests (maybe something with the font idk).

packages/neon/neon_talk/lib/src/pages/room.dart
```diff
@@ -121,14 +129,44 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
...
+                  Widget child = TalkMessage(
+                    chatMessage: message,
+                    lastCommonRead: lastCommonReadSnapshot.data,
+                    previousChatMessage: previousMessage,
+                  );
+
+                  if (previousMessage != null) {
+                    final date = parseTimestampDate(message.timestamp);
+                    final previousDate = parseTimestampDate(previousMessage.timestamp);
+                    if (!date.isAtSameMomentAs(previousDate)) {
```
Leptopoda, line 141: This looks a bit expensive to me as it creates multiple new DateTime objects per message.
I'd prefer a method (maybe extension method) that checks whether both are the same date.

```dart
extension Test on DateTime {
bool isSameDate(DateTime other){
// converting to utc is inexpesive as all times should already be in that timezone
final thisUtc = this.toUtc();
final otherUtc = other.toUtc();

// checking in reverse as it is more likely that the previous message has a different day than year
return thisUtc.day == otherUtc.day && thisUtc.month == otherUtc.month && thisUtc.year == otherUtc.year;
}
}
```

packages/neon/neon_talk/lib/src/pages/room.dart
```diff
@@ -121,14 +129,44 @@ class _TalkRoomPageState extends State<TalkRoomPage> {
                   final previousMessage =
                       messagesResult.requireData.length > index + 1 ? messagesResult.requireData[index + 1] : null;
 
+                  Widget child = TalkMessage(
+                    chatMessage: message,
+                    lastCommonRead: lastCommonReadSnapshot.data,
+                    previousChatMessage: previousMessage,
+                  );
+
+                  if (previousMessage != null) {
```
Leptopoda, line 138: imo the date information is also relevant for the first message in a chat. If there is no previous message just show the date.

packages/neon/neon_talk/lib/src/pages/room.dart
provokateurin: I have a different more primitive idea: We do an integer division of the timestamp by the number of seconds in a day and compare that value with the value for the previous date. That's a lot cheaper to compute and doesn't allocate any date objects (and also easier to understand IMO as it is a conversion and comparison in one step).

Leptopoda (approved):

PR #2100, opened 2024-05-29
feat(neon_framework): Support blurhashes in NeonImages

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1583

I could remove the dropping of the non-existent prop here and make a separate PR that we merge right before we make a new breaking release.

Talk doesn't add the blurhash metadata yet for images in chat messages, but I already filed an issue for that: https://github.com/nextcloud/spreed/issues/12351

Testing this on the dev container is not so easy because the metadata isn't populated due to the crons not working as intended. To trigger it manually for a file use `./occ metadata:get --refresh --reset --as-array`.


Leptopoda:
I have not tested this code yet but looking at the code I'm wondering if we should also show the blurhash when an error occurred?

provokateurin:
Hm yeah that might make sense, as we at least show something to the user and the user also knows that the image /hasn't/wasn't fully loaded (yet).

Leptopoda (approved):
LGTM

provokateurin:
I should probably still change that for errors the blurhash will be shown?

PR #2102, opened 2024-05-30
fix(neon_framework): Show no results message when unified search has no results at all

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/1865

### DIFF
--- a/packages/neon_framework/lib/src/blocs/unified_search.dart
+++ b/packages/neon_framework/lib/src/blocs/unified_search.dart
@@ -188,7 +188,7 @@ class _UnifiedSearchBloc extends InteractiveBloc implements UnifiedSearchBloc {
   }
 
   bool providerMatchesApp(String providerID, AppImplementation app) =>
-      providerID == app.id || providerID.startsWith('${app.id}_');
+      providerID == app.id || providerID.startsWith('${app.id}_') || providerID.startsWith('${app.id}-');
 
   bool hasEntries(Result<core.UnifiedSearchResult> result) => !result.hasData || result.requireData.entries.isNotEmpty;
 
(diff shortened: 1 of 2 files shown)


packages/neon_framework/lib/src/blocs/unified_search.dart
```diff
@@ -188,7 +188,7 @@ class _UnifiedSearchBloc extends InteractiveBloc implements UnifiedSearchBloc {
   }
 
   bool providerMatchesApp(String providerID, AppImplementation app) =>
-      providerID == app.id || providerID.startsWith('${app.id}_');
+      providerID == app.id || providerID.startsWith('${app.id}_') || providerID.startsWith('${app.id}-');
```
Leptopoda, line 193: Do you think a regex would be easier to read?

packages/neon_framework/lib/src/widgets/unified_search_results.dart
```diff
@@ -139,6 +133,14 @@ class NeonUnifiedSearchResults extends StatelessWidget {
     }
   }
 
+  Widget _buildNoResultsTile(BuildContext context) => AdaptiveListTile(
+        leading: Icon(
+          AdaptiveIcons.close,
+          size: largeIconSize,
+        ),
+        title: Text(NeonLocalizations.of(context).searchNoResults),
+      );
```
Leptopoda, line 142:
```suggestion
  Widget _buildNoResultsTile(BuildContext context) {
    return AdaptiveListTile(
        leading: Icon(
          AdaptiveIcons.close,
          size: largeIconSize,
        ),
        title: Text(NeonLocalizations.of(context).searchNoResults),
      );
  }
```

Leptopoda (approved):

packages/neon_framework/lib/src/blocs/unified_search.dart
provokateurin: I don't think so and regex is likely heavier to compute than these simple string interpolations and comparisons.

PR #2103, opened 2024-05-30
feat(app): Add wrapper to run as a Nextcloud server app

provokateurin:
Adding this makes the app package quite bloated with so many different files, so maybe it is better to put it in a separate directory somewhere (either inside the app package or outside of the packages dir). Alternatively I could just move this to an entirely different repo as there are no specific changes needed on the flutter side to make this experiment work.


Leptopoda:
I have no idea what to review here due to the lack of deeper php knowledge.
And I'm not a fan of mixing php and dart code in a single package.
What would be the benefit of having it in here over a separate repo (as you already mentioned)?

provokateurin:
The only benefit I'd see is simultaneous releases, but that isn't really a concern to me. I'm also preferring a separate repo, so I'll just do that.

provokateurin:
Now available at https://github.com/provokateurin/neon_web. @Leptopoda should we link to it somewhere?

PR #2113, opened 2024-06-02
feat(neon_framework): Automatically check accessibility guidelines in widget tests

provokateurin:
Towards https://github.com/nextcloud/neon/issues/300.
At least some automatic checks for missing tooltips and low contrasts.


packages/neon/neon_talk/lib/l10n/en.arb
```diff
@@ -12,5 +12,10 @@
       "type": {}
     }
   },
-  "roomSendMessage": "Send a message..."
+  "roomSendMessage": "Send a message...",
```
Leptopoda, line 15:
```suggestion
  "roomSendMessage": "Write a message...",
```
Let's get really nit picky :sweat_smile: 
Typing here does not send it yet. 
This also aligns it with the upstream string.

Leptopoda (approved):

PR #2120, opened 2024-06-03
Require Dart ^3.0 and Flutter ^3.22

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2119

### DIFF
--- a/packages/app/pubspec.yaml
+++ b/packages/app/pubspec.yaml
@@ -3,7 +3,7 @@ version: 1.0.0
 publish_to: 'none'
 
 environment:
-  sdk: '>=3.2.0 <4.0.0'
+  sdk: ^3.0.0
   flutter: '>=3.16.0'
 
 dependencies:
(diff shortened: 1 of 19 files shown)


provokateurin:
Not really, but I'd still like to align the requirements because the current state is just a mess.

Leptopoda:
Then let's rather upgrade to dart 3.4 directly as I'd love to play a bit with exension types :P

provokateurin:
Should we really do this? It will force all users of the nc package to upgrade as well (including the upgrade to the latest flutter version if they use it in a flutter project). I'd just wait until it is necessary to have Dart 3.4 for some feature we want to have or use.

provokateurin:
Agreed, so let's downgrade to ^3.0.0 instead?

Leptopoda:
if possible yes.
I don't think we rely on any 3.1 features

Leptopoda (approved):

PR #2122, opened 2024-06-04
chore(deps): Update openapi-extractor

provokateurin:


packages/nextcloud/lib/src/api/core.openapi.json
```diff
@@ -3560,12 +3560,6 @@
                                                                 {
                                                                     "$ref": "#/components/schemas/CommentsCapabilities"
                                                                 },
-                                                                {
-                                                                    "$ref": "#/components/schemas/CoreCapabilities"
-                                                                },
-                                                                {
-                                                                    "$ref": "#/components/schemas/CorePublicCapabilities"
-                                                                },
```
Leptopoda, line 3568: how is this change not reflected in the generated dart code 🤔

provokateurin: This was the problem with the duplicate entries you found a while back. We just didn't update the specs/openapi-extractor back then exactly because it doesn't change the generated code (the duplicate entries are right above the highlighted diff).

Leptopoda: Ah I remeber that they where just duplicates ...

Leptopoda (approved):

PR #2127, opened 2024-06-04
fix(neon_notifications): Expose Bloc using public interface

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2125
Using the public interface is how it should be done as the framework and other apps might want to access it too, but the AppImplementation was only advertising it as the internal interface.


Leptopoda (approved):

Leptopoda:
Can we somehow add tests to this?

provokateurin:
Not really, we could just make a test that confirms an instance of the app implementation has a the correct type, but I don't see much value in that because as long as the code doesn't change it is guaranteed to work.

PR #2142, opened 2024-06-08
feat(neon_talk): Make message input multiline

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2139


Leptopoda:
Once the message compose box reaches its maximum size adding another new line does not scroll the text (one would type outside the textfield).
Maybe some text alignment could help :woman_shrugging:

provokateurin:
Once you type a character it will scroll there, only newlines will be below the visible part for some reason. I also tried this on Android and there the newlines were shown properly, so I don't know what is going on there.

Leptopoda:
I looked a bit into this and I don't think there is an easy soultion (other than adding a custom scrollController)

Leptopoda (approved):

PR #2143, opened 2024-06-08
refactor(neon_talk): Make message input round

provokateurin:
Looks a lot nicer than the almost rectangular border.


Leptopoda (approved):
Would it make sense to have this in the talk theme?

provokateurin:
I don't know what other inputs we will have, but IMO only the message input looks good when it is round (I also tried adding it to the global theme but that was a bit ugly on the login screen and so on).

PR #2149, opened 2024-06-08
feat(neon_talk): Display last message time in room list

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1744
Not the sophisticated variant yet, but already much better than no time at all.
The ARB/ICU stuff is a bit ugly because there is no boolean type that could be used (at least I found nothing like that).


packages/neon/neon_talk/lib/src/pages/main.dart
```diff
@@ -95,20 +98,55 @@ class _TalkMainPageState extends State<TalkMainPage> {
...
+        trailing = Padding(
+          padding: const EdgeInsets.symmetric(vertical: 4),
+          child: TalkReadIndicator(
+            chatMessage: lastChatMessage,
+            lastCommonRead: room.lastCommonReadMessage,
+          ),
+        );
+      }
+
+      final timestamp = DateTimeUtils.fromSecondsSinceEpoch(
```
Leptopoda, line 116: Please parse to the local timezone directly

Leptopoda (approved):

provokateurin:
Also adjusted the tests to check against a non-UTC timezone.

PR #2150, opened 2024-06-08
feat(neon_talk): Add tooltip to display full message date

provokateurin:
The fix to display the time in the local timezone ofc doesn't change anything because of https://github.com/nextcloud/neon/issues/2147, but once that is fixed it will be correct.


provokateurin:
Also adjusted the tests to actually test against a non-UTC timezone.

PR #2162, opened 2024-06-11
perf(neon_framework): Store cached requests as BLOB

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2160

I didn't do any performance comparisons (maybe @Leptopoda you can do that on one of your aging devices?), but at least the image flickering seems to be less which would indicate the cache is faster for binary data.

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -28,11 +28,11 @@ typedef UnwrapCallback<T, R> = T Function(R);
 ///
 /// This is used to store a value in the cache.
 @internal
-typedef SerializeCallback<T> = String Function(T);
+typedef SerializeCallback<T> = Uint8List Function(T);
 
 /// A callback to revive cached values [String] into their original type [T].
 @internal
-typedef DeserializeCallback<T> = T Function(String);
+typedef DeserializeCallback<T> = T Function(Uint8List);
 
 /// How often a request will be tried.
 ///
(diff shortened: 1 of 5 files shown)


Leptopoda (commented):
I'll do a performance check shortly

provokateurin:
The dynamite_runtime commit is not breaking because it only widens the type.

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -92,13 +92,13 @@ class RequestManager {
         },
         unwrap: (rawResponse) => unwrap(rawResponse),
         serialize: (data) {
-          const encoder = RawResponseEncoder();
-
-          return json.encode(encoder.convert(data as DynamiteRawResponse));
+          return const RawResponseEncoder()
+              .fuse(const JsonEncoder())
+              .fuse(const Utf8Encoder())
```
Leptopoda, line 97: Any reason you did not use the mentioned `JsonUtf8Encoder`?

Leptopoda (approved):
Looks like the performance has improved quite a bit.

packages/neon_framework/lib/src/utils/request_manager.dart
provokateurin: Only to make it more similar to the deserialization part. If you want I can change it, but I don't know if it makes any difference in the end.

provokateurin: Ok from the class description it sounds like it needs one less copy of the data and thus should be more performant.

PR #2165, opened 2024-06-12
feat(nextcloud): add interceptors

Leptopoda:
I separated commits per package to make the changelogs cleaner.
I can do separate PRs if it helps for reviewability (although I think it's quite clean).


provokateurin (changes requested):
LGTM, but can you also migrate the nextcloud_test part that records the fixtures to these new interceptors?
Also we just use our own custom solution instead of https://pub.dev/packages/http_interceptor?

packages/nextcloud/lib/src/client.dart
```diff
@@ -49,6 +60,36 @@ class NextcloudClient extends DynamiteClient with http.BaseClient {
...
+    for (final interceptor in _interceptors) {
+      if (interceptor.shouldInterceptResponse()) {
+        interceptedResponse = await interceptor.interceptResponse(
+          response: interceptedResponse,
+        );
+      }
+    }
+
+    return interceptedResponse;
+  }
```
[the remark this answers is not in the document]
Leptopoda: It is indirectly tested by the cookie tests already.
I'll add a separate one :+1:

Leptopoda:
Marking as draft as this interceptor implementation does not allow any caching interceptor.
I'll have to look into this further

Leptopoda:
I'd like to merge this as is (after fixing the tests and adding separate ones for the interceptors).
@provokateurin  do you have anything against this? We'll probably have to find a different solution for the cache but you are already working on that.

packages/nextcloud/test/client_test.dart
```diff
@@ -29,6 +29,7 @@ void main() {
         final client = NextcloudClient(
           uri,
           httpClient: mockedClient,
+          // ignore: deprecated_member_use_from_same_package
           cookieJar: cookieJar,
```
provokateurin, line 33: Why not use the interceptor for tests?

packages/nextcloud/test/client_test.dart
```diff
@@ -56,6 +57,7 @@ void main() {
         final client = NextcloudClient(
           uri,
           httpClient: mockedClient,
+          // ignore: deprecated_member_use_from_same_package
```
provokateurin, line 60: Same here

provokateurin (changes requested):
LGTM, but maybe you could also migrate the WebDavCsrfClient to be an interceptor? Can be done as a follow-up though.

packages/nextcloud/lib/src/interceptors/http_interceptor.dart
```diff
@@ -0,0 +1,25 @@
...
+  ///
+  /// Provided requests are not finalized yet. It is an error for an interceptor
+  /// to request itself.
+  FutureOr<S> interceptRequest({required S request});
+
+  /// Whether this interceptor should intercept response.
+  bool shouldInterceptResponse() => true;
+
+  /// Intercepts the given [response].
+  FutureOr<T> interceptResponse({required T response});
```
provokateurin, line 24: But this can be used for caching, no?

provokateurin: Ah no I was wrong...

Leptopoda: Not really.
If a cache interceptor has a stored value there is currently no way to bypass the actual http request being made and return the cached one.

provokateurin: Yeah but we can easly add two more methods to intercept the actual request and return a response. I'll do that later once it is necessary.

packages/nextcloud/lib/src/interceptors/http_interceptor.dart
```diff
@@ -0,0 +1,25 @@
...
+  bool shouldInterceptRequest() => true;
+
+  /// Intercepts the given [request].
+  ///
+  /// Provided requests are not finalized yet. It is an error for an interceptor
+  /// to request itself.
+  FutureOr<S> interceptRequest({required S request});
+
+  /// Whether this interceptor should intercept response.
+  bool shouldInterceptResponse() => true;
```
provokateurin, line 21: I'm not sure about the defaults, I think it's better to no provide one and always let the implementer define it explicitly.

packages/nextcloud/test/client_test.dart
Leptopoda: I'll add tests for the CookieInterceptor but would like to keep the existing test until we remove the deprecated member.

provokateurin:
The CI failure looks like the cookies are not applied

provokateurin:
@Leptopoda close this in favor of https://github.com/nextcloud/neon/pull/2288?

PR #2179, opened 2024-06-16
feat(neon_talk): Enable message deletion

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2079
Depends on https://github.com/nextcloud/neon/pull/2173

Ready for review, just needs the other PR merged first.


packages/neon/neon_talk/lib/src/blocs/room.dart
```diff
@@ -321,6 +324,26 @@ class _TalkRoomBloc extends InteractiveBloc implements TalkRoomBloc {
...
+          token: token,
+          messageId: chatMessage.id,
+        );
+
+        updateLastCommonRead(response.headers.xChatLastCommonRead);
+
+        final m = response.body.ocs.data;
+        updateLastKnownMessageId(m.id);
+
+        prependMessages([m]);
```
provokateurin, line 341: This is bugged and the test data was unfortunately chosen in a way that hides it.

provokateurin: Ok I was stupid and this was right, just the mock implementation of the API was wrong. I was already wondering why this API would return the deleted message and not the system message for it.

Leptopoda (approved):

PR #2182, opened 2024-06-18
chore(nextcloud)!: Remove support for server 27.x and all apps only compatible up to 27.x

provokateurin:
The last 27.x release is set to happen on Thursday, so after that all 27.x versions are EOL. We can finally remove everything related to those versions. This support drop is a bit special because 27.0 and 27.1 are getting EOL at the same time due to the later only being a minor release.
In the end this actually halves the number of tests running in the nextcloud package :rocket:


provokateurin:
Meh I forgot to adjust the bloc tests.

Leptopoda (approved):

PR #2196, opened 2024-06-24
fix(neon_talk): Respect room read-only and participant permissions when showing reply/react message options

provokateurin:


provokateurin:
I messed up the order and accidentally already merged https://github.com/nextcloud/neon/pull/2179 into here, so please ignore those changes.

Leptopoda (approved):

PR #2203, opened 2024-06-25
refactor(nextcloud,neon_framework)!: Move all json_serializable usage to built_value

provokateurin:
Closes https://github.com/nextcloud/neon/issues/967

### DIFF
--- a/packages/neon_framework/lib/src/models/account.dart
+++ b/packages/neon_framework/lib/src/models/account.dart
@@ -1,15 +1,22 @@
 import 'dart:convert';
 
+import 'package:built_value/built_value.dart';
+import 'package:built_value/serializer.dart';
+import 'package:built_value/standard_json_plugin.dart';
 import 'package:cookie_store/cookie_store.dart';
 import 'package:crypto/crypto.dart';
 import 'package:http/http.dart';
-import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
 import 'package:neon_framework/src/utils/findable.dart';
 import 'package:neon_framework/storage.dart';
 import 'package:nextcloud/nextcloud.dart';
 
 part 'account.g.dart';
+
+@SerializersFor([
+  Account,
+])
+final Serializers _serializers = (_$_serializers.toBuilder()..addPlugin(StandardJsonPlugin())).build();
 
 /// Credentials interface
 @internal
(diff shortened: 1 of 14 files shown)


packages/neon_framework/lib/src/models/account.dart
```diff
@@ -26,78 +34,63 @@ abstract interface class Credentials {
...
 
-    client = NextcloudClient(
-      serverURL,
-      loginName: username,
-      password: password,
-      appPassword: password,
-      userAgent: userAgent,
-      cookieJar: cookieStore != null ? CookieJarAdapter(cookieStore) : null,
-      httpClient: httpClient,
-    );
```
provokateurin, line 54: With these changes the client is no longer cached (and all the other fields as well). Can we somehow keep them as fields, but completely ignore them in built_value?

provokateurin: I guess one possible way would be using Expando to cache it outside of the class, but maybe this isn't necessary after all since I don't see something that is super expensive about this behavior.

provokateurin: After thinking more about this we should definitely do this. Otherwise all the Expandos used inside the client itself are useless and execute a lot of unnecessary memory operations.
@Leptopoda let me know if there is any cleaner way than using Expando, otherwise I will just go with it.

packages/neon_framework/lib/src/models/account.dart
```diff
@@ -26,78 +33,66 @@ abstract interface class Credentials {
...
+  }
 
   /// The unique ID of the account.
   ///
   /// Implemented in a primitive way hashing the [username] and [serverURL].
   @override
-  @JsonKey(includeFromJson: false, includeToJson: false)
-  final String id;
+  @BuiltValueField(serialize: false)
+  String get id => sha1.convert(utf8.encode('$username@$serverURL')).toString();
```
Leptopoda, line 91: This is sorta expensive and we should cache the computation by annotating it with `@memoized`

packages/neon_framework/lib/src/models/account.dart
```diff
@@ -26,78 +33,66 @@ abstract interface class Credentials {
...
 
   /// A human readable representation of [username] and [serverURL].
-  @JsonKey(includeFromJson: false, includeToJson: false)
-  final String humanReadableID;
-
-  /// Builds a human readable id for a user and server pair.
-  static String _buildHumanReadableID(String username, Uri serverURL) {
+  @BuiltValueField(serialize: false)
+  String get humanReadableID {
```
Leptopoda, line 95: Same again
This is sorta expensive and we should cache the computation by annotating it with `@memoized`

packages/neon_framework/lib/src/models/account.dart
```diff
@@ -26,78 +33,66 @@ abstract interface class Credentials {
...
+  /// Custom HTTP client used for all requests.
+  @visibleForTesting
+  @BuiltValueField(serialize: false)
+  Client? get httpClient;
 
   /// An authenticated API client.
-  @JsonKey(includeFromJson: false, includeToJson: false)
-  late final NextcloudClient client;
+  @BuiltValueField(serialize: false)
+  NextcloudClient get client {
```
Leptopoda, line 69: I'm not 100% sure about this one but it was cached before so maybe we should do that again.
So another `@memoized`

Leptopoda (changes requested):

packages/neon_framework/lib/src/models/account.dart
Leptopoda: I can't really follow.
Can you explain it in another way?

provokateurin: Oh I didn't know about the memoized annotation, but this solves the caching problem I was going to fix by using Expando.

packages/neon_framework/lib/src/models/account.dart
```diff
@@ -26,83 +33,71 @@ abstract interface class Credentials {
...
+  String get id => sha1.convert(utf8.encode('$username@$serverURL')).toString();
 
   /// A human readable representation of [username] and [serverURL].
-  @JsonKey(includeFromJson: false, includeToJson: false)
-  final String humanReadableID;
-
-  /// Builds a human readable id for a user and server pair.
-  static String _buildHumanReadableID(String username, Uri serverURL) {
+  @memoized
+  @BuiltValueField(serialize: false)
```
Leptopoda, line 99: I don't think these annotations are needed as a getter with a body is never serializable.

Can you please confirm that and remove them if they're not needed?

Leptopoda (approved):

packages/neon_framework/lib/src/models/account.dart
provokateurin: You're right.

PR #2226, opened 2024-07-02
Move all immutable types to built_collection

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2220


Leptopoda:
What is the benefit of this?
Most instances have already been const so mutating them was impossible.
I also do not see the benefit of making the serializer types built_collection`s.

PR #2227, opened 2024-07-02
docs: Add development instance setup

satvik2131:


provokateurin (approved):
Thanks a lot!

provokateurin (changes requested):
Ah it should mention the URLs of the instance (`http://localhost` on the host machine and `http://10.0.2.2` for android emulator).

satvik2131:
> Ah it should mention the URLs of the instance (`http://localhost` on the host machine and `http://10.0.2.2` for android emulator).

sure , will do that , thanks for the feedback :)

PR #2228, opened 2024-07-02
fix(dynamite_runtime): use content type header to deserialize response

Leptopoda:
Adding back the http_parser dependency :)

### DIFF
--- a/packages/dynamite/dynamite_runtime/lib/src/client/response.dart
+++ b/packages/dynamite/dynamite_runtime/lib/src/client/response.dart
@@ -1,9 +1,10 @@
 import 'dart:convert';
-import 'dart:typed_data';
 
 import 'package:built_value/serializer.dart';
 import 'package:dynamite_runtime/http_client.dart';
 import 'package:http/http.dart' as http;
+import 'package:http_parser/http_parser.dart';
+import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 
 /// Response returned by operations of a `DynamiteClient`.
@@ -70,6 +71,8 @@ final class ResponseConverter<B, H> with Converter<http.Response, DynamiteRespon
   /// The serializer to convert the raw header and body into [B] and [H].
   final DynamiteSerializer<B, H> serializer;
 
+  static final Logger _logger = Logger('ResponseConverter');
+
   @override
   DynamiteResponse<B, H> convert(http.Response input) {
     final rawHeaders = input.headers;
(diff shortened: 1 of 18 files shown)


packages/dynamite/dynamite_runtime/lib/src/client/response.dart
```diff
@@ -122,10 +123,24 @@ final class ResponseConverter<B, H> with Converter<http.StreamedResponse, Future
...
+    final rawBody = switch (mediaType) {
+      MediaType(type: 'text') || MediaType(type: 'application', subtype: 'javascript') => response.body,
+      MediaType(type: 'application', subtype: 'json') => json.decode(response.body),
+      MediaType(type: '*', subtype: '*') ||
+      MediaType(type: 'application', subtype: 'octet-stream') ||
+      MediaType(type: 'image') ||
+      _ =>
+        response.bodyBytes,
     };
     final body = _deserialize<B>(rawBody, serializer.serializers, serializer.bodyType);
```
provokateurin, line 145: Do we even need to deserialize in case the string/bytes body is used? For those cases the bodyType should already be Uint8List or String :thinking:

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
```diff
@@ -122,10 +123,24 @@ final class ResponseConverter<B, H> with Converter<http.StreamedResponse, Future
...
+        log('Could not parse $contentType', error: error, stackTrace: stackTrace);
+      }
+    }
+
+    final rawBody = switch (mediaType) {
+      MediaType(type: 'text') || MediaType(type: 'application', subtype: 'javascript') => response.body,
+      MediaType(type: 'application', subtype: 'json') => json.decode(response.body),
+      MediaType(type: '*', subtype: '*') ||
+      MediaType(type: 'application', subtype: 'octet-stream') ||
+      MediaType(type: 'image') ||
```
provokateurin, line 141: Just using the wildcard makes it more readable and has the same effect.

Leptopoda:
With #2249 this should pass now (once the CI is a live again)

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
```diff
@@ -1,9 +1,10 @@
 import 'dart:convert';
-import 'dart:typed_data';
+import 'dart:developer';
```
provokateurin, line 2: Probably unintended change?

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
Leptopoda: I'm not 100% sure with the ContentString but we should be able to do so

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
Leptopoda: I need this for logging.
But we can just pull in package:logging into nextcloud as well

provokateurin: We're already doing so

Leptopoda: my bad this is about the dynamite_runtime package. Should we pull it in?

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
Leptopoda: All tests still pass when changing this :)

packages/dynamite/dynamite_runtime/lib/src/client/response.dart
provokateurin: Yes let's do that

provokateurin (approved):

Leptopoda:
I had to adjust most bloc tests as they rely on both `dynamite_runtime` and `nextcloud` so please give it a new review.
I think we have to do something about that eventually ...

provokateurin (approved):

PR #2229, opened 2024-07-02
feat(nextcloud): add cookbook support

Leptopoda:
towards: #353 
long term we should find a way to support both DateTime and Duration parsing


tool/generate-specs.sh
```diff
@@ -50,6 +50,12 @@ function generate_spec() {
   cd external/nextcloud-drop_account
   generate_spec "." "drop_account"
 )
+(
+  yq -s '.[0]."components"."schemas" = .[1] | .[0] | walk(if type == "object" then with_entries( if (.key == "$ref" and (.value | test("#\/components\/schemas\/") | not)  ) then .value |= sub("#\/"; "#/components/schemas/") else . end ) else . end) | .paths |= with_entries(.key = "/index.php/apps/cookbook\(.key)") ' \
+    external/cookbook/docs/dev/api/0.1.2/openapi-cookbook.yaml \
+    external/cookbook/docs/dev/api/0.1.2/objects.yaml \
+    > /tmp/nextcloud-neon/cookbook.openapi.json
+)
```
provokateurin, line 58: Until now we said every new spec can only be added if it was generated by openapi-extractor. For notes and news we still have handwritten once, but they are legacy and work fine atm. If we add this we might need to rethink what our stance on this is. For me personally this is fine, since the spec seems well maintained and we can reason that the quality is good and any fixes will be accepted upstream.

packages/nextcloud/test/cookbook_test.dart
```diff
@@ -0,0 +1,245 @@
...
+          print(await container.allLogs());
+        }
+        container.destroy();
+      });
+
+      group(cookbook.$CategoriesClient, () {
+        // Get all known categories
+        //
+        //Future<BuiltList<Category>> listCategories() async
+        test('test listCategories', () async {
```
provokateurin: The tests are really not like the other unit tests. Why does it have an extra "title", the signature of the method that is tested and then "test" as part of the description which is really unnecessary?

provokateurin: Ah I see in the next commit this is changed. I think you accidentally added part of the tests to the wrong commit. In the setup step you also try to upload non-existent recipes for testing.

packages/nextcloud/test/cookbook_test.dart
```diff
@@ -16,64 +15,31 @@ void main() {
...
-        resetFixture();
       });
 
-      tearDownAll(() async {
+      tearDown(() async {
         if (Invoker.current!.liveTest.errors.isNotEmpty) {
           print(await container.allLogs());
         }
         container.destroy();
       });
```
provokateurin, line 28: It can not reuse the same container for all tests? If you'd just delete all the recipe files after the test is done, shouldn't that reset everything to the initial state?

provokateurin (changes requested):
The last two commit should be squash since you just add tests for the feature you introduced at the same time. Also the test descriptions have this weird `test ` prefix which is entirely unnecessary.

provokateurin:
Please also add version checks to ensure no known broken version is accepted.

tool/generate-specs.sh
Leptopoda: As long as the specs are high enough quality and cover 100% of the api we should allow them.
AFIK Christian wants to move to the openapi-extractor eventually but as always that'll take some time :)

packages/nextcloud/test/cookbook_test.dart
Leptopoda: the issue is that I want a fresh predefined set of recipes.
The collection introduced has a lot of nasty "bugs" in them like integer IDs and I want to make sure we can parse all of them. Sometimes this is just testing the upstream php code but I've had quite a few bugs already that make me uncomfortable without these tests.

We could upload all recipes to a different folder and use webdav to recreate the original set :woman_shrugging: 
Is it really such a big performance difference?

provokateurin: It depends on the ratio of container startup time vs. test time.
I think you could at least bake the recipes into the docker image if you want to create a new one every time to save some startup time of the test.

packages/nextcloud_test/docker/Dockerfile
```diff
@@ -21,13 +21,18 @@ RUN (sh /entrypoint.sh php -S 0.0.0.0:80 &) && \
     for user in admin user1 user2; do curl -u "$user:$user" -H "ocs-apirequest: true" -s -o /dev/null http://localhost/ocs/v2.php/cloud/user; done
 
 COPY static /usr/src/nextcloud/static
+COPY assets/Recipes /usr/src/nextcloud/data/user1/files/Recipes
```
provokateurin, line 24: Not necessary for the unit tests, but for general testing it would be nice if all users had the recipes (so admin and user2 as well).

provokateurin (changes requested):
CI is still red :/
What about the spec patches, are they going to be upstreamed?

provokateurin:
The submodule should be named nextcloud-cookbook btw

packages/nextcloud_test/docker/Dockerfile
```diff
@@ -50,6 +58,10 @@ RUN curl -L "$DROP_ACCOUNT_URL" | tar -xz -C /
 
 FROM nextcloud
 
+COPY --from=cookbook /cookbook /usr/src/nextcloud/apps/cookbook
+RUN ./occ app:enable cookbook
+RUN ./occ files:scan --all
```
provokateurin, line 63: The files scan should be done after copying the files and not here

packages/nextcloud/lib/src/helpers/cookbook.dart
```diff
@@ -0,0 +1,37 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:nextcloud/src/api/cookbook.openapi.dart' as cookbook;
+import 'package:nextcloud/src/helpers/common.dart';
+import 'package:version/version.dart';
+
+/// Minimum API version of the cookbook app supported.
```
provokateurin, line 6: This is the app and not API version, right?

packages/nextcloud/test/cookbook_test.dart
provokateurin, line 1: The test names still contain the unnecessary `test ` prefix.

provokateurin (changes requested):
>  What about the spec patches, are they going to be upstreamed?

Leptopoda:
> What about the spec patches, are they going to be upstreamed?

I'll open a PR upstream once I have an initial cookbook implementation. I'm sure I'll find more nitpicks with the spec and I'd like to contribute everything in one go.

packages/nextcloud/test/cookbook_test.dart
```diff
@@ -0,0 +1,193 @@
...
+          expect(cookbookVersion[0], JsonObject(preset.version.major));
+          expect(cookbookVersion[1], JsonObject(preset.version.minor));
+        });
+      });
+
+      group('recipes', () {
+        test('callImport', () async {
+          final url = cookbook.UrlBuilder()
+            ..url = 'https://www.chefkoch.de/rezepte/126221054029392/Germknoedel-mit-Powidl.html';
+          final response = await client.cookbook.recipes.$import($body: url.build());
```
provokateurin, line 103: I don't like that it depends on an external service that could go down or change. Can you download the whole page and put it into the static folder?

provokateurin (commented):
LGTM

packages/nextcloud/test/cookbook_test.dart
Leptopoda: I understand the issue but in that case I'd rather create a custom site.

I don't think we can easily add their work into a gpl licenced codebase.

provokateurin: Right, that's a good point. Then we simply have to hope it never breaks.

provokateurin (approved):

provokateurin (changes requested):
Please add cookbook information to the README

provokateurin (approved):

PR #2231, opened 2024-07-02
refactor(dynamite_runtime)!: Make ResponseConverter sync

provokateurin:
Simplifies the usage of the converter when fusing multiple together. Otherwise the next converter needs to accept `Future<DynamiteResponse>` as an input.
Also moves reading the body from the response outside of the converter which clarifies the responsibilities and makes stacktraces more readable.


provokateurin:
Meh I can't reproduce the formatting problem locally...

Leptopoda (approved):

PR #2233, opened 2024-07-03
refactor(neon_framework): Only use Converter in RequestManager and store raw responses in RequestCache

provokateurin:
Towards https://github.com/nextcloud/neon/issues/2155

Unfortunately there was no way to split the changes to RequestCache and to RequestManager into separate commits because they rely on each other.
This should also give us some performance boost for reading and writing cache.
Next steps would be to eliminate the separate methods in RequestManager and only pass in the converter.

### DIFF
--- a/packages/neon_framework/lib/src/storage/request_cache.dart
+++ b/packages/neon_framework/lib/src/storage/request_cache.dart
@@ -1,31 +1,30 @@
 import 'dart:async';
+import 'dart:convert';
 import 'dart:typed_data';
 
+import 'package:http/http.dart' as http;
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:neon_framework/models.dart';
 import 'package:neon_framework/src/platform/platform.dart';
-import 'package:neon_framework/src/utils/request_manager.dart';
-import 'package:nextcloud/utils.dart';
 import 'package:path/path.dart' as p;
 import 'package:path_provider/path_provider.dart';
 import 'package:sqflite_common_ffi/sqflite_ffi.dart';
-import 'package:timezone/timezone.dart' as tz;
 
 final _log = Logger('RequestCache');
 
-/// A storage used to cache a key value pair.
+/// A storage used to cache HTTP requests.
 abstract interface class RequestCache {
-  /// Get's the cached value and parameters for the given [key].
-  Future<({Uint8List value, CacheParameters? parameters})?> get(Account account, String key);
+  /// Gets the cached status code, body and headers for the given [key].
+  Future<http.Response?> get(Account account, String key);
 
-  /// Set's the cached [value] at the given [key].
+  /// Sets the cached [response] at the given [key].
   ///
-  /// If a value is already present it will be updated with the new one.
-  Future<void> set(Account account, String key, Uint8List value, CacheParameters? parameters);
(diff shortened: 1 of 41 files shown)


packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -1,31 +1,34 @@
...
+        Uint8List body,
+        Map<String, String> headers,
+      })?> get(Account account, String key);
+
+  /// Set's the cached [body] at the given [key].
   ///
-  /// If a value is already present it will be updated with the new one.
-  Future<void> set(Account account, String key, Uint8List value, CacheParameters? parameters);
+  /// If a body is already present it will be updated with the new one.
+  Future<void> set(Account account, String key, int statusCode, Uint8List body, Map<String, String> headers);
```
Leptopoda, line 28: What about passing a http.BaseResponse parameter directly?

provokateurin: It needs to be http.Request, but yeah I didn't think about that.

packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -131,25 +126,17 @@ WHERE account = ? AND key = ?
...
-              tz.UTC,
-              expires,
-            )
-          : null,
+    return http.Response.bytes(
+      row['body']! as Uint8List,
+      row['statusCode']! as int,
+      headers: (json.decode(row['headers']! as String) as Map<String, dynamic>).map(
+        (key, value) => MapEntry(key, value as String),
+      ),
```
Leptopoda, line 134: wouldn't a `.cast<String,String>()` also do the trick?

packages/neon_framework/lib/src/storage/request_cache.dart
```diff
@@ -159,53 +146,52 @@ WHERE account = ? AND key = ?
           {
             'account': account.id,
             'key': key,
-            'value': value,
-            'etag': parameters?.etag,
-            'expires': parameters?.expires?.secondsSinceEpoch,
+            'statusCode': response.statusCode,
+            'body': response.bodyBytes,
+            'headers': json.encode(response.headers),
```
Leptopoda, line 151: we are encoding the headers two times.
Once here and once for the insert statement.

Please reduce this by storing it in a variable.

packages/neon_framework/lib/src/widgets/image.dart
```diff
@@ -235,10 +236,10 @@ class _NeonApiImageState extends State<NeonApiImage> {
...
       account: widget.account,
       cacheKey: widget.cacheKey,
-      getCacheParameters: () async => CacheParameters(
-        etag: widget.etag,
-        expires: widget.expires,
-      ),
+      getCacheHeaders: () async => {
+        if (widget.etag != null) 'etag': widget.etag!,
+        if (widget.expires != null) 'expires': formatHttpDate(widget.expires!),
+      },
```
Leptopoda, line 242: I really hate lambdas pointing to a map as that can easily be read as a method body.
Please change it to a method body returning a map.

Leptopoda (approved):

Leptopoda (changes requested):
Sorry wrong button

Leptopoda (approved):

PR #2237, opened 2024-07-03
refactor(neon_framework): Compute RequestManager cache keys automatically

provokateurin:
https://github.com/nextcloud/neon/issues/2155

I'm not sure myself if hashing the request is the best way, as it makes it a lot harder to understand the data that is stored in the cache. OTOH storing all those request parameters in the cache is tedious and we rarely or even never need to inspect the cached data, so it feels a bit pointless.

### DIFF
--- a/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
+++ b/packages/neon/neon_dashboard/lib/src/blocs/dashboard.dart
@@ -69,7 +69,6 @@ class _DashboardBloc extends InteractiveBloc implements DashboardBloc {
     if (!widgets.hasValue || !widgets.value.hasSuccessfulData) {
       await RequestManager.instance.wrap(
         account: account,
-        cacheKey: 'dashboard-widgets',
         subject: widgets,
         getRequest: account.client.dashboard.dashboardApi.$getWidgets_Request,
         converter: ResponseConverter(account.client.dashboard.dashboardApi.$getWidgets_Serializer()),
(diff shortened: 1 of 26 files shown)


Leptopoda:
>  The primary cache key consists of the request method and target URI.
   However, since HTTP caches in common use today are typically limited
   to caching responses to GET, many caches simply decline other methods
   and use only the URI as the primary cache key.
   
   https://www.rfc-editor.org/rfc/rfc7234 :thinking:

provokateurin:
In our case that is not enough once we have https://github.com/nextcloud/neon/pull/2129 as then all parameters go into the body. I think we need to keep it generic in this way to keep it working with all the quirk Nextcloud APIs.

Leptopoda:
I'd still like to restrict the cache to some methods. Caching a `HEAD` request sounds like a bad idea.
Do you think it makes sense to hash this on a separate isolate?

provokateurin:
Sure we can restrict the caching to GET methods, but I wouldn't be surprised if we need to revert it later.
As you said in https://github.com/nextcloud/neon/issues/2155 it would be nice to make the RequestCache super generic as well, so I was thinking of removing the hashing and just storing the request method, url and body directly in the cache, just like we do it for the response parts. This also makes the cache inspectable for debugging.

Leptopoda:
> I was thinking of removing the hashing and just storing the request method, url and body directly in the cache, just like we do it for the response parts. This also makes the cache inspectable for debugging.

I was also thinking about this. Maybe let's keep it a generic request for the interface and the persistence implementation can still decide to hash it if needed.

provokateurin:
Already a good example why only caching GET requests won't do it: PROPFIND used by WebDAV needs to be cached as well.

Leptopoda (approved):

PR #2249, opened 2024-07-08
fix(nextcloud): notes api returns empty array when deleting a note

Leptopoda:
I hope that I did not misread the php code but it looks like it always returns an empty array.
https://github.com/nextcloud/notes/blob/b8d90049ef51a2b80dbb10e756c01fdd139c6135/lib/Controller/NotesController.php#L301-L309


provokateurin (approved):
Yep you read the code correctly, I guess I was a bit sloppy when I wrote the spec :+1:

PR #2277, opened 2024-07-16
feat(neon_talk): Enable editing of chat messages

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2071


packages/neon/neon_talk/lib/l10n/en.arb
```diff
@@ -18,6 +18,19 @@
...
+  "@roomMessageLastEdited": {
+    "placeholders": {
+      "name": {
+        "type": "String"
+      },
+      "time": {
+        "type": "String"
+      }
+    }
+  },
```
Leptopoda, line 33: I'm not sure if you are aware of this, but `intl` can also format dates.

```suggestion
  "roomMessageEdit": "Edit",
  "roomMessageEdited": "edited",
  "roomMessageLastEdited": "Last edited by {name} at {time}",
  "@roomMessageLastEdited": {
    "placeholders": {
      "name": {
        "type": "String"
      },
      "time": {
        "type": "DateTime",
        "format": "yMd"
      }
    }
  },
```

provokateurin: In this case I need `yMd` and `jm`, but there is no constructor for that. Otherwise this would have been a nice option.

packages/neon/neon_talk/lib/src/widgets/message.dart
```diff
@@ -465,19 +465,45 @@ class _TalkCommentMessageState extends State<TalkCommentMessage> {
...
+            label = Row(
+              children: [
+                label,
+                Tooltip(
+                  message: TalkLocalizations.of(context).roomMessageLastEdited(
+                    widget.chatMessage.lastEditActorDisplayName!,
+                    DateFormat.yMd().add_jm().format(timestamp),
+                  ),
+                  child: Text(
+                    ' (${TalkLocalizations.of(context).roomMessageEdited})',
```
Leptopoda, line 497: why isn't the `( )` part of the localized string?

Leptopoda (commented):
changes look good overall :)

packages/neon/neon_talk/lib/src/widgets/message.dart
provokateurin: Because they will always be present, so they don't need to be translated?

packages/neon/neon_talk/lib/src/widgets/message_input.dart
```diff
@@ -258,6 +279,35 @@ class _TalkMessageInputState extends State<TalkMessageInput> {
       leading: icon,
     );
   }
+
+  Widget buildContextMessage({
+    required spreed.$ChatMessageInterface chatMessage,
+    required VoidCallback onDismiss,
+  }) {
+    return Padding(
```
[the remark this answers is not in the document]
provokateurin: It has no logic and is only used twice in the same context. I prefer this for now, but if it would be used somewhere else it makes sense.

Leptopoda (approved):

PR #2283, opened 2024-07-17
Fix semanticsLabel for Nextcloud client logo

Apocalypse96:
# Issue -> https://github.com/nextcloud/neon/issues/2273
## Fix Incorrect Semantics Label for Nextcloud Logo

**Issue:** The `semanticsLabel` for the Nextcloud logo was incorrectly transcribed as "Nextcloud logo". This needed to be updated for better accessibility and accuracy.

**Changes Made:**

- Updated the `semanticsLabel` in the `app_implementation.dart` file to provide a more accurate description.

### DIFF
--- a/packages/neon_framework/lib/src/models/app_implementation.dart
+++ b/packages/neon_framework/lib/src/models/app_implementation.dart
@@ -171,7 +171,7 @@ abstract class AppImplementation<T extends Bloc, R extends AppImplementationOpti
               'assets/app.svg.vec',
               packageName: 'neon_$id',
             ),
-            semanticsLabel: NeonLocalizations.of(context).nextcloudLogo,
+            semanticsLabel: "Nextcloud Client Logo",
           );
         },
       );


packages/neon_framework/lib/src/models/app_implementation.dart
```diff
@@ -171,7 +171,7 @@ abstract class AppImplementation<T extends Bloc, R extends AppImplementationOpti
               'assets/app.svg.vec',
               packageName: 'neon_$id',
             ),
-            semanticsLabel: NeonLocalizations.of(context).nextcloudLogo,
+            semanticsLabel: "Nextcloud Client Logo",
```
Leptopoda, line 174: Thanks for working on this issue but this is not what we need.
The problem is that these pictures contain the logo of the application in question. 
For example this is the files or news icon you can see in the drawer.

I think this should rather use `name(context)`.

Leptopoda (changes requested):

Leptopoda:
And as with your previous PR you'll have to 
- signoff the commit
- use a conventional commit message (something like "fix(neon_framework): semanticsLabel for drawer icon"
- add a unit test to avoid future regressions

provokateurin:
Closing for now, please re-open if you plan to follow up to the review comments.

PR #2287, opened 2024-07-18
feat(neon_talk): Add helpers for parsing timestamps

provokateurin:


Leptopoda:
I'm not sure if this should be in the nextcloud package.
Is there anything against having it in neon_talk?

provokateurin:
Not really, but the helpers can be useful for other people as well, so I don't see why they shouldn't be in the nextcloud package.

Leptopoda:
I'd rather not have them in the nextcloud package, as these fields offer a quite different experience to the rest of the BuiltValue models.

These extensions are not cached, so multiple requests will lead to parsing the time multiple times.
It might also be confusing that to change the time one has to change the string representation. 
Yes a quick look at the code will make that clear but I only see a point of cocnfusion.

Leptopoda (approved):

PR #2288, opened 2024-07-18
feat(neon_http_client): A http client for use by the neon framework

Leptopoda:
Replaces: #2165
This client will later also handle the functionality of the RequestManager like:
- cache
- retry requests
- timeout requests


packages/neon_http_client/lib/src/client.dart
```diff
@@ -0,0 +1,113 @@
...
+/// An exception caused by a [HttpInterceptor].
+final class InterceptionFailure extends NeonClientException {
+  /// Creates a new interceptor failure exception.
+  InterceptionFailure(super.message, [super.uri]);
+}
+
+/// A client configuring the clients for all Nextcloud APIs.
+///
+/// To access the APIs of a particular app import the extensions through `package:nextcloud/{id}.dart`.
+final class NeonClient with http.BaseClient {
```
provokateurin, line 21:
```suggestion
final class NeonHttpClient with http.BaseClient {
```
Otherwise the name is a bit ambigous

packages/nextcloud/lib/src/client.dart
```diff
@@ -16,38 +16,55 @@ class NextcloudClient extends DynamiteClient with http.BaseClient {
...
+      client = CookieJarClient(
+        httpClient: httpClient,
+        cookieJar: cookieJar,
+        baseHeaders: {
+          if (userAgent != null) HttpHeaders.userAgentHeader: userAgent,
+        },
+      );
+    } else {
+      client = httpClient ?? http.Client();
+    }
```
provokateurin, line 42:
```suggestion
    var client = httpClient ?? http.Client();
    if (cookieJar != null || userAgent != null) {
      client = CookieJarClient(
        httpClient: client,
        cookieJar: cookieJar,
        baseHeaders: {
          if (userAgent != null) HttpHeaders.userAgentHeader: userAgent,
        },
      );
    }
```
Easier to understand the logic this way

packages/neon_http_client/lib/src/client.dart
```diff
@@ -0,0 +1,113 @@
...
+import 'package:neon_http_client/src/interceptors/interceptors.dart';
+import 'package:universal_io/io.dart';
+
+/// An exception caused by an error in a [NeonClient].
+sealed class NeonClientException extends http.ClientException {
+  NeonClientException(super.message, [super.uri]);
+}
+
+/// An exception caused by a [HttpInterceptor].
+final class InterceptionFailure extends NeonClientException {
```
provokateurin, line 13:
```suggestion
final class InterceptionException extends NeonClientException {
```
Common naming scheme for exceptions

packages/neon_http_client/lib/src/client.dart
```diff
@@ -0,0 +1,113 @@
...
+  /// Creates a new interceptor failure exception.
+  InterceptionFailure(super.message, [super.uri]);
+}
+
+/// A client configuring the clients for all Nextcloud APIs.
+///
+/// To access the APIs of a particular app import the extensions through `package:nextcloud/{id}.dart`.
```
provokateurin, line 20: Wrong docs

packages/neon_http_client/LICENSE
provokateurin, line 1: Should be AGPL, no?

packages/neon_http_client/lib/src/client.dart
```diff
@@ -0,0 +1,113 @@
...
+  /// Creates a new Neon http client.
+  ///
+  ///
+  /// It is good practice to set the [userAgent] to allow server admins to identify clients.
+  /// A custom HTTP client can be provided through [client].
+  /// Additionally a [cookieStore] can be specified to save cookies across requests.
+  /// Some endpoints require the use of a cookies persistence.
+  NeonClient({
+    http.Client? client,
+    Iterable<HttpInterceptor>? interceptors,
```
provokateurin, line 31: Do we actually have to expose this? I didn't see any usage of it at the moment and I'd say this client should handle the needs by providing extra parameters like the userAgent or cookieStore which are nicer to use.

packages/neon_http_client/lib/src/client.dart
```diff
@@ -0,0 +1,113 @@
...
+    for (final interceptor in _interceptors) {
+      if (interceptor.shouldInterceptResponse(interceptedResponse)) {
+        try {
+          interceptedResponse = await interceptor.interceptResponse(
+            response: interceptedResponse,
+            url: url,
+          );
+        } catch (_, stackTrace) {
+          Error.throwWithStackTrace(
+            InterceptionFailure('Failed to intercept response', request.url),
```
provokateurin, line 97:
```suggestion
            InterceptionFailure('Failed to intercept response', url),
```

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
```diff
@@ -0,0 +1,98 @@
...
+  @override
+  bool shouldInterceptRequest(http.BaseRequest request) {
+    return !request.headers.containsKey(HttpHeaders.cookieHeader);
+  }
+
+  @override
+  Future<http.BaseRequest> interceptRequest({required http.BaseRequest request}) async {
+    if (!shouldInterceptRequest(request)) {
+      throw StateError('Request already contains a cookie header and should not be intercepted.');
+    }
```
provokateurin, line 26: Checking this again feels useless to me, the caller should be responsible for only calling when allowed

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
```diff
@@ -0,0 +1,98 @@
...
+  @override
+  bool shouldInterceptResponse(http.StreamedResponse response) {
+    return response.headers.containsKey(HttpHeaders.setCookieHeader);
+  }
+
+  @override
+  Future<http.StreamedResponse> interceptResponse({required http.StreamedResponse response, required Uri url}) async {
+    if (!shouldInterceptResponse(response)) {
+      throw StateError('Response should not be intercepted.');
+    }
```
provokateurin, line 63: Here again

packages/neon_http_client/lib/src/interceptors/http_interceptor.dart
```diff
@@ -0,0 +1,26 @@
...
+
+/// Interceptor that can manipulate http requests and responses.
+abstract interface class HttpInterceptor {
+  /// Whether this interceptor should intercept requests.
+  bool shouldInterceptRequest(http.BaseRequest request);
+
+  /// Intercepts the given [request].
+  ///
+  /// Provided requests are not finalized yet. It is an error for an interceptor
+  /// to request itself.
```
provokateurin, line 13: > It is an error for an interceptor to request itself.

I'm not sure I understand, you mean the interception callback should not execute the request?

packages/neon_http_client/lib/src/interceptors/http_interceptor.dart
```diff
@@ -0,0 +1,26 @@
...
+  /// Whether this interceptor should intercept response.
+  bool shouldInterceptResponse(http.StreamedResponse response);
+
+  /// Intercepts the given [response].
+  ///
+  /// Until package:http 2.0 makes [http.BaseResponseWithUrl] mandatory the request url is used.
+  FutureOr<http.StreamedResponse> interceptResponse({
+    required http.StreamedResponse response,
+    required Uri url,
+  });
```
provokateurin, line 25: Wouldn't it make more sense to use http.Response here as then the interceptors can also modify the body?

packages/neon_http_client/lib/src/client.dart
provokateurin: Ah maybe this is handy for the SVG dimensions rewrite interceptor which doesn't really belong into this package.

packages/neon_http_client/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
+name: neon_http_client
+description: A RFC compliant cookie_jar alternative
```
provokateurin, line 2: Wrong description

packages/neon_http_client/pubspec.yaml
```diff
@@ -0,0 +1,30 @@
...
+  http: ^1.0.0
+  meta: ^1.0.0
+  universal_io: ^2.0.0
+
+dev_dependencies:
+  http_client_conformance_tests:
+    git:
+      url: https://github.com/dart-lang/http
+      path: pkgs/http_client_conformance_tests
+      ref: 757438e938ebe81f4b130e6372897b5f977615a6
```
provokateurin, line 24: Shouldn't this be pinned to the tag of the http package version we use?

packages/neon_http_client/test/client_test.dart
```diff
@@ -0,0 +1,170 @@
...
+        client: MockClient((request) async {
+          return Response('', 200);
+        }),
+        cookieStore: cookieStore,
+      );
+
+      await client.get(uri);
+
+      verify(() => cookieStore.loadForRequest(any(that: equals(uri)))).called(1);
+      verifyNever(() => cookieStore.saveFromResponse(any(that: equals(uri)), any()));
```
provokateurin, line 72: It should also test the method is called when the response has a set-cookie header

packages/neon_http_client/test/client_test.dart
```diff
@@ -0,0 +1,170 @@
...
+          () => interceptor.interceptRequest(
+            request: any(named: 'request', that: equals(request)),
+          ),
+        );
+        verifyNever(
+          () => interceptor.interceptResponse(
+            response: any(named: 'response', that: equals(fakeResponse)),
+            url: any(named: 'url', that: equals(uri)),
+          ),
+        );
```
provokateurin, line 104:
```suggestion
        verifyNever(
          () => interceptor.interceptRequest(
            request: any(named: 'request'),
          ),
        );
        verifyNever(
          () => interceptor.interceptResponse(
            response: any(named: 'response'),
            url: any(named: 'url'),
          ),
        );
```
Otherwise a mismatched value would cause the count to be 0 if there was something wrong (only necessary because this is a negative case)

packages/neon_http_client/test/interceptors/http_interceptor_test.dart
```diff
@@ -0,0 +1,32 @@
...
+
+  @override
+  bool shouldInterceptResponse(http.StreamedResponse response) => throw UnimplementedError();
+}
+
+void main() {
+  group('HttpInterceptor', () {
+    test('can be implemented', () {
+      expect(_TestHttpInterceptor(), isNotNull);
+    });
```
provokateurin, line 30: What's the point of this?

provokateurin (changes requested):
Very nice work! I think this was absolutely the right move to allow us finally progressing on the caching stuff.

packages/neon_http_client/lib/src/client.dart
Leptopoda: This is on purpose.
It's easier to debug a thrown exception with the url the request was made against, rather than having to figure out what redirects ended up giving the response.

what do you think?

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
Leptopoda: > `should`

I'd rather have a sanity check to catch any wrongdoing. Should i can change it to an assertion?

packages/neon_http_client/lib/src/interceptors/http_interceptor.dart
Leptopoda: https://pub.dev/documentation/http/latest/http/Client/send.html

The send function of the client itself expects a streamed response.
In case the body needs to be modified, a stream transformer is probably the right thing to use

packages/neon_http_client/pubspec.yaml
Leptopoda: I don't think these packages are released in parallel.
We should always use the latest tests available, but not having any tag at all feels wrong.

packages/neon_http_client/lib/src/client.dart
Leptopoda: I also renamed the file to match the class name.
We have way to many `client.dart` files

packages/neon_http_client/lib/src/client.dart
Leptopoda: I can see the point but later on we'll also need to wrap the `TimeoutException` and just calling it `TimeoutFailure` doesn't seem that bad.

packages/neon_http_client/lib/src/client.dart
provokateurin: But knowing the final url is also helpful :thinking:
Then just leave it as it was.

packages/neon_http_client/lib/src/client.dart
provokateurin: Hm we can do that for TimeoutException, but if it is not necessary here then it should follow the common naming scheme.

packages/neon_http_client/pubspec.yaml
provokateurin: Ok makes sense :+1: Let's hope renovate will actually update this

packages/neon_http_client/lib/src/client.dart
Leptopoda: Yeah I'll leave it.
I don't really expect our interceptors to throw anything meaningful. They can only throw deeper errors, like a sqlite exception from the cookie persistence.
In that case, the url isn't that helpful anyway.

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
provokateurin: Yeah assertion is good

packages/neon_http_client/test/client_test.dart
Leptopoda: I'm not sure, this is the right place.
This is just testing that specifying a cookieStore also adds the CookieStoreInterceptor. Testing the interceptor itself is done in its test.

If it makes more sense we can make the _interceptros field visible for testing and also verify that without making a request.

provokateurin: I'd prefer that

packages/neon_http_client/LICENSE
provokateurin: Still links to the BSD-3 license

packages/neon_http_client/lib/neon_http_client.dart
```diff
@@ -0,0 +1,12 @@
+/// An [RFC6265](https://httpwg.org/specs/rfc6265.html) compliant cookie store.
+///
+/// Cookies are stored inside of `CookieStore`s parsing the received cookies
+/// into `StorableCookie`s that can be saved in a `CookiePersistence`.
+///
+/// This package has an implementation independent api and provides utilities
+/// that make it easy to extend and implemented into existing storage backends
+/// like SQLite databases.
```
provokateurin, line 8: Wrong docs

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
```diff
@@ -0,0 +1,100 @@
...
+  /// Load the cookies for specified [uri].
+  FutureOr<List<Cookie>> loadForRequest(Uri uri);
+
+  /// Save the [cookies] for specified [uri].
+  FutureOr<void> saveFromResponse(Uri uri, List<Cookie> cookies);
+}
+
+/// A HttpInterceptor persisting cookies in the provided [cookieStore].
+@internal
+final class CookieStoreInterceptor extends CookieInterceptor {
```
provokateurin, line 83: Why do we need this? There will never be another implementation of the interceptor so the abstraction is not necessary.

packages/neon_http_client/lib/src/neon_http_client.dart
```diff
@@ -0,0 +1,114 @@
+import 'package:built_collection/built_collection.dart';
+import 'package:cookie_store/cookie_store.dart';
+import 'package:http/http.dart' as http;
+import 'package:meta/meta.dart';
+import 'package:neon_http_client/src/interceptors/interceptors.dart';
+import 'package:universal_io/io.dart';
+
+/// An exception caused by an error in a [NeonHttpClient].
+sealed class NeonClientException extends http.ClientException {
```
provokateurin, line 9:
```suggestion
sealed class NeonHttpClientException extends http.ClientException {
```
Since the class was renamed this should be renamed as well

packages/neon_http_client/test/neon_http_client_test.dart
```diff
@@ -0,0 +1,160 @@
...
+        await client.send(request);
+
+        verifyNever(
+          () => interceptor.interceptRequest(
+            request: any(named: 'request'),
+          ),
+        );
+        verifyNever(
+          () => interceptor.interceptResponse(
+            response: any(named: 'response', that: equals(fakeResponse)),
```
provokateurin, line 91:
```suggestion
            response: any(named: 'response'),
```
Dangerous for a negative case

provokateurin (changes requested):

packages/neon_http_client/lib/src/interceptors/cookie_interceptor.dart
Leptopoda: Now we already have the abstraction, so why not keep it?
It's never a bad idea to wrap foreign interfaces.

packages/neon_http_client/test/interceptors/http_interceptor_test.dart
Leptopoda: I like to have a test per dart file.
This way everything is explicitly covered.

If we where to remove the interceptors of this package (like moving them into the framework) the interface would still be verified to be implementable (not sealed ...).

provokateurin: Hm ok :woman_shrugging:

provokateurin:
CI will fail, check analyze logs

provokateurin (approved):

PR #2289, opened 2024-07-19
refactor(neon_talk): Parse timestamps to local timezone

provokateurin:
Follow-up to https://github.com/nextcloud/neon/pull/2287.
I originally parse the timestamps as UTC because they were in the nextcloud package where the local timezone might not be desired, but in neon_talk we always want the local timezone.


Leptopoda:
Can you add a test for that?

Leptopoda (approved):
Ok 🤷‍♀️

PR #2293, opened 2024-07-21
fix(nextcloud_test): Skip CSRF token requests in fixtures

provokateurin:
Needed for https://github.com/nextcloud/neon/issues/1901


Leptopoda:
Why exactly is this needed?

provokateurin:
This messes with the order of tests, since we only acquire a token once. This means the request is only present in the fixture of the first test that runs. Now if the order is changed (e.g. using a random seed) the fixtures will fail.

Leptopoda (approved):

PR #2294, opened 2024-07-21
test(nextcloud): Make tests more generic

provokateurin:
Needed for https://github.com/nextcloud/neon/issues/1901

All resources used/created by the tests need to be reset/removed such that they don't interfere with the other tests.


packages/nextcloud/test/webdav_test.dart
```diff
@@ -201,14 +201,20 @@ void main() {
       setUpAll(() async {
         container = await DockerContainer.create(preset);
         client = await TestNextcloudClient.create(container);
+
+        await client.webdav.mkcol(PathUri.parse('webdav'));
```
Leptopoda, line 205: Can we just call this `test`?
Looking at the fixtures, I was really confused why the uri was `/webdav/webdav`.

If it needs to be unique, I'd rather use `webdav_test`

packages/nextcloud/test/notes_test.dart
```diff
@@ -153,13 +153,18 @@ void main() {
...
+        );
+
         var response = await client.notes.getSettings();
         expect(response.statusCode, 200);
         expect(() => response.headers, isA<void>());
 
-        expect(response.body.notesPath, 'Notes');
-        expect(response.body.fileSuffix, '.md');
-        expect(response.body.noteMode, notes.Settings_NoteMode.rich);
+        expect(response.body, expectedSettings);
```
Leptopoda, line 167:
```suggestion
        expect(response.body, equalsBuilt(expectedSettings));
```

This produces better error messages in case the test fails.

packages/nextcloud/test/user_status_test.dart
```diff
@@ -190,12 +190,7 @@ void main() {
           final response = await client.userStatus.statuses.findAll();
           expect(response.statusCode, 200);
           expect(() => response.headers, isA<void>());
-          expect(response.body.ocs.data, hasLength(1));
-          expect(response.body.ocs.data[0].userId, 'user1');
-          expect(response.body.ocs.data[0].message, null);
-          expect(response.body.ocs.data[0].icon, null);
-          expect(response.body.ocs.data[0].clearAt, null);
-          expect(response.body.ocs.data[0].status, user_status.$Type.online);
+          expect(response.body.ocs.data, isNotEmpty);
```
Leptopoda, line 193: why was some test information removed here?

packages/nextcloud/test/dashboard_test.dart
```diff
@@ -21,17 +21,12 @@ void main() {
       test('Get widgets', () async {
         final response = await client.dashboard.dashboardApi.getWidgets();
         expect(response.statusCode, 200);
-        expect(
```
Leptopoda, line 24: again; why was some information removed?

Leptopoda (changes requested):

packages/nextcloud/test/user_status_test.dart
provokateurin: Because other users might have statuses too and we can not rely on the first status to be the one we need. Now that I think about it this can probably be changed to find the one we want to check.

packages/nextcloud/test/dashboard_test.dart
provokateurin: Not reliable, if you have a minimal version of the server then apps like recommendations are missing and we can not rely on them here.

packages/nextcloud/test/user_status_test.dart
Leptopoda: yep. I think the `containsOnce` matcher would be fitting

packages/nextcloud/test/user_status_test.dart
```diff
@@ -190,12 +190,17 @@ void main() {
...
+          expect(
+            response.body.ocs.data,
+            containsOnce(
+              user_status.Public(
```
Leptopoda, line 197: this could also be wrapped in `equalsBuilt()`

Leptopoda (approved):

PR #2295, opened 2024-07-21
refactor(nextcloud): reorganize package structure

Leptopoda:
It's a bit messy around the webdav stuff, but I hope it's still reviewable.


provokateurin (approved):
So much better than before :)

PR #2300, opened 2024-07-22
feat(neon_http_client): add csrf interceptor 

Leptopoda:
The coverage is a bit different from the one reported by the `flutter_test` package, as they both use different coverage collectors.

Let's hope it won't be too off from the current one

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -41,7 +41,11 @@ scripts:
   test: >
     melos run test:dart &&
     melos run test:flutter
-  test:dart: melos exec --no-flutter --concurrency=1 --fail-fast --dir-exists=test -- flutter test --concurrency=$(nproc --all) --coverage
+  test:dart: >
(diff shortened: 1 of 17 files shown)


provokateurin:
Wasn't there a reason why we used flutter for running tests to collect coverage? I remember something wasn't working, but can't find the PR anymore.

provokateurin (approved):
Works as expected, but please check my last comment before merging.

Leptopoda:
The way we collected the coverage before was broken and I think that you mainly wanted to unify both test cases.
 https://github.com/nextcloud/neon/pull/1858
 
 what do you think?

provokateurin:
Please break the tests on purpose to ensure the job will fail, since that was one of the problems we had before

provokateurin:
Thanks for the test, seems like the CI will still fail properly :+1:

PR #2301, opened 2024-07-22
feat(neon_http_client): allow timing out requests

Leptopoda:
Why is the timeout an optional constant and not the fallback value of the constructor?
Because the nextcloud_test package should not time out requests :)


packages/neon_http_client/test/neon_http_client_test.dart
```diff
@@ -156,5 +156,38 @@ void main() {
         );
       });
     });
+
+    group('timeout', () {
+      test('does not time out with not timeout', () async {
```
provokateurin, line 161:
```suggestion
      test('does not time out without timeout', () async {
```

provokateurin (approved):

PR #2308, opened 2024-07-22
feat(nextcloud): Implement chunked upload

provokateurin:
https://github.com/nextcloud/neon/issues/1123

This PR isn't finished because I want to discuss the direction first.
For now I made the WebDavClient re-usable for WebDAV extensions by allowing to configure the endpoint. This will also be necessary for https://github.com/nextcloud/neon/issues/718.
I'm not sure if this is the right way or if the methods should be re-implemented.
There is also the problem with the username because currently we use the `remote.php/webdav` endpoint which doesn't require it, but any other endpoint like `remote.php/dav/files/{username}` or  `remote.php/dav/uploads/{username}` need it. Unfortunately the root client only knows the login name which is not necessarily the username so it needs to be passed manually with the endpoint.

For the chunked upload part I want to add some helper methods for uploading a stream or file that is automatically divided in chunks so it is easily usable.


Leptopoda:
This might not answer your questions about how to handle the clients, but the http package already provides a `MultipartRequest` using the `multipart/form-data` content type. From a quick search in the server repo, this is also what the server expects.
 https://github.com/nextcloud/server/blob/f9d4becf60da69f272f6c5700bbdf5cb99761bc4/apps/files/js/jquery.fileupload.js#L142-L145
https://pub.dev/documentation/http/latest/http/MultipartRequest-class.html

provokateurin:
That's the old unchunked upload which doesn't help at all here. I followed https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/chunking.html to implement it which is completely custom.

Leptopoda:
What methods will the caldav and carddav client need?
The proposed patch only makes use of the put method.

provokateurin:
They would also use all the other methods because they just operate on the same "filesystem" structures. Using streams and files is probably not useful there as the data is usually small and not stored in files, so those would be WebDAV specific.

provokateurin:
I can make a quick diagram later to explain what I have in mind for the structure.

Leptopoda:
Sure, that would help a lot.
I'm going to read a bit into carddav and caldav but in general I'd prefer to have a client that we extend rather than wrap.

provokateurin:
I feel like working on this without looking more into https://github.com/nextcloud/neon/issues/718 makes no sense because I don't know what will be necessary for it. If I knew that it would be easy to chose the right way to abstract this.
I'll give it a shot then, let's see where it goes.

PR #2323, opened 2024-07-27
feat(nextcloud_test): Add local test target

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1901

I can split this PR into multiple ones if desired.

Removing the maximum version check is done because major versions of apps don't necessarily indicate incompatible APIs. In fact, none of the existing APIs should ever have breaking changes except for cases when they are removed which also shouldn't happen.

I validated that the latest version of the server an all relevant apps work just fine with our current tests :tada:


packages/nextcloud_test/lib/src/test_target/test_target.dart
```diff
@@ -0,0 +1,68 @@
...
+
+/// Creates a new [TestTargetFactory].
+TestTargetFactory createTestTargetFactory() {
+  return DockerContainerFactory();
+}
+
+/// Creates a new [NextcloudClient] for a given [username].
+///
+/// It is expected that the password of the user matches the its [username].
+FutureOr<NextcloudClient> createTestTargetClient(
```
Leptopoda, line 52: Why isn't this a function of the `TestTargetInstance`?
I'd rather make it a super class that is extended, so the instance doesn't need to be passed into the call every thime.

packages/nextcloud_test/lib/src/test_target/test_target.dart
```diff
@@ -0,0 +1,68 @@
...
+
+  /// URL where the target is available from itself.
+  Uri get targetURL;
+
+  /// Creates an app password for [username] on the instance.
+  FutureOr<String> createAppPassword(String username);
+}
+
+/// Creates a new [TestTargetFactory].
+TestTargetFactory createTestTargetFactory() {
```
Leptopoda, line 39: What is the benefit of this function?
I don't see how this helps.


Why not have this as a factory constructor in the `TestTargetFactory` class.

packages/nextcloud/doc/development.md
```diff
@@ -0,0 +1,41 @@
...
+All tests should be standalone and never affect any other tests.
+With enough caution this is feasible to implement, but to verify the tests do not depend on the order of execution they should be run multiple times with different seeds:
+
+```sh
+for i in {0..4}; do fvm flutter test --test-randomize-ordering-seed $i; done
+```
+
+## Testing against development versions
+
+Make sure to have the following apps in your development setup and the appropriate versions you want to test checked out:
```
Leptopoda, line 17: > in your development setup

What do you mean by that?
That you need the git submodules checked out?

Leptopoda (commented):
I really like the abstraction between the docker and local version :)
Nice work :+1:

packages/nextcloud/doc/development.md
provokateurin: The setup when you develop on the server or an app for server. I'll clarify this

packages/nextcloud_test/lib/src/test_target/local.dart
```diff
@@ -0,0 +1,114 @@
...
+  void destroy() {}
+
+  @override
+  Uri hostURL = Uri.parse(Platform.environment['URL']!);
+
+  @override
+  late Uri targetURL = hostURL;
+
+  @override
+  FutureOr<String> createAppPassword(String username) async {
```
Leptopoda, line 89: This always returns a future
```suggestion
  Future<String> createAppPassword(String username) async {
```

packages/nextcloud_test/lib/src/test_target/test_target.dart
```diff
@@ -0,0 +1,75 @@
...
+  /// URL where the target is available from itself.
+  Uri get targetURL;
+
+  /// Creates an app password for [username] on the instance.
+  FutureOr<String> createAppPassword(String username);
+
+  /// Creates a new [NextcloudClient] for a given [username].
+  ///
+  /// It is expected that the password of the user matches the its [username].
+  FutureOr<NextcloudClient> createClient({
```
Leptopoda, line 51: This always returns a future
```suggestion
  Future<NextcloudClient> createClient({
```

Leptopoda (approved):

PR #2324, opened 2024-07-27
feat(nextcloud): add news endpoints for marking multiple articles as read starred

Leptopoda:


packages/nextcloud/test/news_test.dart
provokateurin, line 1: Would be nice to group the tests by subclients (like it is done in all other tests)

provokateurin (changes requested):
The first commit is wrong because booleans in query parameters are represented as 0/1 in Nextcloud. If the parameters were sent in the body then they could need to be booleans indeed.

packages/nextcloud/test/news_test.dart
Leptopoda: I also thought about it (or even have a separate dart file) but didn't do it for better reviewability.
I can do it as a separte commit.

Leptopoda:
> The first commit is wrong because booleans in query parameters are represented as 0/1 in Nextcloud. If the parameters were sent in the body then they could need to be booleans indeed.

Can you please clarify that? The docs say that it's meant to be a boolean:
https://nextcloud.github.io/news/api/api-v1-3/#initial-sync
And judging from the code that's the right thing
https://github.com/nextcloud/news/blob/350eea3aa1c62d0ccc18b522bed27a3bb13b208e/lib/Controller/ItemApiController.php#L61

provokateurin:
Yes, but the documentation shows a JSON body as the example which we don't make use of here.
You can see the old quirk the openapi-extractor had when it still used query paramters: https://github.com/nextcloud/openapi-extractor/commit/826eaf299f23cfcbb6507f5589bcdccab13225ff#diff-ab3a3d862e1786c66098dfc9cba0909b72dbdf080660a1e971ac7851157b2bb8L80
I don't know the location where this is implemented on the server side, but my guess would be that it is within Symfony.
I have no better explanation that this, it's just how it works :woman_shrugging:

Leptopoda:
I don't know where you see a json body

> /items?type=3&getRead=false&batchSize=-1

But if you say so ...

provokateurin:
The docs must be wrong there, only 0/1 work

packages/nextcloud/test/news_test.dart
Leptopoda: I decided to do this in a followup.

.vscode/settings.json
provokateurin, line 1: Unintended change?

provokateurin (approved):

packages/nextcloud/test/news_test.dart
Leptopoda: See #2325

PR #2325, opened 2024-07-28
refactor(nextcloud_test): unify test instance handling

Leptopoda:
Why?
1. this cleans up the tests themselves, as they no longer need to handle the client creation and target lifecycle.
2. The tests can be split into multiple files (maybe per sub client) for better structure


packages/nextcloud_test/lib/src/test_target/test_target.dart
```diff
@@ -6,28 +6,40 @@ import 'package:cookie_store/cookie_store.dart';
...
     final dir = Platform.environment['DIR'];
-    if (url != null && url.isNotEmpty && dir != null && dir.isNotEmpty) {
-      return LocalFactory();
+    final url = Platform.environment['URL'];
+    Uri? uri;
+    if (url != null) {
+      uri = Uri.tryParse(url);
+    }
+
+    if (uri != null && dir != null && dir.isNotEmpty) {
```
provokateurin, line 31: The hard fail was on purpose. If you give it faulty config now it will not complain and just use docker.

Leptopoda:
Making the test factory a singleton is not strictly needed, as the `NextcloudTester` could handle the creation and lifecycle.
But IMO there is no need of running tests against two backends (docker and local) at the same time.

provokateurin (changes requested):
Very nice cleanups, thank you :+1:

packages/nextcloud_test/lib/src/test_target/test_target.dart
Leptopoda: So you mean I should throw if the url can not be parsed?

provokateurin: Yes, same for empty DIR variable (that wasn't checked even before I think)

provokateurin (approved):

PR #2332, opened 2024-07-29
fix(nextcloud): fix news test fixtures

Leptopoda:
The second commit removes the retry policy and implements a global timeout/concurrency policy.


tool/setup.sh
```diff
@@ -10,6 +10,9 @@ echo "y" | fvm install
 
 if [ ! -v GITHUB_REPOSITORY ]; then
   fvm flutter precache
+else
+  echo "concurrency: $(nproc --all)
+timeout: 30s" > ~/.dart_test.yaml
```
provokateurin, line 15: I don't like this, can't we just leave the timeout to the default value?

Leptopoda: Sure, we can keep it unspecified as it already defaults to 30s.
But I like having our `melos run test` respect the value I have set for my machine (that isn't the number of cpu cores).

I'll revert that for now and leave this thought for a future discussion.

provokateurin (approved):

Issue #1123, opened 2023-11-06
Implement chunked uploading


provokateurin:
It helps to workaround max upload size limits and also enables clients to recover from a failed upload and continue where left off. Especially helpful for https://github.com/nextcloud/neon/issues/21 if big files are used.
Documentation is at https://docs.nextcloud.com/server/latest/developer_manual/client_apis/WebDAV/chunking.html but we sadly don't have an OpenAPI spec for it because it is a custom implementation. I'd say we write a spec for it by hand (and maybe contribute it upstream).

provokateurin:
Also relevant: https://github.com/nextcloud/server/issues/47682

Issue #718, opened 2023-09-06
Support CalDAV and CardDAV


provokateurin:

provokateurin:
The WebDAV client has to be refactored in order to create a generic DAV implementation. Then WebDAV, CalDAV and CardDAV just instantiate a DavClient and give it the correct data types for dealing with the props.

PR #2339, opened 2024-08-01
test(nextcloud): refactor test file structure

Leptopoda:
I hope the conventional commit types are right :)


packages/nextcloud/test/modles/version_check_test.dart
provokateurin, line 1: The folder name is called `modles` but should be called `models`.

packages/nextcloud/test/api/notifications/notifications_helpers_test.dart
```diff
@@ -0,0 +1,146 @@
...
+            'type': 'type',
+            'id': 'id',
+            'delete': false,
+            'delete-all': false,
+          }),
+        );
+      });
+    });
+
+    test('to string', () {
```
provokateurin, line 130: IMO this test is a bit unnecessary, toString() will never be called in production code and rarely while developing

provokateurin (commented):
LGTM, but I'm not sure if we should have webdav in lib/src/api and not in lib/src since it is very different from all the other apis. Not really related to this PR though as this was changed earlier already.

Leptopoda:
> I'm not sure if we should have webdav in lib/src/api

I can see your point. Let's see, where #1123 and #718 go to. Feel free to change it.

packages/nextcloud/test/api/notifications/notifications_helpers_test.dart
Leptopoda: I like having it to verify every attribute is also contained in the string, but I agree that it isn't necessary.
Now we have the test so :woman_shrugging:

provokateurin:
The changes to the news test should also change the paths of the fixtures :thinking:

Leptopoda:
We should add a CI that validates that.
I think a simple check that the repo didn't have any changes after the test would be enough?

provokateurin:
I think just failing when a new fixture had to be created is good enough. Since we do not use retries anymore the failure should be propagated to the command.

provokateurin (approved):

PR #2350, opened 2024-08-05
refactor: change package file structure

Leptopoda:
I didn't know what to do with the cookie_store packages.
Should I move them to the framework, or just keep them as they are?


provokateurin:
What about moving the `packages/neon/neon_*` packages to `packages/neon_framework/packages` and `packages/dynamite/dynamite_*` to `packages/dynamite/packages`?

Leptopoda:
I agree with dynamite, but am unsure about the neon_* ones.
With my proposed way, each `$consumer/packages` contains the packages that are needed for `$consumer` while moving them into the framework would switch that up.

provokateurin:
I see, they are a bit weird.

BTW you also need to fix all the paths in tool/ and any documentation.

provokateurin:
Maybe cookie_store should go into neon_http_client? OTOH it is a pretty standalone package that we could publish at some point so that others can use it too.

Leptopoda:
> BTW you also need to fix all the paths in tool/ and any documentation.

I've tried my best to search for all packages I changed and hope I didn't miss any.

> Maybe cookie_store should go into neon_http_client? 

Let's leave it there for now.

provokateurin:
Instead of making a long and tedious review I just fixed all the problems. I found quite a few, but I think now everything is fixed.

provokateurin:
@Leptopoda Let me know if you find my changes good, then I'll approve and we can merge :rocket:

provokateurin:
BTW I already love working with this structure, it is so much better than what we had before! :heart:

packages/neon_framework/example/pubspec.lock
```diff
@@ -1666,5 +1666,5 @@ packages:
     source: hosted
     version: "3.1.2"
 sdks:
-  dart: ">=3.4.0 <4.0.0"
+  dart: ">=3.5.0 <4.0.0"
```
provokateurin, line 1669: I think this might cause the analyzer issues? At least it looks weird to me because there should be nothing that is causing this :thinking:

provokateurin: I was able to revert the unintended changes to the pubspec.lock, but it didn't fix the problem.

Leptopoda: I can confirm your findings.
Pinning all sdk versions to 3.4 does not trigger this issue.
The `context.mounted` not working is still an issue that requires us to ignore the errors (I'll report upstream).

.github/renovate.json
```diff
@@ -30,7 +30,6 @@
       ],
       "matchCurrentVersion": "!/^0/",
       "matchDepNames": [
-        "!app",
```
Leptopoda, line 33: Why this change?

provokateurin: Because the example packages shouldn't be treated like internal packages where we always update to the latest version.

Leptopoda: but shouldn't we always use the latest version for all our examples?
1. verify that our examples still work with the latest versions.
2. with this being the app we ship to end users, I'd like to make sure to pull in all available fixes

provokateurin: Hm ok makes sense

provokateurin (approved):

provokateurin:
:tada:

PR #2352, opened 2024-08-05
fix(nextcloud)!: Revert request body usage for non-PUT/POST/PATCH requests

provokateurin:
https://github.com/nextcloud/neon/pull/2129
https://github.com/nextcloud/neon/issues/2305

While the change is technically not breaking because the previous change wasn't released yet, this makes it easier to find the change in the changelog and understand what has happened.


Leptopoda (approved):
LGTM
Is there a reason for this still being a draft?

provokateurin:
CI was failing because I only adjusted the nextcloud tests and not the bloc tests

PR #2368, opened 2024-08-12
chore(tool): Disable custom_lint

provokateurin:
https://github.com/invertase/dart_custom_lint/issues/261 prevents us to upgrade to Dart 3.5/Flutter 3.24.
Once fixed this commit can be reverted.

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -36,8 +36,8 @@ scripts:
   format: dart format --fix --line-length 120 .
   format:check: dart format --output=none --set-exit-if-changed --line-length 120 .
   analyze: >
-    dart analyze --fatal-infos . &&
-    dart run custom_lint --fatal-infos .
+    dart analyze --fatal-infos . #&&
+    #dart run custom_lint --fatal-infos .
   test: >
     melos run test:dart &&
     melos run test:flutter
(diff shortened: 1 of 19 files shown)


provokateurin:
I updated https://github.com/nextcloud/neon/pull/2360 to ensure it will work

Leptopoda (approved):
LGTM, but wouldn't it be enough to disable the analyzer plugin and `custom_lint` dependency?

`neon_lints` would still depend on `custom_lint_builder`, but the functionality wouldn't be exported. That way we would avoid commenting out all the dart files.

provokateurin:
I'll check it

provokateurin:
I added it to https://github.com/nextcloud/neon/pull/2360 to ensure I properly disabled it now and it will work with Dart 3.5.

provokateurin:
For the record:
```
ERROR: /home/runner/.pub-cache/hosted/pub.dev/custom_lint_builder-0.6.4/lib/src/client.dart:46:1: Error: 'LintCode' is imported from both 'package:analyzer/src/dart/error/lint_codes.dart' and 'package:custom_lint_core/src/lint_codes.dart'.
```
So the custom_lint_builder dependency in neon_lints has to go too.

PR #2375, opened 2024-08-13
fix(neon_framework): Only request avatars in valid sizes

provokateurin:
The server restricts the valid sizes to 64 and 512: https://github.com/nextcloud/server/blob/21db61817467169d225af1e4a96bb37f9feaf70e/core/Controller/AvatarController.php#L70-L80

To avoid multiple cache entries with the same content we can already apply this limitation on the client side.

### DIFF
--- a/packages/neon_framework/lib/src/widgets/user_avatar.dart
+++ b/packages/neon_framework/lib/src/widgets/user_avatar.dart
@@ -62,7 +62,12 @@ class _UserAvatarState extends State<NeonUserAvatar> {
       builder: (context, constraints) {
         final brightness = Theme.of(context).brightness;
         size = constraints.constrain(Size.square(widget.size ?? largeIconSize)).shortestSide;
-        final pixelSize = (size * MediaQuery.of(context).devicePixelRatio).toInt();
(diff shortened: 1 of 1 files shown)


Leptopoda (approved):
I guess the spec can not reflect this?

Leptopoda:
I wonder if we should change the cache logic that the higher quality image is always prefered.

```
cache contains size 64 and 64 is requested:
-> return cached

cache contains size 64 and 256 is requested:
-> request 256 and store it in the cache

cache contains size 256 and 64 is requested:
-> return 256 without making a new request. the UI will be able to scacle it 
```

provokateurin:
> I guess the spec can not reflect this?

It can, I didn't think about adding it. Will do that!

> I wonder if we should change the cache logic that the higher quality image is always prefered.

In theory a good idea, but so far I've never seen an avatar >64 in my cache and implementing this would need quite some deep level of access to the cache, right? I don't think the effort will be worth it.

provokateurin:
Sent https://github.com/nextcloud/server/pull/47211 upstream, I'm not going to make a patch for this though.

PR #2381, opened 2024-08-13
chore(app): Update outdated web sqflite files

provokateurin:
For a while these files could not be regenerated due to the package:web migration and subsequent stabilization/breaking change of the package. Since we finally were able to upgrade today this is no longer a problem.


provokateurin:
No idea how the wasm file was reduced by 99%, but I confirmed it still works :woman_shrugging:

Leptopoda (approved):

PR #2386, opened 2024-08-15
refactor(neon_framework): Remove url launcher hack

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2349

Using url_launcher directly is not possible because the URI needs to be complete because we often only get the path without the host etc.

### DIFF
--- a/packages/neon_framework/lib/src/pages/login_flow.dart
+++ b/packages/neon_framework/lib/src/pages/login_flow.dart
@@ -2,14 +2,14 @@ import 'dart:async';
 
 import 'package:flutter/material.dart';
 import 'package:meta/meta.dart';
-import 'package:neon_framework/l10n/localizations.dart';
+import 'package:neon_framework/models.dart';
 import 'package:neon_framework/src/bloc/result.dart';
 import 'package:neon_framework/src/blocs/login_flow.dart';
 import 'package:neon_framework/src/router.dart';
 import 'package:neon_framework/src/widgets/error.dart';
 import 'package:neon_framework/src/widgets/linear_progress_indicator.dart';
+import 'package:neon_framework/utils.dart';
 import 'package:nextcloud/core.dart' as core;
-import 'package:url_launcher/url_launcher_string.dart';
 
 @internal
 class LoginFlowPage extends StatefulWidget {
@@ -39,10 +39,7 @@ class _LoginFlowPageState extends State<LoginFlowPage> {
 
     initSubscription = bloc.init.listen((result) async {
       if (result.hasData) {
-        await launchUrlString(
-          result.requireData.login,
-          mode: LaunchMode.externalApplication,
-        );
+        await launchUrl(NeonProvider.of<Account>(context), result.requireData.login);
       }
     });
 
--- a/packages/neon_framework/lib/src/testing/mocks.dart
+++ b/packages/neon_framework/lib/src/testing/mocks.dart
@@ -2,7 +2,6 @@
 
 import 'package:flutter/widgets.dart';
 import 'package:go_router/go_router.dart';
-import 'package:meta/meta.dart';
 // ignore: depend_on_referenced_packages
 import 'package:mocktail/mocktail.dart';
 import 'package:neon_framework/blocs.dart';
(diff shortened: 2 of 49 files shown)


provokateurin:
Always using a real Account ensures the `completeUri` functionality is properly used. Less mocks, less problems in the real world.

packages/neon_framework/lib/src/testing/mocks.dart
```diff
@@ -23,10 +22,17 @@ import 'package:shared_preferences_platform_interface/shared_preferences_platfor
...
+  Uri? serverURL,
+  String? username,
+  String? password,
+}) {
+  return Account(
+    (b) => b
+      ..serverURL = serverURL ?? Uri.parse('https://cloud.example.com:8443/nextcloud')
+      ..username = username ?? 'username'
+      ..password = password ?? 'password',
+  );
```
Leptopoda, line 35: Please add the default username and password value to the parameter list.

I'm also not a huge fan of this common code. I can see that you don't want to duplicate it over and over again, but I prefer unit tests to be independent.

packages/neon_framework/lib/src/pages/login_flow.dart
```diff
@@ -39,10 +38,7 @@ class _LoginFlowPageState extends State<LoginFlowPage> {
 
     initSubscription = bloc.init.listen((result) async {
       if (result.hasData) {
-        await launchUrlString(
-          result.requireData.login,
-          mode: LaunchMode.externalApplication,
```
Leptopoda, line 44: Why did you change the launch mode?
My last information was that it is preferred to use the browser, as webview might have issues with proxy setups

packages/neon_framework/lib/src/utils/launch_url.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:flutter/cupertino.dart';
+import 'package:neon_framework/models.dart';
+import 'package:neon_framework/utils.dart';
+import 'package:url_launcher/url_launcher.dart' as url_launcher;
+
+Future<bool> launchUrl(BuildContext context, String url) async {
+  final account = NeonProvider.of<Account>(context);
+  final uri = account.completeUri(Uri.parse(url));
+  return url_launcher.launchUrl(uri);
+}
```
Leptopoda, line 10: I'm not a big fan of this.
Passing around the BuildContext into utility functions.

I'd rather have this as an extension to the account itself.

Leptopoda (changes requested):

packages/neon_framework/lib/src/utils/launch_url.dart
provokateurin: `Account.launchUrl(...)` doesn't make sense to me. Why is the Account launching the URL? We only need the Account to ensure the URL is valid which is a different thing.

packages/neon_framework/lib/src/pages/login_flow.dart
provokateurin: I forgot to change it

packages/neon_framework/lib/src/testing/mocks.dart
provokateurin: I think this is better because it ensures we always use a URL that is special.
This doesn't make tests depending on each other and I don't see a downside in this shared mock account.

packages/neon_framework/lib/src/utils/launch_url.dart
Leptopoda: This is a bad timing for this kind of discussion.
In general, I'd say this should be handled by a bloc or for now in the widget itself.

It isn't obvious to the caller that this method needs access to the `Account` and therefore requires it to be in injected into the tree.
Maybe just make the account a parameter instead of the context and let the resolving happen outside this method.

We can discuss letting this is for now knowing that blocs will come to properly fix this.

provokateurin: Ok, I'll go with the account as a parameter.

packages/neon_framework/lib/src/pages/login_flow.dart
Leptopoda: Still the same.

provokateurin: No, the method that wraps it always uses LaunchMode.externalApplication

Leptopoda (approved):

PR #2389, opened 2024-08-15
Revert "chore(tool): Disable custom_lint"

provokateurin:
https://github.com/invertase/dart_custom_lint/issues/261 got fixed and everything works again.


provokateurin:
Ugh more weird analyzer stuff going on :grimacing:

provokateurin:
Most of the issues also make sense, the rules should have been disabled for some packages. The annoying thing is I can not reproduce it locally at all. So with all these analyzer issues there must be a deeper problem that's not only a problem of custom_lint.

provokateurin:
https://github.com/nextcloud/neon/pull/2392

PR #2393, opened 2024-08-15
feat(neon_lints): Add rule to prefer prefixed nextcloud imports

provokateurin:
Avoids namespace pollution and is already the way it is done almost everywhere (except for the files app it seems).

### DIFF
--- a/packages/neon_lints/lib/src/prefer_prefixed_nextcloud_import.dart
+++ b/packages/neon_lints/lib/src/prefer_prefixed_nextcloud_import.dart
@@ -1,0 +1,44 @@
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+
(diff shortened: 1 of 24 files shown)


packages/neon_lints/lib/src/prefer_prefixed_nextcloud_import.dart
```diff
@@ -0,0 +1,44 @@
...
+    CustomLintContext context,
+  ) {
+    context.registry.addImportDirective((node) {
+      final match = _regex.firstMatch(node.uri.stringValue ?? '');
+      if (match == null) {
+        return;
+      }
+
+      final id = match.group(1)!;
+      if (id == 'nextcloud' || id == 'utils' || id == 'ids') {
```
Leptopoda, line 32: `ids` is already deprecated, so I'm not sure why we should test for it.
Please add tests for the nextcloud and utils import to the example package

Leptopoda (approved):

PR #2401, opened 2024-08-16
fix(neon_framework): Exclude app icons from semantics

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2273


packages/neon_framework/lib/src/models/app_implementation.dart
```diff
@@ -156,25 +161,26 @@ abstract class AppImplementation<T extends Bloc, R extends AppImplementationOpti
   ///
   /// It is mandatory to provide a precompiled SVG under `assets/app.svg.vec`.
   /// SVGs can be precompiled with `https://pub.dev/packages/vector_graphics_compiler`
-  Widget buildIcon({
+  Widget buildIcon(
+    BuildContext context, {
```
Leptopoda, line 165: Still not a fan of passing around the context

provokateurin: But using a Builder widget is better? I don't see a problem with passing the context :woman_shrugging:

Leptopoda: You have no control, where the context is coming from.
A builder also functions as a boundary for rebuilds.

Leptopoda (approved):

PR #2403, opened 2024-08-16
feat(neon_framework): add account_repository 

Leptopoda:
Finally a end to end unit tested login flow.
1. I did not add a separate conventional commit type for the account_repository. As it is mostly going to be used by the framework itself, I feel like `neon_framework` is good enough. I don't want too many types (open for discussions)
2. This is still a draft, as I was not able to get all tests to pass. Especially the widget tests of all pages have issues with the widget finder. I have no idea what is causing this and will continue to look into it.


packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,329 @@
...
+      serverURL: serverURL,
+    );
+
+    try {
+      final response = await client.authentication.core.getStatus();
+      return response.body;
+    } on http.ClientException catch (error, stackTrace) {
+      Error.throwWithStackTrace(FetchStatusFailure(error), stackTrace);
+    }
+  }
```
Leptopoda, line 177: This is still in the account_repository.
We can still refactor it later on, and I'm not sure yet about how to best manage the NextcloudClients for it.

Once authenticated, the client of the account should be used everywhere, but here we explicitly need to work without any authentication.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,329 @@
...
+
+  (NextcloudClient, core.ClientFlowLoginV2PollRequestApplicationJson)? _flow;
+
+  final BehaviorSubject<({String? active, BuiltMap<String, Account> accounts})> _accounts =
+      BehaviorSubject.seeded((active: null, accounts: BuiltMap()));
+
+  /// A stream of account information.
+  ///
+  /// The initial state of the stream will be `(active: null, accounts: BuiltMap())`.
+  Stream<({String? active, BuiltList<Account> accounts})> get accounts => _accounts.stream.map((e) {
```
provokateurin, line 98: Or at least here the ID could be turned into the Account.

packages/neon_framework/packages/account_repository/lib/src/models/credentials.dart
```diff
@@ -0,0 +1,61 @@
...
+  static Serializer<Credentials> get serializer => _$credentialsSerializer;
+
+  /// Url of the server.
+  Uri get serverURL;
+
+  /// The user id.
+  String get username;
+
+  /// App password.
+  String? get password;
```
provokateurin, line 36: Might be a good point to rename the field to `appPassword` as this is different from the user `password`.

packages/neon_framework/packages/account_repository/lib/src/utils/authentication_client.dart
```diff
@@ -0,0 +1,44 @@
+// ignore_for_file: prefer_prefixed_nextcloud_import
+
+import 'package:meta/meta.dart';
+import 'package:nextcloud/core.dart' as nc_core;
+import 'package:nextcloud/nextcloud.dart';
+import 'package:nextcloud/provisioning_api.dart' as nc_provisioning_api;
+
+/// A AuthenticationClient client mock for testing.
```
provokateurin, line 8:
```suggestion
/// An AuthenticationClient client mock for testing.
```

packages/neon_framework/packages/account_repository/lib/src/utils/authentication_client.dart
```diff
@@ -0,0 +1,44 @@
+// ignore_for_file: prefer_prefixed_nextcloud_import
+
+import 'package:meta/meta.dart';
+import 'package:nextcloud/core.dart' as nc_core;
+import 'package:nextcloud/nextcloud.dart';
+import 'package:nextcloud/provisioning_api.dart' as nc_provisioning_api;
```
provokateurin, line 6:
```suggestion
import 'package:nextcloud/core.dart' as core;
import 'package:nextcloud/nextcloud.dart';
import 'package:nextcloud/provisioning_api.dart' as provisioning_api;
```
(CI should also tell you this with the new rule)

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
```diff
@@ -0,0 +1,52 @@
...
+  required http.Client httpClient,
+  required String userAgent,
+  required Credentials credentials,
+}) {
+  final cookieStore = NeonStorage().cookieStore(
+    accountID: credentials.id,
+    serverURL: credentials.serverURL,
+  );
+
+  final neonClient = NeonHttpClient(
```
provokateurin, line 18:
```suggestion
  final httpClient = NeonHttpClient(
```

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
```diff
@@ -0,0 +1,52 @@
...
+  );
+}
+
+/// Builds an unauthenticated [NextcloudClient] for the given [serverURL].
+NextcloudClient buildUnAuthenticatedClient({
+  required http.Client httpClient,
+  required String userAgent,
+  required Uri serverURL,
+}) {
+  final neonClient = NeonHttpClient(
```
provokateurin, line 41:
```suggestion
  final httpClient = NeonHttpClient(
```

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+    }),
+  ]);
+
+  group('AccountRepository', () {
+    test('failure equality', () {
+      expect(
+        FetchStatusFailure(accountList),
+        equals(FetchStatusFailure(accountList)),
+      );
+    });
```
provokateurin, line 97: Why only check this for equality?

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+          emits((accounts: BuiltList<Account>(), active: null)),
+        );
+      });
+    });
+
+    test('hasAccounts', () async {
+      when(() => storage.readAccounts()).thenAnswer((_) async => BuiltList());
+      await repository.loadAccounts();
+
+      expect(repository.hasAccounts, isFalse);
```
provokateurin, line 112: Maybe also check the value before even calling repository.loadAccounts()

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+
+      expect(repository.hasAccounts, isTrue);
+    });
+
+    test('accountByID', () async {
+      when(() => storage.readAccounts()).thenAnswer((_) async => accountList);
+      await repository.loadAccounts();
+
+      expect(repository.accountByID('invalid'), isNull);
+      expect(repository.accountByID('c73bc81eda746a735de697a0f0cbdb08f96ea3f3'), isNotNull);
```
provokateurin, line 125: Wouldn't it be better to not hardcode the value and take it from the list? This shouldn't test that the id hashing works, just that you can find accounts by their ids.
Same for all the other places where a hardcoded id is used in the tests.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+                .having((e) => e.active, 'active', equals('b0682d652840ef50a4115cc77109bedd8c577ccc')),
+          ),
+        );
+
+        verifyNever(() => storage.readLastAccount());
+      });
+
+      test('uses last active account when specified', () async {
+        when(() => storage.readAccounts()).thenAnswer((_) async => accountList);
+        when(() => storage.readLastAccount()).thenAnswer((_) async => 'c73bc81eda746a735de697a0f0cbdb08f96ea3f3');
```
provokateurin, line 207: Also verify the method was actually called once. It could still not get called and the first account was picked instead.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+            isA<_AccountStream>()
+                .having((e) => e.accounts.map((e) => e.credentials), 'accounts', containsAllInOrder(accountList))
+                .having((e) => e.active, 'active', equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')),
+          ),
+        );
+      });
+
+      test('ignores last active account when not in accounts', () async {
+        when(() => storage.readAccounts()).thenAnswer((_) async => accountList);
+        when(() => storage.readLastAccount()).thenAnswer((_) async => 'invalid');
```
provokateurin, line 222: Verify again

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+            isA<_AccountStream>()
+                .having((e) => e.accounts.map((e) => e.credentials), 'accounts', containsAllInOrder(accountList))
+                .having((e) => e.active, 'active', equals('b0682d652840ef50a4115cc77109bedd8c577ccc')),
+          ),
+        );
+      });
+
+      test('prefers last active account over initial account', () async {
+        when(() => storage.readAccounts()).thenAnswer((_) async => accountList);
+        when(() => storage.readLastAccount()).thenAnswer((_) async => 'c73bc81eda746a735de697a0f0cbdb08f96ea3f3');
```
provokateurin, line 237: Verify again

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+                  'organisationScope': 'v2-local',
+                  'phoneScope': 'v2-local',
+                  'profile_enabledScope': 'v2-local',
+                  'roleScope': 'v2-local',
+                  'storageLocation': '/usr/src/nextcloud/data/admin',
+                  'twitterScope': 'v2-local',
+                  'websiteScope': 'v2-local',
+                },
+              },
+            }),
```
provokateurin, line 428: Any reason to not mock it?

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+    storage = _AccountStorageMock();
+
+    repository = AccountRepository(
+      userAgent: 'userAgent',
+      httpClient: _FakeClient(),
+      storage: storage,
+    );
+  });
+
+  final accountList = BuiltList<Credentials>([
```
provokateurin, line 76:
```suggestion
  final credentialsList = BuiltList<Credentials>([
```
The name is a bit confusing throughout the tests

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,329 @@
...
+
+    try {
+      final response = await client.authentication.users.getCurrentUser();
+
+      return Account((b) {
+        b
+          ..client = client
+          ..credentials.replace(credentials)
+          ..credentials.username = response.body.ocs.data.id;
+      });
```
provokateurin, line 255: This is a bit messed up.
The Credentials already contain the username, but you still need to fetch it?
You probably meant for it to contain the loginName.
The loginName has to be used together with the appPassword because they are linked.
Using the username together with the appPassword should not work.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+        expect(
+          repository.accounts,
+          emits(
+            isA<_AccountStream>()
+                .having((e) => e.accounts, 'accounts', hasLength(1))
+                .having((e) => e.active, 'active', equals('b0682d652840ef50a4115cc77109bedd8c577ccc')),
+          ),
+        );
+        verify(() => storage.saveAccounts(any(that: contains(accountList.first)))).called(1);
+        verify(() => storage.saveLastAccount(any(that: equals('b0682d652840ef50a4115cc77109bedd8c577ccc')))).called(1);
```
provokateurin, line 473:
```suggestion
        verify(() => storage.saveLastAccount('b0682d652840ef50a4115cc77109bedd8c577ccc')).called(1);
```
Why so complicated?

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+        when(() => storage.saveLastAccount(any())).thenAnswer((_) async => {});
+        when(() => storage.saveAccounts(any())).thenAnswer((_) async => {});
+
+        await repository.loadAccounts();
+      });
+
+      test('removes active account', () async {
+        when(() => appPassword.deleteAppPassword()).thenAnswer(
+          (_) async => _DynamiteResponseMock<core.AppPasswordDeleteAppPasswordResponseApplicationJson, void>(),
+        );
```
provokateurin, line 489: Verify that it was called once

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+                .having((e) => e.accounts, 'accounts', hasLength(1))
+                .having((e) => e.active, 'active', equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')),
+          ),
+        );
+        verify(() => storage.saveLastAccount(any(that: equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')))).called(1);
+        verify(() => storage.saveAccounts(any(that: equals([accountList[1]])))).called(1);
+      });
+
+      test('rethrows http exceptions as `DeleteCredentialsFailure`', () async {
+        when(() => appPassword.deleteAppPassword()).thenThrow(http.ClientException(''));
```
provokateurin, line 505: Verify again

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+        expect(
+          repository.accounts,
+          emits(
+            isA<_AccountStream>()
+                .having((e) => e.accounts.map((e) => e.credentials), 'accounts', containsAllInOrder(accountList))
+                .having((e) => e.active, 'active', equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')),
+          ),
+        );
+
+        verify(() => storage.saveLastAccount(any(that: equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')))).called(1);
```
provokateurin, line 560:
```suggestion
        verify(() => storage.saveLastAccount('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')).called(1);
```

packages/neon_framework/packages/account_repository/test/account_storage_test.dart
```diff
@@ -0,0 +1,127 @@
...
+      });
+    });
+
+    group('saveLastAccount', () {
+      test('persists account id to disk', () async {
+        when(() => lastAccount.setString(any())).thenAnswer((_) async => true);
+
+        await storage.saveLastAccount('accountID');
+
+        verify(() => lastAccount.setString(any(that: equals('accountID')))).called(1);
```
provokateurin, line 115:
```suggestion
        verify(() => lastAccount.setString('accountID')).called(1);
```

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
+
+        expect(
+          repository.accounts,
+          emits(
+            isA<_AccountStream>()
+                .having((e) => e.accounts, 'accounts', hasLength(1))
+                .having((e) => e.active, 'active', equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')),
+          ),
+        );
+        verify(() => storage.saveLastAccount(any(that: equals('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')))).called(1);
```
provokateurin, line 500:
```suggestion
        verify(() => storage.saveLastAccount('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')).called(1);
```

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,564 @@
...
```
provokateurin, line 520:
```suggestion
        verify(() => storage.saveLastAccount('c73bc81eda746a735de697a0f0cbdb08f96ea3f3')).called(1);
```

packages/neon_framework/packages/account_repository/test/account_storage_test.dart
```diff
@@ -0,0 +1,127 @@
...
+    accounts = _SingleValueStoreMock();
+    lastAccount = _SingleValueStoreMock();
+
+    storage = AccountStorage(
+      accountsPersistence: accounts,
+      lastAccountPersistence: lastAccount,
+    );
+  });
+
+  final accountList = [
```
provokateurin, line 29:
```suggestion
  final credentialsList = [
```

packages/neon_framework/packages/account_repository/test/account_storage_test.dart
```diff
@@ -0,0 +1,127 @@
...
+    }),
+    Credentials((b) {
+      b
+        ..serverURL = Uri.https('other-serverUrl')
+        ..username = 'username'
+        ..password = 'password';
+    }),
+  ];
+
+  final serializedAccounts = BuiltList<String>([
```
provokateurin, line 44:
```suggestion
  final serializedCredentials = BuiltList<String>([
```

packages/neon_framework/packages/account_repository/test/account_storage_test.dart
```diff
@@ -0,0 +1,127 @@
...
+void main() {
+  late SingleValueStore accounts;
+  late SingleValueStore lastAccount;
+  late AccountStorage storage;
+
+  setUp(() {
+    registerFallbackValue(_FakeBuiltList());
+
+    accounts = _SingleValueStoreMock();
+    lastAccount = _SingleValueStoreMock();
```
provokateurin, line 21:
```suggestion
    accountsStore = _SingleValueStoreMock();
    lastAccountStore = _SingleValueStoreMock();
```
These names are really confusing when reading the tests

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
provokateurin: Maybe Credentials should just have the loginName and Account should then have the additional username field

packages/neon_framework/packages/account_repository/test/models/credentials_test.dart
```diff
@@ -0,0 +1,142 @@
...
+        equals('b0682d652840ef50a4115cc77109bedd8c577ccc'),
+      );
+    });
+
+    test('creates humanReadableID', () {
+      expect(
+        createCredentials(
+          serverURL: Uri.https('example.com'),
+          username: 'JohnDoe',
+        ).humanReadableID,
```
provokateurin, line 85: Extract this into a variable like the other cases below

provokateurin (changes requested):
I only reviewed the account_repository so far.
I'll review the neon_framework changes when the existing comments have been sorted out.
IMO you should add a verify for all the API calls that are made in the tests.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
Leptopoda: I agree and I have a wip branch where I implement this.
The issue is that we want to give it `core.$Client` as a constructor parameter for easier mocking.
The core capabilities require an authenticated user while this status check needs to be done before we are authenticated.

This means we would need to insert the `server_repository` with an unauthenticated `core.$Client` (and throw when one tries to access other methods).
Once the user is logged in we would need to rebuild that repository with an authenticated client.
When a user then tries to log in to a different server we would again need a different client.

I just think that this is to much state specific information in the repositroy and therefore leave it in here for now.
It isn't that much code duplication if we also add the same method to the `server_repository`.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,329 @@
...
+
+  /// Whether accounts are logged in.
+  bool get hasAccounts {
+    return _accounts.hasValue && _accounts.value.accounts.isNotEmpty;
+  }
+
+  /// Retrieves the account associated with the given [accountID].
+  Account? accountByID(String accountID) {
+    return _accounts.value.accounts[accountID];
+  }
```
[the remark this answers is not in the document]
Leptopoda: I don't think this makes sense.
We also receive the accountID from push notifications or importing settings, where we can't be sure that the account is still available.

Yes; we can add a callback to check that (basically checking that the key is present in the accounts map), but that would also be tedious.

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
Leptopoda: This is already the parameter name of the http client backing the neon client.

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
Leptopoda: This is already the parameter name of the http client backing the neon client.

packages/neon_framework/packages/account_repository/lib/src/utils/authentication_client.dart
Leptopoda: Not if you ignore the rule :P
In this case I ignored the rule so the client being called is still `client.xxx.core` instead of `client.xxx.coreClient`.
Having the field named core would conflict with the import prefix.

This is also needed for our `find_untested_apis.sh` script to funciton.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: It would be a lot of mocks (OCS, OCSdata, UserDetails, ...)
Should I mock it nevertheless? Now that I think about it 3 Mocks would probably be cleaner than these 50 lines.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
Leptopoda: This makes sense. Same as renaming the passowrd field to `appPassowrd`.
I'll change this in a followup as it would make a lot of changes in the neon code and requires thorough reviewing to make sure the display-name isn't used in a place, where the user id is expected.

This followup can also remove the `Account.password` helper method forcing consumers to access it through the credentials (`Account.credentials.password`).

Leptopoda: FYI: This is again behavior copied from the current code

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
provokateurin: Capabilities doesn't need to be authenticated. It is only possible that it will return different data when logged in vs. not.

provokateurin: But we probably rely on some authenticated capability, so it doesn't matter.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: This is the setup part of the logout code.
1. it is a test setup so it should expect or validate anything
2. it is the logout flow so it should test the account loading

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
provokateurin: Yes that would better

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,329 @@
...
+      serverURL: serverURL,
+    );
+
+    try {
+      final initResponse = await client.authentication.clientFlowLoginV2.init();
+
+      final pollBody = core.ClientFlowLoginV2PollRequestApplicationJson((b) {
+        b.token = initResponse.body.poll.token;
+      });
+      _flow = (client, pollBody);
```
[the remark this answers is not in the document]
Leptopoda: My main concern is that we're going to build a new client for every poll make (currently once a second).
Creating a client isn't as expensive as it was beccause we now reuse the base client but it still requires creating:
- new NeonHttpClient
    - new BaseHeaderInterceptor
    - new CSRFInterceptor
- new NextcloudClient
    - new nextcloud core client

Sending the request dierectly isn't a good option as it wouldn't set the useragent or timeout.

Leptopoda:
I've dropped the neon_framework changes for easier reviewability.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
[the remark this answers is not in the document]
provokateurin: > Sending the request dierectly isn't a good option as it wouldn't set the useragent or timeout.

Not if the NeonHttpClient is used

Can we instead create the client for this outside the repository, so both methods just get it passed and we only need a single instance that isn't stored in the repository?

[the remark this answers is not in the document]
provokateurin: And how about returning the flow instead of storing it in the repository at all? Having this hidden state feels weird to me, it should be more explicit by returning and passing it again.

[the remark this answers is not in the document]
Leptopoda: I've changed it for now (build a new client and return the token).
This isn't any hotpath in the app, and we don't expect users to only use the login flow :)

We can still change it later on.

Leptopoda:
>  IMO you should add a verify for all the API calls that are made in the tests.

Thanks for mentioning. I've added them everywhere.

packages/neon_framework/packages/account_repository/lib/src/utils/authentication_client.dart
provokateurin: You can prefix it with `$` which is the alternative for these cases (also happened in notes and notifications).

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
provokateurin: Then `neonHttpClient`? I find just `neonClient` harder to understand

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
provokateurin: I don't understand what you mean :thinking:

packages/neon_framework/packages/account_repository/pubspec.yaml
```diff
@@ -5,6 +5,7 @@ publish_to: none
 
 environment:
   sdk: ^3.0.0
+  flutter: ^3.22.0
```
provokateurin, line 8: Intentional and necessary?
I thought repositories should be pure Dart.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -466,46 +477,50 @@ void main() {
...
     group('logOut', () {
       setUp(() async {
-        when(() => storage.readAccounts()).thenAnswer((_) async => accountList);
+        when(() => storage.readAccounts()).thenAnswer((_) async => credentialsList);
         when(() => storage.saveLastAccount(any())).thenAnswer((_) async => {});
         when(() => storage.saveAccounts(any())).thenAnswer((_) async => {});
 
         await repository.loadAccounts();
+
+        resetMocktailState();
```
provokateurin, line 496: Why is this necessary?

packages/neon_framework/packages/account_repository/pubspec.yaml
Leptopoda: the repo depends on `neon_framework` for the storages. This pulls in `intl` which requires flutter.
I'm working on a separate storage layer that does not pull this in.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: Because the `await repository.loadAccounts();` will call some storage methods.
If we were to verify storage access in a test, these initial calls would throw off the call count.

This is similar to the `resetFixtures` call we have for the nextcloud package.

packages/neon_framework/packages/account_repository/pubspec.yaml
provokateurin: In this PR or separately?

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
provokateurin: Ah makes sense :+1:

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: My bad. The github UI displayed the comment in the wrong place after force pushing so I got confused

packages/neon_framework/packages/account_repository/pubspec.yaml
Leptopoda: Sparate one.
CI still fails to bootstrap :(

Leptopoda:
Please review my latest changes.
I added a testing library so our `createModel` can be reused by other tests

tool/setup.sh
provokateurin, line 1: Please make a separate commit for this fix

tool/find-untested-neon-apis.sh
provokateurin, line 1: Please make a separate commit for this as well

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,325 @@
...
+part 'account_storage.dart';
+
+/// {@template account_failure}
+/// A base failure for the account repository failures.
+/// {@endtemplate}
+sealed class AccountFailure with EquatableMixin implements Exception {
+  /// {@macro account_failure}
+  const AccountFailure(this.error);
+
+  /// The error which was caught.
```
provokateurin, line 24:
```suggestion
  /// The error which was thrown.
```

packages/neon_framework/packages/account_repository/lib/src/utils/http_client_builder.dart
```diff
@@ -0,0 +1,52 @@
...
+    credentials.serverURL,
+    loginName: credentials.username,
+    password: credentials.password,
+    appPassword: credentials.password,
+    httpClient: neonHttpClient,
+  );
+}
+
+/// Builds an unauthenticated [NextcloudClient] for the given [serverURL].
+NextcloudClient buildUnAuthenticatedClient({
```
provokateurin, line 36:
```suggestion
NextcloudClient buildUnauthenticatedClient({
```

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,325 @@
...
+  /// Logs out the user from the server.
+  ///
+  /// May throw a [DeleteCredentialsFailure].
+  Future<void> logOut(String accountID) async {
+    final value = _accounts.value;
+
+    Account? account;
+    final accounts = value.accounts.rebuild((b) {
+      account = b.remove(accountID);
+    });
```
provokateurin, line 284: If the account is null then the whole method can just return early, no?

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -0,0 +1,325 @@
...
+  ///
+  /// Not yet implemented see: https://github.com/nextcloud/neon/issues/2177
+  Future<void> deleteAccount() async {
+    throw UnimplementedError();
+  }
+
+  /// Sets the active `account` to [accountID].
+  ///
+  /// Throws an [ArgumentError] when the id is not in the accounts map.
+  Future<void> switchAccount(String accountID) async {
```
provokateurin, line 315: Wouldn't it make sense to pass the entire Account and then use the ID to look it up internally?
Any user of the repository already handles the Account, so having to pass the ID instead feels a little off.

packages/neon_framework/packages/account_repository/lib/src/account_storage.dart
provokateurin, line 1: Why is this called AccountStorage and readAccount, saveAccounts etc. while it really only handles Credentials?
As a consumer this would be quite unexpected since I'd see readAccounts and expect it to return some kind of Iterable<Account> and not Iterable<Credentials>

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,584 @@
...
+
+      expect(repository.accountByID('invalid'), isNull);
+      expect(repository.accountByID(credentialsList.first.id), isNotNull);
+    });
+
+    test('isLogInQRCode', () {
+      expect(
+        AccountRepository.isLogInQRCode(url: 'nc://login/user:JohnDoe&password:super_secret&server:example.com'),
+        isTrue,
+      );
```
provokateurin, line 164: Also add a test for the path regex

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,584 @@
...
+
+        verify(() => storage.readLastAccount()).called(1);
+      });
+
+      test('prefers last active account over initial account', () async {
+        when(() => storage.readAccounts()).thenAnswer((_) async => credentialsList);
+        when(() => storage.readLastAccount()).thenAnswer((_) async => credentialsList[1].id);
+        await repository.loadAccounts(
+          rememberLastUsedAccount: true,
+          initialAccount: credentialsList.first.id,
```
provokateurin, line 276: Should initialAccount rather be called fallbackAccount?
IMO this combination of inputs to the method makes no sense, unless the initialAccount is only a fallback, which it is.
Both parameters are imperative about using a specific account, but it isn't clear which takes precedence.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,584 @@
...
+                .having((e) => e.active, 'active', equals(accountsList[1])),
+          ),
+        );
+
+        verify(() => storage.readLastAccount()).called(1);
+      });
+    });
+
+    group('getServerStatus', () {
+      test('removes active account', () async {
```
provokateurin, line 293: Wrong test name

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,584 @@
...
+    group('loginFlowInit', () {
+      test('returns login endpoint and token', () async {
+        when(() => clientFlowLoginV2.init()).thenAnswer((_) async {
+          return DynamiteResponse(
+            200,
+            core.LoginFlowV2((b) {
+              b.login = 'https://login_url';
+              b.poll
+                ..token = 'token'
+                ..endpoint = 'token';
```
provokateurin, line 327:
```suggestion
                ..endpoint = 'endpoint';
```
If the implementation accidentally returned the endpoint instead of the token field this would go unnoticed.

packages/neon_framework/packages/account_repository/test/utils/http_client_builder_test.dart
```diff
@@ -0,0 +1,95 @@
...
+
+    verify(
+      () => storageMock.cookieStore(
+        accountID: any(named: 'accountID', that: equals(credentials.id)),
+        serverURL: any(named: 'serverURL', that: equals(Uri.https('serverURL'))),
+      ),
+    ).called(1);
+  });
+
+  test('buildUnAuthenticatedClient', () {
```
provokateurin, line 69:
```suggestion
  test('buildUnauthenticatedClient', () {
```

provokateurin (changes requested):
In many places you are using `await expectLater()`, but it is unnecessary because you already match using `completion()`.

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: I'm not sure if `fallback` makes sense.
IMO a fallback value is used when no other value is available or the available value is invalid.
In this case the fallback would be the first account available `active = accounts.keys.firstOrNull;`.

provokateurin: Right, should we even take it into account then?
Maybe this functionality should be done through switchAccount instead if necessary.

Leptopoda: Let's leave it for now and see how the other refactors go.
I explicitly treid to mimic the current `AccountsBloc` and where possible would try similar with other repository migrations.
I fear that we'll try to rewrite everything at once.

I can open an issue for this if you want.

provokateurin (approved):

Leptopoda:
> For the [completes] and [completion] matchers, as well as [throwsA] and related matchers when they're matched against a [Future], the returned future completes when the matched future completes.

provokateurin:
Oh but then there are some places were you use completion() without await expectLater()

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
```diff
@@ -0,0 +1,602 @@
...
+    group('getServerStatus', () {
+      test('fetches the server status', () async {
+        final mockResponse = _DynamiteResponseMock<core.Status, void>();
+        when(() => coreClient.getStatus()).thenAnswer((_) async => mockResponse);
+        when(() => mockResponse.body).thenReturn(_FakeStatus());
+
+        expect(
+          repository.getServerStatus(Uri()),
+          completion(isA<core.Status>()),
+        );
```
provokateurin, line 306: For example

provokateurin (changes requested):

packages/neon_framework/packages/account_repository/test/account_repository_test.dart
Leptopoda: You are right.
This wasn't an issue because there is no method call we try to validate after the async suspension.

I'll change them nevertheless to allow future alterations.

provokateurin (approved):

PR #2413, opened 2024-08-21
refactor(cookie_store): remove universal_io dependency.

Leptopoda:
The `Cookie` does not depend on the dart vm and is therefore also usable on web.
To verify this, I've also enabled unit tests on chrome.

### DIFF
--- a/packages/cookie_store/dart_test.yaml
+++ b/packages/cookie_store/dart_test.yaml
@@ -1,0 +1,10 @@
+platforms:
+  - vm
+  - chrome
+
+define_platforms:
+  chromium:
+    name: Chromium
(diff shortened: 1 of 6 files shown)


provokateurin:
So you can use it on web, even though it comes from dart:io?

Leptopoda:
Yes. Most of dartio can also be used on web.
If you take a look at `universal_io` you can see that only a few classes are not available (`HttpClient`, `BytesBuilder`, `InternetAddress`, `Platform`).

https://github.com/dint-dev/universal_io/blob/master/lib/src/_exports_in_browser.dart

provokateurin (approved):

PR #2422, opened 2024-08-25
ci: Use fdroidserver patch for publish workflow

provokateurin:
https://github.com/nextcloud/neon/issues/18

To test out the patch https://gitlab.com/fdroid/fdroidserver/-/merge_requests/1511 made by @wrenix

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,6 +1,7 @@
 name: Publish
 
 on:
+  pull_request:
   schedule:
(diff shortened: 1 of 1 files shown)


provokateurin:
I'm not sure if the remaining problem is related to the first problem or not

provokateurin:
Makes sense. Thanks for the patch btw :+1:

provokateurin:
I doubt it, it says it is resigning the files. I also wouldn't know why this would cause the files to be missing.

PR #2428, opened 2024-08-27
fix(dynamite)!: Do not assume subtypes of someOfs are always nullable

provokateurin:
I hit this problem with the provisioning_api change you can see in the diff.
Resolving the subtypes of someOfs assumed they would be always nullable which is reasonable for the records.
However that does not work in case we optimize the type into a single Dart type.
If that happens it has to follow different rules and only be optionally nullable depending on then nullablility of the someOf schema itself.

Pretty nasty bug especially given the single place where it actually failed and the low chance to hit that. There was a change in Nextcloud 30 which can now actually return null for those fields (ofc I first thought this was an issue of the spec).

### DIFF
--- a/packages/nextcloud/lib/src/api/provisioning_api/provisioning_api.openapi.dart
+++ b/packages/nextcloud/lib/src/api/provisioning_api/provisioning_api.openapi.dart
@@ -6232,11 +6232,11 @@ typedef UserDetailsQuota_Quota = ({num? $num, String? string});
 
 @BuiltValue(instantiable: false)
 sealed class $UserDetailsQuotaInterface {
-  num get free;
+  num? get free;
   UserDetailsQuota_Quota? get quota;
-  num get relative;
-  num get total;
-  num get used;
+  num? get relative;
+  num? get total;
+  num? get used;
 
   /// Rebuilds the instance.
   ///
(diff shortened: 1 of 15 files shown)


packages/nextcloud/lib/src/api/provisioning_api/provisioning_api.openapi.dart
```diff
@@ -6232,11 +6232,11 @@ typedef UserDetailsQuota_Quota = ({num? $num, String? string});
 
 @BuiltValue(instantiable: false)
 sealed class $UserDetailsQuotaInterface {
-  num get free;
+  num? get free;
```
Leptopoda, line 6235: I disagree with this change.
```json
            "UserDetailsQuota": {
                "type": "object",
                "properties": {
                    "free": {
                        "anyOf": [
                            {
                                "type": "number",
                                "format": "double"
                            },
                            {
                                "type": "integer",
                                "format": "int64"
                            }
                        ]
                    },
                    ...
                  },
                },
```

I don't see where null is allowed here.
The value has to be either a _non-nullable_ `integer` or a _non-nullable_ `double`. The union of both in dart is a _non-nullable_ `num`.

IMO the upstream specs are broken.

provokateurin: Ah yes, I was probably thinking again that anyOf means >=0 but it means >=1 matching schemas.
I'll check if this is also something that is wrong in the openapi-extractor.

provokateurin:
I can not reproduce the original bug anymore and judging from the code the specifications and our generated code are correct.

PR #2431, opened 2024-08-27
fix(nextcloud): Use correct WebDAV namespaces for oc:checksum, oc:share-type and oc:tags

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2410

Just some stupid copy-pasta. To be fair the code in that file is not really readable because of all the annotations, fields and parameters...


Leptopoda:
Can we add a test?

provokateurin:
Done

Leptopoda:
I'd rather have a separate test case for the checksum, tags and share. Therefore when one fails it's easier to identify the root cause.

Please also make a separate commit for the checksum commit (4bed78d012f4c51bda9f3ef11c04056b665416d3) as I'd like to discuss something specific to that without blocking this fix.

Leptopoda (approved):

PR #2436, opened 2024-08-29
feat(nextcloud): Add support for checksums for uploading WebDAV files

provokateurin:
Extracted from https://github.com/nextcloud/neon/pull/2431

### DIFF
--- a/packages/nextcloud/lib/src/api/webdav/webdav_client.dart
+++ b/packages/nextcloud/lib/src/api/webdav/webdav_client.dart
@@ -10,6 +10,18 @@ import 'package:nextcloud/src/api/webdav/webdav.dart';
 import 'package:nextcloud/utils.dart';
 import 'package:universal_io/io.dart' show File, FileStat;
 
+/// The algorithms supported for the oc:checksum prop and OC-Checksum header.
+const supportedChecksumAlgorithms = ['md5', 'sha1', 'sha256', 'sha3-256', 'adler32'];
+
+/// The pattern of supported checksum algorithms.
+///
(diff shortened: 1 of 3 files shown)


packages/nextcloud/lib/src/api/webdav/webdav_client.dart
```diff
@@ -571,13 +586,21 @@ class WebDavClient {
...
   }) {
     if (lastModified != null) {
       request.headers['X-OC-Mtime'] = lastModified.secondsSinceEpoch.toString();
     }
     if (created != null) {
       request.headers['X-OC-CTime'] = created.secondsSinceEpoch.toString();
     }
+    if (checksum != null) {
+      final parts = checksum.split(':');
+      if (parts.length != 2 || !supportedChecksumAlgorithms.contains(parts[0].toLowerCase()) || parts[1].isEmpty) {
```
Leptopoda, line 599: This is hard to read.
Please use a string parsing package like string_scanner (https://pub.dev/packages/string_scanner).

Iirc we already depend on it for the http date parsing.

Leptopoda (changes requested):

packages/nextcloud/lib/src/api/webdav/webdav_client.dart
Leptopoda: You are right. As this is only for validation a regex would probably be the best option.

Please also document it in the doc comment.

provokateurin:
The WebDavClient parameters (order) and doc comments are quite messy. I'll give it a cleanup after this PR.

Leptopoda (approved):

PR #2437, opened 2024-08-29
ci: Use patched fdroidserver to fix publish workflow

provokateurin:
Same as https://github.com/nextcloud/neon/pull/2435, but from me :P


provokateurin:
@wrenix awesome it works!! Also tested with an F-Droid client

provokateurin:
Thank you so much @wrenix :heart:

PR #2450, opened 2024-08-30
chore(ci): for android publish use master fdroidserver again

wrenix:
https://gitlab.com/fdroid/fdroidserver/-/merge_requests/1512
is in master now


Leptopoda (approved):

provokateurin (changes requested):
I think we should pin it in some way, either using a git commit hash or tag.

provokateurin (approved):
Once a version with the fix is released we should switch to that, but this is good for now :+1:

PR #2454, opened 2024-08-31
feat(neon_http_client): Add AuthenticationInterceptor

provokateurin:
Closes https://github.com/nextcloud/neon/issues/365

Quite simple to implement with our interceptors :smile:

Every Account gets it's own NeonHttpClient and therefore it's own AuthenticationInterceptor, so no separation is required.

I removed the catch handlers because the interceptor needs to be able to throw an exception for requests that need to be blocked.

### DIFF
--- a/packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
+++ b/packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
@@ -4,6 +4,7 @@ import 'package:built_collection/built_collection.dart';
 import 'package:cookie_store/cookie_store.dart';
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
+import 'package:neon_http_client/src/interceptors/authorization_throttling_interceptor.dart';
 import 'package:neon_http_client/src/interceptors/interceptors.dart';
 import 'package:neon_http_client/src/utils/utils.dart';
 import 'package:universal_io/io.dart';
@@ -47,6 +48,8 @@ final class NeonHttpClient with http.BaseClient {
       if (interceptors != null) {
         builder.addAll(interceptors);
       }
+
+      builder.add(AuthorizationThrottlingInterceptor(baseURL: baseURL));
 
       if (cookieStore != null) {
         builder.add(
--- a/packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
+++ b/packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
@@ -101,7 +101,11 @@ final class NeonHttpClient with http.BaseClient {
           interceptedRequest = await interceptor.interceptRequest(
             request: interceptedRequest,
           );
-        } catch (_, stackTrace) {
+        } catch (error, stackTrace) {
+          if (error is http.ClientException) {
+            rethrow;
+          }
+
           Error.throwWithStackTrace(
             InterceptionException('Failed to intercept request', request.url),
             stackTrace,
@@ -131,7 +135,11 @@ final class NeonHttpClient with http.BaseClient {
             response: interceptedResponse,
             url: url,
           );
-        } catch (_, stackTrace) {
+        } catch (error, stackTrace) {
+          if (error is http.ClientException) {
+            rethrow;
+          }
+
           Error.throwWithStackTrace(
             InterceptionException('Failed to intercept response', request.url),
             stackTrace,
--- a/packages/neon_framework/packages/neon_http_client/test/neon_http_client_test.dart
+++ b/packages/neon_framework/packages/neon_http_client/test/neon_http_client_test.dart
@@ -2,6 +2,7 @@ import 'package:cookie_store/cookie_store.dart';
 import 'package:http/http.dart';
 import 'package:http/testing.dart';
 import 'package:mocktail/mocktail.dart';
+import 'package:neon_http_client/src/interceptors/authorization_throttling_interceptor.dart';
 import 'package:neon_http_client/src/interceptors/interceptors.dart';
 import 'package:neon_http_client/src/neon_http_client.dart';
 import 'package:nextcloud/nextcloud.dart';
@@ -46,8 +47,8 @@ void main() {
       );
 
       expect(
-        client.interceptors.first,
-        isA<BaseHeaderInterceptor>(),
+        client.interceptors,
+        contains(isA<BaseHeaderInterceptor>()),
       );
     });
 
@@ -60,8 +61,22 @@ void main() {
       );
 
       expect(
+        client.interceptors,
+        contains(isA<CookieStoreInterceptor>()),
+      );
+    });
+
+    test('adds authorization throttling interceptor before cookie store', () {
+      final cookieStore = _MockCookieStore();
+
+      client = NeonHttpClient(
+        baseURL: Uri(),
+        cookieStore: cookieStore,
+      );
+
+      expect(
         client.interceptors.first,
-        isA<CookieStoreInterceptor>(),
+        isA<AuthorizationThrottlingInterceptor>(),
       );
     });
 
--- a/packages/neon_framework/packages/neon_http_client/test/neon_http_client_test.dart
+++ b/packages/neon_framework/packages/neon_http_client/test/neon_http_client_test.dart
@@ -4,6 +4,7 @@ import 'package:http/testing.dart';
 import 'package:mocktail/mocktail.dart';
 import 'package:neon_http_client/src/interceptors/interceptors.dart';
 import 'package:neon_http_client/src/neon_http_client.dart';
+import 'package:nextcloud/nextcloud.dart';
 import 'package:test/test.dart';
 
 class _MockCookieStore extends Mock implements CookieStore {}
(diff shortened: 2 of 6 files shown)


provokateurin:
I'm not so happy with the name of the interceptor, let me know if there is something more fitting.

packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
```diff
@@ -97,16 +102,9 @@ final class NeonHttpClient with http.BaseClient {
...
-        } catch (_, stackTrace) {
-          Error.throwWithStackTrace(
-            InterceptionException('Failed to intercept request', request.url),
-            stackTrace,
-          );
-        }
+        interceptedRequest = await interceptor.interceptRequest(
+          request: interceptedRequest,
+        );
       }
```
Leptopoda, line 108: I'm not a fan of removing the try catch.
An http client should wrap all errors into `http.ClientException`s. In the unlikely case, a Interceptor trows I still want the exceptions to be wrapped as we only catch `http.ClientException`s down the road (see the request manager).

We either need to find a different solution or just rethrow the exception un-wrapped.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,37 @@
...
+final class AuthenticationInterceptor implements HttpInterceptor {
+  var _blocked = false;
+
+  @override
+  bool shouldInterceptRequest(BaseRequest request) {
+    return (request.headers['authorization'] ?? '').isNotEmpty && _blocked;
+  }
+
+  @override
+  FutureOr<BaseRequest> interceptRequest({required BaseRequest request}) {
```
Leptopoda, line 18:
```suggestion
  Never interceptRequest({required BaseRequest request}) {
```

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,37 @@
...
+import 'package:nextcloud/nextcloud.dart';
+
+@internal
+final class AuthenticationInterceptor implements HttpInterceptor {
+  var _blocked = false;
+
+  @override
+  bool shouldInterceptRequest(BaseRequest request) {
+    return (request.headers['authorization'] ?? '').isNotEmpty && _blocked;
+  }
```
Leptopoda, line 15:
```suggestion
  bool shouldInterceptRequest(BaseRequest request) {
  	final authorization = request.headers['authorization'];
    return authorization != null  && authorization.isNotEmpty && _blocked;
  }
```

You can also use pattern matching if you prefer to.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,37 @@
...
+  @override
+  FutureOr<BaseRequest> interceptRequest({required BaseRequest request}) {
+    throw DynamiteStatusCodeException(Response('', 401));
+  }
+
+  @override
+  bool shouldInterceptResponse(StreamedResponse response) {
+    if ((response.request?.headers['authorization'] ?? '').isEmpty) {
+      return false;
+    }
```
Leptopoda, line 26: Please also use null aware syntax here

Leptopoda (changes requested):

packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
```diff
@@ -64,9 +65,13 @@ final class NeonHttpClient with http.BaseClient {
...
-      builder.add(
-        CSRFInterceptor(client: baseClient, baseURL: baseURL),
-      );
+      builder
+        ..add(
+          CSRFInterceptor(client: baseClient, baseURL: baseURL),
+        )
+        ..add(
+          AuthenticationInterceptor(),
+        );
```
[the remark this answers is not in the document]
provokateurin: The nextcloud package should not ever use neon_http_client.
This is an HTTP client specifically built for the framework.
I get your point for testing, but everything should be tested as close to reality as possible, so including this interceptor at runtime.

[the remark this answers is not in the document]
Leptopoda: Then how do you want to handle the persistence of cookies that is needed for the nextcloud tests?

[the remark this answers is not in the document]
provokateurin: As we tell everyone who wants to use the nextcloud package with cookies: Provide your own custom HTTP client.
Since we have cookie_store this is easy to implement and we only need it for testing.
Relying on stuff from the framework is not ok for this purpose.

packages/neon_framework/packages/neon_http_client/test/neon_http_client_test.dart
```diff
@@ -137,7 +138,26 @@ void main() {
         ).called(1);
       });
 
-      test('rethrows errors as InterceptionFailure', () async {
+      test('does not rethrow http.ClientExceptions', () async {
```
Leptopoda, line 156:
```suggestion
      test('does rethrow http.ClientExceptions', () async {
```

packages/neon_framework/packages/neon_http_client/lib/src/neon_http_client.dart
```diff
@@ -101,7 +101,11 @@ final class NeonHttpClient with http.BaseClient {
           interceptedRequest = await interceptor.interceptRequest(
             request: interceptedRequest,
           );
-        } catch (_, stackTrace) {
+        } catch (error, stackTrace) {
+          if (error is http.ClientException) {
+            rethrow;
+          }
+
```
Leptopoda, line 109: This needs to be documented somewhere.
What about the abstract interceptor class?

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  @override
+  bool shouldInterceptResponse(StreamedResponse response) {
+    final authorization = response.request?.headers['authorization'];
+    if (authorization == null || authorization.isEmpty) {
+      return false;
+    }
+
+    return response.statusCode == 401 || (response.request?.url.path.endsWith('/index.php/login/v2/poll') ?? false);
```
Leptopoda, line 28: 1. Please add the baseURL as a parameter to this class so you can match the real URL.
2. please move the url matching to the interception method. The url there might be more accurate.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,37 @@
...
+import 'package:meta/meta.dart';
+import 'package:neon_http_client/src/interceptors/http_interceptor.dart';
+import 'package:nextcloud/nextcloud.dart';
+
+@internal
+final class AuthenticationInterceptor implements HttpInterceptor {
+  var _blocked = false;
+
+  @override
+  bool shouldInterceptRequest(BaseRequest request) {
```
Leptopoda, line 11: I think this interceptor should only intercept requests to the nextcloud server. Please match the baseURL

Leptopoda (changes requested):
Did you test that this also work together with the nextcloud package?
I think 401 responses are thrown so the response  interceptor wouldn't ever be used.

provokateurin:
> Did you test that this also work together with the nextcloud package?
I think 401 responses are thrown so the response interceptor wouldn't ever be used.

You're thinking on the wrong layer. The response interceptor comes way before the response is parsed in the nextcloud package and an exception might be thrown.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @override
+  Never interceptRequest({required BaseRequest request}) {
+    assert(
+      shouldInterceptRequest(request),
+      'Request should not be intercepted.',
+    );
+
+    throw DynamiteStatusCodeException(Response('', 401));
```
Leptopoda, line 34: maybe the response message should mention that the interceptor blocked the request.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
```diff
@@ -0,0 +1,72 @@
...
+  StreamedResponse interceptResponse({required StreamedResponse response, required Uri url}) {
+    assert(
+      shouldInterceptResponse(response),
+      'Response should not be intercepted.',
+    );
+
+    final request = response.request;
+    if (request == null) {
+      return response;
+    }
```
Leptopoda, line 61: This is already checked above. I don't think there is a need to check it again.

Leptopoda (approved):

Leptopoda:
> You're thinking on the wrong layer. The response interceptor comes way before the response is parsed in the nextcloud package and an exception might be thrown.

Yep. I mixed something up in my head.

> Maybe UnauthorizedInterceptor is a better name? 

I'm not sure about the naming and don't have any opinion on it. What about `CheckAuthenticationInterceptor`?

provokateurin:
> I'm not sure about the naming and don't have any opinion on it. What about `CheckAuthenticationInterceptor`?

It still doesn't really convey what it is doing. I think something like `UnauthorizedThrottlingInterceptor` describes the functionality the best way, but it is still not intuitive. Maybe there simply is no good name for it.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authentication_interceptor.dart
provokateurin: Not really possible since the only parameter is the response and putting this message in the body field is wrong.

PR #2457, opened 2024-09-01
feat(tool,neon_framework): Update logo and use adaptive icons

provokateurin:
Closes https://github.com/nextcloud/neon/issues/447

This was quite some work to get done, but I think it looks way better now:
image
image

It was not possible to split this into multiple commits as everything is coupled tightly.


Leptopoda (commented):
The diff LGTM

The only thing is that nextcloud apps use a gradient for the background. Can you also achieve that? Should be fairly simple.

provokateurin:
Unfortunately this is a lot harder to achieve than simply editing the SVG, so I would like to skip that for now. We can still make an issue about it as I agree that we should have it at some point.

provokateurin:
Ok I will give it a try, but it is probably a lot harder than you'd think.

provokateurin:
Works for adaptive icons:
image
As for the non-adaptive icons it doesn't work because inkscape doesn't work with CSS gradient: `Ignoring unrecognized CSS property: background-image`.

provokateurin:
I just realized there are some mistakes...

Leptopoda:
> As for the non-adaptive icons it doesn't work because inkscape doesn't work with CSS gradient: Ignoring unrecognized CSS property: background-image.

This shouldn't be a problem. Adaptive icons are supported since api 26 (8.0) while our min sdk is currently api 24 (7.0) so this shouldn't really matter.

Thanks for the quick changes; I'll do a review once you got it fixed and I have a bit of time.

provokateurin:
> This shouldn't be a problem. Adaptive icons are supported since api 26 (8.0) while our min sdk is currently api 24 (7.0) so this shouldn't really matter.

I agree. I think everywhere else we also show the icon in blue with transparent background, so there we don't need the gradient either.

> Thanks for the quick changes; I'll do a review once you got it fixed and I have a bit of time.

Done

Leptopoda (approved):
:heart:

PR #2464, opened 2024-09-04
perf(nextcloud_test_presets): Use docker image digests to improve caching

provokateurin:
Locally it helps docker a lot to detect that images didn't change and it should also help in CI to avoid unnecessary build time.
The only downside is that we get more preset updates, but it's not too often and the benefits outweigh this downside.

### DIFF
--- a/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
+++ b/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
@@ -108,7 +108,12 @@ Future<List<Version>> _getServerVersions(http.Client httpClient) async {
       try {
         final tag = result as Map<String, dynamic>;
 
-        final version = Version.parse(tag['name'] as String);
+        final name = tag['name'] as String;
+        if (!name.endsWith('-fpm-alpine')) {
+          continue;
+        }
+
+        final version = Version.parse(name);
         final normalizedVersion = Version(version.major, version.minor, 0);
 
         if (version < core.minVersion) {
--- a/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
+++ b/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
@@ -19,8 +19,8 @@ Future<void> generatePresets() async {
 
   final httpClient = http.Client();
 
-  final serverVersions = await _getServerVersions(httpClient);
-  serverVersions.sort((a, b) => b.compareTo(a));
+  final serverReleases = await _getServerReleases(httpClient);
+  serverReleases.sort((a, b) => b.compareTo(a));
   final apps = await _getApps(appIDs, httpClient);
 
   for (final app in apps) {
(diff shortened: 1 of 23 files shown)


packages/nextcloud/packages/nextcloud_test_presets/lib/src/models/app_release.dart
```diff
@@ -16,14 +16,15 @@ class AppRelease {
...
-  Version? findLatestServerVersion(List<Version> serverVersions) {
-    final compatibleReleases = serverVersions
+  ServerRelease? findLatestServerRelease(List<ServerRelease> serverReleases) {
+    final compatibleReleases = serverReleases
         .where(
-          (serverVersion) =>
-              serverVersion >= minimumServerVersion && serverVersion < maximumServerVersion.incrementMajor(),
+          (serverRelease) =>
+              serverRelease.version >= minimumServerVersion &&
+              serverRelease.version < maximumServerVersion.incrementMajor(),
```
Leptopoda, line 24: With the server release being a separate class this should be a method in there. Maybe something like `bool isInRange(ServerRelease other)`.

packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
```diff
@@ -63,31 +63,32 @@ Future<void> generatePresets() async {
...
-      final release = app.findLatestCompatibleRelease(serverVersion) ??
-          app.findLatestCompatibleRelease(serverVersion, allowUnstable: true) ??
+      final release = app.findLatestCompatibleRelease(serverRelease) ??
+          app.findLatestCompatibleRelease(serverRelease, allowUnstable: true) ??
           app.findLatestRelease();
       buffer.writeln('${app.id.toUpperCase()}_URL=${release.url}');
     }
 
-    File('${serverPresetsDir.path}/${serverVersion.major}.${serverVersion.minor}').writeAsStringSync(buffer.toString());
+    File('${serverPresetsDir.path}/${serverRelease.version.major}.${serverRelease.version.minor}')
```
Leptopoda, line 76: The `"$major.$minor"` thing should also be moved into the class (maybe a getter called preset).

packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
```diff
@@ -63,31 +63,32 @@ Future<void> generatePresets() async {
   }
   serverPresetsDir.createSync();
 
-  for (final serverVersion in serverVersions) {
-    final buffer = StringBuffer()..writeln('SERVER_VERSION=$serverVersion');
+  for (final serverRelease in serverReleases) {
+    final buffer = StringBuffer()..writeln('SERVER_VERSION=${serverRelease.version}@${serverRelease.digest}');
```
Leptopoda, line 67: Also used multiple times.

Mabe a getter called `dockerTag`

Leptopoda (changes requested):

packages/nextcloud/packages/nextcloud_test_presets/lib/src/models/server_release.dart
```diff
@@ -0,0 +1,4 @@
+import 'package:version/version.dart';
+
+/// A release of the Nextcloud server.
+typedef ServerRelease = ({Version version, String digest});
```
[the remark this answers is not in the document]
provokateurin: built_value makes no sense as we never update properties after creation, so final fields are fine and other than a toString() method no serialization is necessary.

Leptopoda (approved):

PR #2471, opened 2024-09-07
fix(dynamite): Escape HTML in docs

provokateurin:
Closes https://github.com/nextcloud/neon/issues/2470

To avoid converting angle brackets inside code blocks the number of preceding back ticks needs to be checked in order to determine if the match is inside a code block or not.


packages/dynamite/lib/src/helpers/docs.dart
```diff
@@ -8,7 +10,16 @@ Iterable<String> escapeDescription(String? description) sync* {
     return;
   }
 
-  final lines = description.split('\n');
+  final htmlEscapedDescription = description.replaceAllMapped(RegExp('<([^>]+)>'), (match) {
+    final codeBlockQuotes = description.split('').whereIndexed((i, char) => char == '`' && i < match.start).length;
```
Leptopoda, line 14:
```suggestion
    final codeBlockQuotes = description.split('').take(match.start).where((i) => char == '`').length;
```

Leptopoda (approved):

PR #2475, opened 2024-09-08
fix(nextcloud)!: Throw DynamiteStatusCodeExceptions for invalid WebDAV status codes

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2307

### DIFF
--- a/packages/neon_framework/packages/files_app/lib/src/utils/task.dart
+++ b/packages/neon_framework/packages/files_app/lib/src/utils/task.dart
@@ -108,7 +108,7 @@ class FilesDownloadTaskMemory extends FilesTaskMemory implements FilesDownloadTa
   });
 
   Future<void> execute(NextcloudClient client) async {
-    final stream = client.webdav.getStream(
+    final stream = await client.webdav.getStream(
       uri,
       onProgress: progressController.add,
     );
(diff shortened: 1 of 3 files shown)


packages/nextcloud/lib/src/webdav/webdav_client.dart
```diff
@@ -355,40 +390,33 @@ class WebDavClient {
...
-            onDone: () {
-              onProgress?.call(1);
-              controller.close();
-            },
-          );
-        },
-        // ignore: avoid_types_on_closure_parameters
-        onError: (Object error) {
-          controller.addError(error);
-        },
```
Leptopoda, line 390: Why did you remove the on error callback?

packages/nextcloud/test/webdav/webdav_test.dart
Leptopoda, line 1: Please add tests for the invalid status code (probably requires a mocked http client)

Leptopoda (changes requested):
I'm not so sure about the first commit.
Maybe we should wait with any refactors in that regard until we've decided on how we tackle other webdav based clients.

packages/nextcloud/lib/src/webdav/webdav_client.dart
provokateurin: This was needed so the error would appear in the stream. The function is now async and the error will happen on the await and not in the stream. You can check the test for this, it still works.

provokateurin:
>  Maybe we should wait with any refactors in that regard until we've decided on how we tackle other webdav based clients.

I think any refactor to support CalDAV and CardDAV will need this even more as the clients get even more different from the generated ones by extending the base DAV client in some form (e.g. polymorphism or some other way)

Leptopoda (approved):

PR #2482, opened 2024-09-12
feat(notifications_push_repository): Init

provokateurin:
https://github.com/nextcloud/neon/issues/577
Depends on https://github.com/nextcloud/neon/pull/2407 (due to an ugly cyclic dependency)

The framework part can probably be improved, but I just wanted to put out the repository so it can get reviewed.

The repository now handles all possible things that could go wrong and retries registering at UnifiedPush or Nextcloud if necessary. This should make push notifications super reliable and usable.


Leptopoda:
Should we land the repository and the migration changes in separate PRs?
Maybe we can land both into a temporary branch that we then merge once both are reviewed.

Both commits are already huge to begin with.

provokateurin:
Hm if we only have 2 PRs I'd say just merge the repository into main and then later the refactor.
A special branch is not really necessary as there are no incremental changes that need to be merged together into main (the repository is standalone after all).

provokateurin:
I removed the framework commit and pushed it to a separate branch. Once merged I'll open a new PR for that.

provokateurin:
I rebased this, but while testing I found an error with the push subscription registration.
You can already review this, but I will have to look the problem so expect some (hopefully minor) changes.

provokateurin:
Uhm Github seems to be slow with processing updates...

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+    required AccountRepository accountRepository,
+    required NotificationsPushStorage storage,
+    required OnMessageCallback onMessage,
+  })  : _accountRepository = accountRepository,
+        _storage = storage,
+        _onMessage = onMessage;
+
+  final AccountRepository _accountRepository;
+  final NotificationsPushStorage _storage;
+  final OnMessageCallback _onMessage;
```
Leptopoda, line 34: I'm not sure about this callback.
Why does this return an unencrypted message? This way, the caller needs to also handle the storage for decryption.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+  var _initialized = false;
+  StreamSubscription<BuiltList<Account>>? _accountsListener;
+  late final RSAPrivateKey _privateKey;
+  late BuiltList<Account> _accounts;
```
Leptopoda, line 38: Please remove this interlocked state and just fetch it from the account_repo on demand.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+  Future<void> initialize() async {
+    if (_initialized) {
+      return;
+    }
+
+    _accounts = (await _accountRepository.accounts.first).accounts;
+    _accountsListener = _accountRepository.accounts.skip(1).map((e) => e.accounts).listen((accounts) async {
+      _accounts = accounts;
+      await _updateSubscriptions();
+    });
```
Leptopoda, line 77:
```suggestion
    _accountsListener = _accountRepository.accounts.map((e) => e.accounts).listen((accounts) async {
	// maybe pass the accounts to the update function if you want to save the lookup.
      await _updateSubscriptions();
    });
```

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+            b.remove(accountID);
+          } else {
+            _log.finer('Saving subscription for $accountID');
+            b[accountID] = subscription;
+          }
+        },
+      ),
+    );
+  }
+}
```
Leptopoda, line 243: This would fit better into the storage layer, but we can also keep it here until #1538 has landed.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+    String accountID,
+    Account? account,
+    PushSubscription subscription,
+  ) async {
+    if (subscription.pushDevice != null) {
+      _log.finer('Unregistering $accountID from Nextcloud');
+
+      try {
+        await account?.client.notifications.push.removeDevice();
+      } on DynamiteApiException catch (error) {
```
Leptopoda, line 215:
```suggestion
      } on http.ClientException catch (error) {
```

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+            ..proxyServer = '$endpoint#',
+        ),
+      );
+
+      return subscription.rebuild(
+        (b) => b
+          ..endpoint = endpoint
+          ..pushDevice.replace(response.body.ocs.data),
+      );
+    } on DynamiteApiException catch (error) {
```
Leptopoda, line 198:
```suggestion
    } on http.ClientException catch (error) {
```

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+            ..devicePublicKey = _privateKey.publicKey.toFormattedPEM()
+            ..proxyServer = '$endpoint#',
+        ),
+      );
+
+      return subscription.rebuild(
+        (b) => b
+          ..endpoint = endpoint
+          ..pushDevice.replace(response.body.ocs.data),
+      );
```
Leptopoda, line 197: This effectively builds a new subscription (every attribute is set) so I don't see the point of rebuilding an old one over just creating one.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+
+      try {
+        await account?.client.notifications.push.removeDevice();
+      } on DynamiteApiException catch (error) {
+        _log.warning('Failed to unregister $accountID at Nextcloud', error);
+      }
+    }
+
+    // Remove the push device either way, as we need to register for a new one later
+    return subscription.rebuild((b) => b.pushDevice = null);
```
Leptopoda, line 221: Why do we still keep the endpoint?
What value does it have to us?

If we don't need it, just unregister from the NC endpoint and remove the subscription entirely.

packages/neon_framework/packages/notifications_push_repository/lib/src/utils/encryption.dart
```diff
@@ -0,0 +1,46 @@
...
+      PushNotification.fromEncrypted(
+        data,
+        accountID,
+        privateKey,
+      ),
+    );
+  }
+
+  return builder.build();
+}
```
Leptopoda, line 46: I don't like this being here.
The only class accessing the `NotificationsPushStorage` should be the `NotificationsPushRepository`.

If we decrypt the notifications in the repo we should be able to remove this helper.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
```diff
@@ -0,0 +1,243 @@
...
+
+  final AccountRepository _accountRepository;
+  final NotificationsPushStorage _storage;
+  final OnMessageCallback _onMessage;
+  var _initialized = false;
+  StreamSubscription<BuiltList<Account>>? _accountsListener;
+  late final RSAPrivateKey _privateKey;
+  late BuiltList<Account> _accounts;
+  String? _selectedDistributor;
+  String? _previousDistributor;
```
Leptopoda, line 40: Please explain to me why we need both distributors.
Isn't it enough to just store one?

Leptopoda (changes requested):
Thanks for the work.
I'm not a big fan of the code style in the repo.
Many methods in there do not take any parameters and rely on the global state of the repo. I've already outlined most of my concerns, and maybe fixing them is enough.

packages/neon_framework/packages/notifications_push_repository/lib/src/models/push_notification.dart
```diff
@@ -0,0 +1,63 @@
...
+
+  /// The subject of this notification.
+  notifications.DecryptedSubject get subject;
+}
+
+final Serializers _serializers = (Serializers().toBuilder()
+      ..add(notifications.DecryptedSubject.serializer)
+      ..add(PushNotification.serializer)
+      ..addPlugin(StandardJsonPlugin()))
+    .build();
```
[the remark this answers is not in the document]
provokateurin: This was just copied from the framework. I can change it :+1:

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_storage.dart
```diff
@@ -0,0 +1,63 @@
+import 'dart:convert';
```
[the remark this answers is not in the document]
provokateurin: I didn't notice this was done in the account_repository. I tried to implement it in the same way.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
provokateurin: You mean encrypted message?
I really wanted to change this, but the problem is that for this callback to work it needs to be a top-level or static method.
This makes it really painful as we can not modify the data (e.g. decrypt and parse) before passing it on.
I don't see how this can be done in a better way than it is right now :/

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
provokateurin: I was thinking about that, not sure why I didn't do it in the end.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
provokateurin: The endpoint shouldn't be set here as it is already set before, that's a bug here.

provokateurin: The endpoint shouldn't be set again here as it is already set before.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
provokateurin: Because it is separate from the Nextcloud subscription.
The endpoint can still be valid while we had to remove/update the Nextcloud subscription.

packages/neon_framework/packages/notifications_push_repository/lib/src/utils/encryption.dart
provokateurin: See above, this is not possible and we need this static method unfortunately.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
provokateurin: We need to be able to check if the new one is a different one.
I'll refactor this to pass a parameter to _updateSubscriptions() instead when calling changeDistributor().

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
Leptopoda: Why not return the relevant data (pushDevice) and let the caller build the subscription?
That would also clean up the other method.

packages/neon_framework/packages/notifications_push_repository/lib/src/notifications_push_repository.dart
Leptopoda: Meh, I thought that might be the case.
Let's keep it for now, although I really dislike having to do this :(

packages/neon_framework/packages/notifications_push_repository/lib/src/utils/encryption.dart
Leptopoda: :(

packages/neon_framework/packages/notifications_push_repository/lib/src/models/push_notification.dart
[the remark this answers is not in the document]
provokateurin: I don' think I get what you mean :thinking:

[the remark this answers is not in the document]
Leptopoda: I think we should consolidate the serializer from `push_notifications.dart` and `push_subscription.dart` into a single dart file similar to the account repository: https://github.com/nextcloud/neon/blob/86de9d777753667775b305b61ed07ab20047ffd7/packages/neon_framework/packages/account_repository/lib/src/models/serializers.dart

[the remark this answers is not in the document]
provokateurin: I don't see how this is beneficial

Leptopoda (approved):

PR #2484, opened 2024-09-12
chore(deps): remove unused dev dependencies

Leptopoda:
Web tests are still a bit flaky when run together, and I didn't set up the test yet (my bash skills are terrible).
Some initial feedback is appreciated, as I don't expect much to change.

To test it yourself:
- build the server `dart run dart_frog_cli:dart_frog build` (in the nc_test_api package)
- run the server `dart run packages/nextcloud_test_api/build/bin/server.dart` (in the nc package)
- run the tests `dart test`

As mentioned above, vm tests are fine, and all tests pass fine when run individually.


Leptopoda:
I also have a branch using a hybrid test, but that requires spawning a separate docker container for every test case.

provokateurin:
Doesn't need to be on any milestone since it only affects tests

PR #2488, opened 2024-09-12
refactor(neon_framework): rename credentials password to appPassowrd

Leptopoda:
As discussed in the initial account_repository PR (https://github.com/nextcloud/neon/pull/2403#discussion_r1720686033).

### DIFF
--- a/packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
+++ b/packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
@@ -219,7 +219,7 @@ class AccountRepository {
         b
           ..serverURL = Uri.parse(response.server)
           ..username = response.loginName
-          ..password = response.appPassword;
+          ..appPassword = response.appPassword;
       });
     } on http.ClientException catch (error, stackTrace) {
       if (error case DynamiteStatusCodeException(statusCode: 404)) {
--- a/packages/neon_framework/lib/src/blocs/push_notifications.dart
+++ b/packages/neon_framework/lib/src/blocs/push_notifications.dart
@@ -136,7 +136,7 @@ class _PushNotificationsBloc extends Bloc implements PushNotificationsBloc {
 
   Future<void> registerUnifiedPushInstances(({Account? active, BuiltList<Account> accounts}) event) async {
     // Notifications will only work on accounts with app password
-    for (final account in event.accounts.where((a) => a.password != null)) {
+    for (final account in event.accounts.where((a) => a.credentials.appPassword != null)) {
       await UnifiedPush.registerApp(account.id);
     }
   }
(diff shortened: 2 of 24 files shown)


packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -247,7 +247,7 @@ class AccountRepository {
         b
           ..client = client
           ..credentials.replace(credentials)
-          ..credentials.username = response.body.ocs.data.id;
+          ..credentials.loginName = response.body.ocs.data.id;
```
provokateurin, line 250: This is not correct as it is actually the username.
We should have loginName in the Credentials and username in the Account.
The app password we receive is bound to the loginName, while we need the username in different places (e.g. to generate the user ID).

packages/neon_framework/packages/talk_app/lib/src/blocs/room.dart
```diff
@@ -481,7 +481,7 @@ class _TalkRoomBloc extends InteractiveBloc implements TalkRoomBloc {
           b.clear();
 
           for (final entry in reactions.entries) {
-            if (entry.value.where((r) => r.actorId == account.username).isNotEmpty) {
+            if (entry.value.where((r) => r.actorId == account.credentials.loginName).isNotEmpty) {
```
provokateurin, line 484: This is not going to work as the actorId is always a username and not a loginName (same for all the other changes in Talk).

provokateurin (changes requested):
loginName and username can not be mixed and we need both.

packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
Leptopoda: Huh, so this means our current logic is already wrong?

provokateurin: I think so. I always login via username, but if you'd try to login via E-Mail it should not work as the loginName is different from the username.

provokateurin: I think it is actually fine for OCS APIs because they only use the Bearer auth where the token is not connected to the loginName.
For non-OCS APIs this is not the case because we have to use Basic auth which connects loginName and token.

I am pretty sure I can actually test a fix for this because I believe I have a case where this matters and it is not working at all at the moment.

Leptopoda: That's good to hear.
I have a local branch that fixes this behavior but it requires #1538 to be clean.

provokateurin: Huh why does it depend on that?

Leptopoda: I already refactored the entire username handling into a UserDetails bloc that makes use of the new storage lib.

I can change it to use the current storage although I don't see the point of it if we decide to go with the storage anyways.

Leptopoda:
I just dropped the login name change and will fix that in the UserDetails bloc refactor.

provokateurin (approved):

PR #2490, opened 2024-09-13
fix(neon_framework): account switching

Leptopoda:
I forgot that go_router will cache the route you are already on.
This fixes the 401 issues I had when switching accounts.


provokateurin (approved):
Makes sense :+1:

PR #2501, opened 2024-09-15
fix(nextcloud)!: Move RichObjectParameter from spreed to core and fix rich parameter typing for broken apps

provokateurin:
Split off from https://github.com/nextcloud/neon/pull/2456

I squashed the previously separate commits as they need each other to make sense.


packages/neon_framework/packages/talk_app/lib/src/widgets/rich_object/file.dart
```diff
@@ -35,8 +32,8 @@ class TalkRichObjectFile extends StatelessWidget {
 
         if (parameter.width != null && parameter.height != null) {
           deviceSize = Size(
-            _parseDimension(parameter.width!).toDouble(),
-            _parseDimension(parameter.height!).toDouble(),
+            int.parse(parameter.width!).toDouble(),
+            int.parse(parameter.height!).toDouble(),
```
Leptopoda, line 36:
```suggestion
            double.parse(parameter.width!),
            double.parse(parameter.height!),
```

Leptopoda (changes requested):
Not ideal, that we lose the typing on the spreed and notifications :(
LGTM overall, please just address the small parsing issue.

packages/neon_framework/packages/talk_app/lib/src/widgets/rich_object/file.dart
provokateurin: I did this to make sure the values are valid integers.
But yeah we shouldn't do this kind of enforcing that only tests the API and not our code.

Leptopoda (approved):

PR #2502, opened 2024-09-15
refactor(neon_http_client): use nextcloud csrf request

Leptopoda:
some housekeeping to use and test the now documented csrf endpoint


provokateurin:
Why are you splitting the logic for vm and web and basically making vm noop? IMO we should use the same logic regardless of the target. In this case it only increases code complexity while having not much benefits.
The other two commits LGTM.

Leptopoda:
We already had this split logic (it was just in a single class).
I found it less complex to split it into two separate classes, as there is only one condition being made.

I also find the unit tests more readable now that they are split.
The new solution will also avoid creating the nextcloud client on io as it isn't even needed (no real impact but cleaner nevertheless).

I can drop the commit if you prefer to keep it unified.

provokateurin:
Instead of having two implementations I would rather just not add it on vm.
Everything else can stay the same then.

Leptopoda:
I unified the csrf logic as previously discussed in our call.
I also changed the fixture validation so we also get one for the csrf test.

provokateurin (approved):

Leptopoda:
I'll change the webdav csrf interceptor (in the nc package) as a followup

Issue #2508, opened 2024-09-18
Invalid tables spec causes deserialization errors in the tables client.


Leptopoda:
The tables specs state that the result of a `api1.indexTableRowsSimple` request returns a `List<Strings>`
https://github.com/nextcloud/neon/blob/1e1d4035622b23b011516428653cd37ef9fd7699/packages/nextcloud/lib/src/api/tables/tables.openapi.json#L3794-L3908

But it actually returns a `List<List<String>>`:
```json
[
  [
    "What",
    "How to do",
    "Ease of use",
    "Done"
  ],
  [
    "Open the tables app",
    "Click on tables icon in the menu bar.",
    5,
    "true"
  ],
  [
    "Add your first row",
    "Just click on \"new row\" and enter some data inside of the form. At the end click on the bottom \"save\".",
    5,
    "false"
  ],
  [
    "Edit a row",
    "Hover the mouse over a row you want to edit. Click on the pen on the right side. Maybe you want to add a \"done\" status to this row.",
    5,
    "false"
  ],
  [
    "Add a new column",
    "You can add, remove and adjust columns as you need it. Click on the three-dot-menu on the upper right of this table and choose \"create column\". Fill in the data you want, at least a title and column type.",
    4,
    "false"
  ],
  [
    "Read the docs",
    "If you want to go through the documentation, this can be found here: https:\\/\\/github.com\\/nextcloud\\/tables\\/wiki",
    3,
    "false"
  ]
]
```

I don't think this is an issue in the openapi-extractor, as the typing of the relevant php code is also refers to a `string[]`.
Opening an issue upstream to hopefully get it fixed soon.
https://github.com/nextcloud/tables/blob/5cc780ce4260c262e3b733f5cb9d7959a9262938/lib/Controller/Api1Controller.php#L1032-L1060

We might be able to patch the spec for now, but I'm not sure how many endpoints are affected by this issue.

PR #2509, opened 2024-09-18
fix(nextcloud)!: tables row endpoints

Leptopoda:
fixes: #2508

CAUTION: I just took a look at the returned data and changed the spec to match that. It might still be possible that the api returns something different that doesn't match the patched spec.
Consulting the tables source code wasn't an option, as the wrong typing likely originates from it (no information to gain there).

I'm not sure how to move forward from here.
I don't have any confidence that the specs for that api are of any use. When something as fundamental as the typing of the row data is wrong, there are likely many more issues to be found. I dislike having to make a breaking release every time we try to fix it.
We might need to remove the tables client from the Nextcloud package again.

### DIFF
--- a/packages/nextcloud/lib/src/api/tables/patches/1-broken-spec.json
+++ b/packages/nextcloud/lib/src/api/tables/patches/1-broken-spec.json
@@ -1,0 +1,52 @@
+[
+    {
+        "op": "replace",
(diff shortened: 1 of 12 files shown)


Leptopoda:
Marking as a draft because the specs are just fundamentally broken and will require upstream changes

Leptopoda:
Closing in favor of an upstream fix like: https://github.com/nextcloud/tables/pull/1386

PR #2526, opened 2024-09-28
feat(neon_lints): update lints

Leptopoda:
I disabled `document_ignores` for now as it would be too much effort to document all of them now.

### DIFF
--- a/packages/neon_lints/lint_maker.yaml
+++ b/packages/neon_lints/lint_maker.yaml
@@ -16,11 +16,12 @@ base:
         avoid_as: false
         avoid_catches_without_on_clauses: false
         diagnostic_describe_all_properties: false
+        document_ignores: false
         lines_longer_than_80_chars: false
         no_default_cases: false
         one_member_abstracts: false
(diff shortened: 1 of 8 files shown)


packages/neon_lints/lint_maker.yaml
```diff
@@ -16,11 +16,12 @@ base:
         avoid_as: false
         avoid_catches_without_on_clauses: false
         diagnostic_describe_all_properties: false
+        document_ignores: false
```
provokateurin, line 19: I feel like we should turn this on eventually.
Almost every time where the docs are missing it is a mistake and not done for a good reason.

provokateurin (commented):
Given https://github.com/dart-lang/linter/issues/5065, should the type references really be converted or just ignored for now?

Leptopoda:
> Given dart-lang/linter#5065, should the type references really be converted or just ignored for now?

I don't see this moving forward in the near future, so we'd need to disable the lint. but I'm not so sure

provokateurin (approved):

PR #2528, opened 2024-09-29
fix(neon_framework): Migrate to notifications_push_repository

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/577

Tests still to be written


Leptopoda:
So far LGTM
Can't wait for the tests :)

provokateurin:
@Leptopoda ready with tests :)
There are some more things I want to fix regarding push notifications, but I would like to do those as follow-ups to at least have working push notifications again.

packages/neon_framework/pubspec.yaml
```diff
@@ -33,7 +32,12 @@ dependencies:
     sdk: flutter
   flutter_bloc: ^8.1.6
   flutter_blurhash: ^0.8.2
-  flutter_local_notifications: ^17.0.0
+  # https://github.com/MaikuB/flutter_local_notifications/pull/2418
+  flutter_local_notifications:
+    git:
+      url: https://github.com/MaikuB/flutter_local_notifications
+      ref: d9789d8496545ca2c5a42b78b9f49dc8efe2929b
+      path: flutter_local_notifications
```
provokateurin, line 40: I had to fix this myself to allow mocking the plugin in tests. From the repository git activity it looks like we might get a new release soon so we could just wait for that if we don't want to depend on this single git commit.

provokateurin: No activity since, so we just have to stay on this commit for now.

Leptopoda (approved):

PR #2537, opened 2024-10-01
feat(nextcloud_test_presets): Add and verify app release checksums

provokateurin:
Ensures we always download a valid file and that the release can not be replaced without noticing it.

### DIFF
--- a/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
+++ b/packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 import 'dart:io' show Directory, File, Link;
 
+import 'package:crypto/crypto.dart';
 import 'package:http/http.dart' as http;
 import 'package:nextcloud/core.dart' as core;
 import 'package:nextcloud_test_presets/src/models/models.dart';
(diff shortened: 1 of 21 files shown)


Leptopoda (commented):
Do we really need to store the file?
I'm not a fan of many files cluttering my tmp folder if not needed.

1. The sha checksum should directly convert the byte stream
2. If we need the caching to avoid multiple downloads just store the hash in a map.

provokateurin:
> If we need the caching to avoid multiple downloads just store the hash in a map.

Good idea, somehow I didn't think about that solution :woman_facepalming:

packages/nextcloud/packages/nextcloud_test_presets/lib/src/generate_presets.dart
```diff
@@ -39,17 +42,31 @@ Future<void> generatePresets() async {
...
+
+        if (urlChecksums[appRelease.url] == null) {
+          final request = http.Request('GET', Uri.parse(appRelease.url));
+
+          final streamedResponse = await httpClient.send(request);
+          if (streamedResponse.statusCode != 200) {
+            throw Exception('Unable to get app, status code: ${streamedResponse.statusCode}');
+          }
+
+          final response = await http.Response.fromStream(streamedResponse);
```
Leptopoda, line 62: I think we should use the converter directly on the stream. I think this avoids having the entire content in memory.

Leptopoda (approved):

PR #2540, opened 2024-10-02
ci: Run all Dart steps with both upgraded and downgraded dependencies

provokateurin:
Running the tests also ensures the code actually works and it's not just throwing no linting errors.

I only added the overrides to packages with tests as there is no way to automatically verify the overrides fix the build issues otherwise.
There seem to be some runtime issues with flutter_svg and vector_graphics, but I didn't look further into those.
This is only a step in the right direction and to verify that everything is fine with downgraded dependencies we need to have tests for every package.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -8,10 +8,16 @@ concurrency:
 
 jobs:
   dart-ci:
-    name: Dart CI
+    name: Dart CI (${{ matrix.dependencies }})
     runs-on: ubuntu-24.04
     permissions:
       packages: write
+    strategy:
+      fail-fast: false
+      matrix:
+        dependencies:
+          - upgrade
+          - downgrade
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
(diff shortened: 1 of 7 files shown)


Leptopoda:
What is the state of this PR?
Is it ready for review? The last commit suggests that there is still work to do.
I'm not a fan of the dev overrides and am hesitant on approving them. 

Other than that, great idea, and thanks for tackling this issue.

provokateurin:
> What is the state of this PR?
Is it ready for review? The last commit suggests that there is still work to do.

It is basically ready, I just left the commit there to trigger the CI to show that it works.

>  I'm not a fan of the dev overrides and am hesitant on approving them.

Me neither, but it's the only way we can check that our own dependency constraints are working correctly.

provokateurin:
I would at least be more comfortable with the overrides if we'd use pub workspaces so we only have to define them once.

provokateurin:
The overrides are now applied from a patch when running with downgraded dependencies.
They can still get outdated without us noticing it, but it will only affect CI and not development or production.

provokateurin:
Works:
https://github.com/nextcloud/neon/actions/runs/11633891510/job/32399971796?pr=2540
https://github.com/nextcloud/neon/actions/runs/11633891510/job/32399972235?pr=2540

Leptopoda (approved):
Thanks a lot. I'm way more comfortable that way.

provokateurin:
Eh I just realized having the pubspec.lock changes in the patch as well is probably not a good idea as it is going to need manual fixing every week when renovate tries to update it.
I had it in the patch to ensure we know if some of the lower constraints change (as then it would fail in CI), but it probably has to be removed.
I will wait until the next renovate run to see what happens in reality.

provokateurin:
Just as I expected it breaks right away: https://github.com/nextcloud/neon/actions/runs/11656047519/job/32451613800?pr=2629

PR #2557, opened 2024-10-09
refactor(nextcloud)!: Migrate to DAV v2 endpoint for WebDAV

provokateurin:
Replaces https://github.com/nextcloud/neon/issues/2306
Should help with https://github.com/nextcloud/neon/pull/2308 and https://github.com/nextcloud/neon/issues/718 as those require using the v2 endpoints.


packages/nextcloud/lib/src/api/webdav/webdav_client.dart
```diff
@@ -51,15 +52,26 @@ class WebDavClient extends DynamiteClient {
...
 
-  Uri _constructUri([PathUri? path]) => constructUri(baseURL, path);
+  String? _username;
+
+  Future<Uri> _constructUri([PathUri? path]) async {
+    if (_username == null) {
+      final client = provisioning_api.$Client.fromClient(this);
+      // Consumers already need to catch http.ClientException so we don't need to catch and rethrow here.
+      final response = await client.users.getCurrentUser();
+      _username = response.body.ocs.data.id;
```
Leptopoda, line 62: So we could take the `loginName`  used by the root `NextcloudClient` as long as it was instantiated using the username and app password?
But because we also support instantiating it with a different loginName like email, we have to do this hack?

provokateurin: No, here it has to be the username.
The problem is even with basic auth it is not guaranteed that it is the username, as you can also use loginName:password for that.
So sadly no optimization potential here.

provokateurin:
What we could do alternatively is to require the username to be passed when creating the WebDavClient. That would be a bit cleaner as then all the http.Request getters can be sync and we avoid doing a network request (since the consumer of the package might already know the username anyway like we do in Neon).

Leptopoda:
This is also exactly what I thought.
Let's require the username at creation.

provokateurin:
We could add an extension in the framework that would allow us to still call `account.client.webdav` without the username by automatically inserting the account username and internally using the underlying extension with the username parameter.
I don't think we really need it, but if you like it I can add it.

PR #2576, opened 2024-10-19
fix(neon_framework): Fix Android push notification icon

provokateurin:
Before/after:
image

### DIFF
--- a/packages/neon_framework/test/push_utils_test.dart
+++ b/packages/neon_framework/test/push_utils_test.dart
@@ -119,7 +119,11 @@ void main() {
       FlutterLocalNotificationsPlatform.instance = localNotificationsPlatform;
       when(
         () => localNotificationsPlatform.initialize(
-          any(),
+          any(
+            that: predicate(
+              (s) => s is AndroidInitializationSettings && s.defaultIcon == '@drawable/ic_launcher_outline',
+            ),
+          ),
           onDidReceiveNotificationResponse: any(named: 'onDidReceiveNotificationResponse'),
         ),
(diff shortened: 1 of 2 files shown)


packages/neon_framework/test/push_utils_test.dart
```diff
@@ -119,7 +119,11 @@ void main() {
       FlutterLocalNotificationsPlatform.instance = localNotificationsPlatform;
       when(
         () => localNotificationsPlatform.initialize(
-          any(),
+          any(
+            that: predicate(
```
Leptopoda, line 123: Please use a type matcher instead of a predicate

provokateurin: That's not possible: https://github.com/MaikuB/flutter_local_notifications/blob/master/flutter_local_notifications/lib/src/platform_specifics/android/initialization_settings.dart

Leptopoda: I don't get why.
```dart
any(
isA<Type>().having((e) => e.attribute, "attribute", "expected")
)
```

Should work the same and produce better failure messages.
I just noticed that I missed this entirely in my last review and all other tests also make use of `predicate` matchers so let's just continue with it here.

Leptopoda (approved):

packages/neon_framework/test/push_utils_test.dart
provokateurin: Ah I thought you wanted to have it compared to another instance of AndroidInitializationSettings directly.

PR #2582, opened 2024-10-23
fix(neon_http_client): Make authorization throttling less agressive

provokateurin:
The detection flagged some responses that were fine (even though I believe nothing should really use the 401 status code except the authentication middleware), which resulted in the app becoming unusable due to the blocking.

I limited it to OCS responses for now as most requests use that format and we can correctly detect invalid credentials 100% of the time through the custom status code 997.
For non-OCS APIs and WebDAV it might be possible as well, but I think they are not so important and we can add them if ever needed.

I know there are those really ugly nested if blocks, but the alternative would be to chain null-aware casts and I don't think that is much better than this.

### DIFF
--- a/packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
+++ b/packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
@@ -1,4 +1,7 @@
+import 'dart:convert';
+
 import 'package:http/http.dart';
+import 'package:http_parser/http_parser.dart';
 import 'package:interceptor_http_client/interceptor_http_client.dart';
 import 'package:meta/meta.dart';
 import 'package:nextcloud/nextcloud.dart';
@@ -49,7 +52,7 @@ final class AuthorizationThrottlingInterceptor implements HttpInterceptor {
   }
 
   @override
-  StreamedResponse interceptResponse({required StreamedResponse response, required Uri url}) {
+  Future<StreamedResponse> interceptResponse({required StreamedResponse response, required Uri url}) async {
     assert(
       shouldInterceptResponse(response),
       'Response should not be intercepted.',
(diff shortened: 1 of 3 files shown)


packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
```diff
@@ -49,15 +52,52 @@ final class AuthorizationThrottlingInterceptor implements HttpInterceptor {
...
+          if (ocs is Map<String, dynamic>) {
+            final ocsMeta = ocs['meta'];
+            if (ocsMeta is Map<String, dynamic>) {
+              final ocsMetaStatusCode = ocsMeta['statuscode'];
+              if (ocsMetaStatusCode == 997) {
+                _blocked = true;
+              }
+            }
+          }
+        }
```
Leptopoda, line 87: Pattern matching should help (https://dart.dev/language/pattern-types#map):
```suggestion
        if (jsonBody case {'ocs': {'meta': {'statuscode': 997}}}) {
          _blocked = true;
        }
```

I hope there isn't any typo, but something like the above might do the trick.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
```diff
@@ -49,15 +52,52 @@ final class AuthorizationThrottlingInterceptor implements HttpInterceptor {
...
 
     final authorization = response.request!.headers['authorization'];
     if (authorization != null && authorization.isNotEmpty && response.statusCode == 401) {
-      _blocked = true;
+      final contentType = response.headers['content-type'];
+      if (contentType == null) {
+        return response;
+      }
+
+      final mediaType = MediaType.parse(contentType);
```
Leptopoda, line 68: IIrc this might throw a FormatException.
Please check and catch the appropriate one.

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
```diff
@@ -49,15 +52,52 @@ final class AuthorizationThrottlingInterceptor implements HttpInterceptor {
...
+            final ocsMeta = ocs['meta'];
+            if (ocsMeta is Map<String, dynamic>) {
+              final ocsMetaStatusCode = ocsMeta['statuscode'];
+              if (ocsMetaStatusCode == 997) {
+                _blocked = true;
+              }
+            }
+          }
+        }
+      } catch (_) {}
```
Leptopoda, line 88: This should only catch `FormatException`s

packages/neon_framework/packages/neon_http_client/test/interceptors/authorization_throttling_interceptor_test.dart
```diff
@@ -13,8 +15,23 @@ void main() {
       final authorizedRequest = Request('GET', baseURL);
       authorizedRequest.headers['authorization'] = 'test';
       final authorizedResponse = StreamedResponse(
-        const Stream.empty(),
+        Stream.value(
+          utf8.encode(
+            json.encode(
```
Leptopoda, line 20: No real difference as this is a test but there is `JsonUtf8Encoder`

packages/neon_framework/packages/neon_http_client/test/interceptors/authorization_throttling_interceptor_test.dart
```diff
@@ -48,8 +65,23 @@ void main() {
...
+                  },
+                },
+              },
+            ),
+          ),
+        ),
         401,
+        headers: {
+          'content-type': 'application/json; charset=utf-8',
+        },
```
Leptopoda, line 84: The stram and header generation can be extended into a helper method `buildStream` and `buildHeaders`.

Leptopoda (changes requested):
Now we also need tests validating that:
- non json responses are not intercepted
- json responses not matching the pattern are not intercepted

packages/neon_framework/packages/neon_http_client/lib/src/interceptors/authorization_throttling_interceptor.dart
provokateurin: Works :heart_eyes:

Leptopoda (approved):

PR #2585, opened 2024-10-24
fix(neon_framework): Group push notifications on Android

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2581

image

I know it's not perfect yet, but it avoids getting spammed by push notifications if there are multiple.

### DIFF
--- a/packages/neon_framework/lib/src/utils/push_utils.dart
+++ b/packages/neon_framework/lib/src/utils/push_utils.dart
@@ -1,11 +1,11 @@
 import 'dart:async';
 import 'dart:convert';
-import 'dart:typed_data';
 import 'dart:ui';
 
 import 'package:account_repository/account_repository.dart';
 import 'package:built_collection/built_collection.dart';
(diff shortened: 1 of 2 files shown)


packages/neon_framework/lib/src/utils/push_utils.dart
```diff
@@ -185,6 +185,29 @@ class PushUtils {
             ),
             payload: json.encode(pushNotification.toJson()),
           );
+
+          // Other platforms don't support grouping, so don't send another notification there.
+          if (defaultTargetPlatform == TargetPlatform.android) {
+            // Sine we only support Android SDK 24+, we can just generate an empty summary notification as it will not be shown anyway.
```
Leptopoda, line 191:
```suggestion
            // Since we only support Android SDK 24+, we can just generate an empty summary notification as it will not be shown anyway.
```

Leptopoda (approved):

PR #2603, opened 2024-10-28
fix(neon_framework): Handle delete-multiple background push notifications

provokateurin:
I discovered this while investigating https://github.com/nextcloud/neon/issues/2591 (I even first thought this was the initial problem).
Added documentation upstream (https://github.com/nextcloud/notifications/pull/2089) because it was not documented back when it was added (https://github.com/nextcloud/notifications/pull/1429).


Leptopoda (approved):
LGTM

PR #2607, opened 2024-10-29
fix(tool): Group custom_lint packages in renovate config

Leptopoda:
should unblock #2596 and #2559


provokateurin (approved):
Thanks, I already noticed it as well but didn't have the time to fix it.

PR #2623, opened 2024-11-01
refactor(neon_framework): Implement a custom emoji picker

provokateurin:
Fixes https://github.com/nextcloud/neon/issues/2499

I thought implementing a custom emoji picker would be hard because all the existing packages suck and are unmaintained, but it turned out it was quite simple in the end.

Because the available emojis are generated from the metadata they are also more complete than what we had before, because the package was manually adding them to a list.

This picker does not support alternates (e.g. skin tones) yet, but it can be added easily.
It also does not save any recently used emojis, but we can add that as well.

Due to having all the metadata available we can now also add shortcode (e.g. `:smile:`) completion for text inputs (e.g. Talk message input) which is pretty neat.

For some reason the Tabbar renders a bit weird on Android and I believe it might be connected to https://github.com/flutter/flutter/issues/119623 as I was seeing that issue as well. On Linux it renders perfectly and the emojis are centered as well :woman_shrugging:

### DIFF
--- a/packages/neon_framework/generate_emojis.dart
+++ b/packages/neon_framework/generate_emojis.dart
@@ -1,0 +1,225 @@
+import 'dart:convert';
+
+import 'package:code_builder/code_builder.dart';
+import 'package:collection/collection.dart';
+import 'package:dart_style/dart_style.dart';
+import 'package:string_normalizer/string_normalizer.dart';
+import 'package:universal_io/io.dart';
+
+/// Android 14 only supports Unicode 15.0, so we use that for now, even though newer versions are available:
+/// https://developer.android.com/guide/topics/resources/internationalization#versioning-nougat
+const unicodeVersion = '15_0';
+
+void main() {
+  final formatter = DartFormatter(pageWidth: 120);
+  final emitter = DartEmitter(
+    orderDirectives: true,
+    useNullSafetySyntax: true,
+  );
+
+  final inputFile = File('../../external/emoji-metadata/emoji_${unicodeVersion}_ordering.json');
+  final input = json.decode(inputFile.readAsStringSync());
+
+  final library = Library((b) {
+    b.docs.add('// GENERATED CODE - DO NOT MODIFY BY HAND');
+
+    b.body.add(
+      Class(
+        (b) => b
+          ..name = 'Emoji'
+          ..docs.add('/// Holds the metadata of a Unicode emoji.')
+          ..constructors.add(
+            Constructor(
+              (b) => b
+                ..constant = true
+                ..docs.add('/// Creates a new [Emoji].')
+                ..optionalParameters.addAll([
+                  Parameter(
+                    (b) => b
+                      ..name = 'base'
+                      ..named = true
+                      ..toThis = true
+                      ..required = true,
+                  ),
+                  Parameter(
+                    (b) => b
+                      ..name = 'alternates'
+                      ..named = true
+                      ..toThis = true
+                      ..required = true,
+                  ),
+                  Parameter(
+                    (b) => b
+                      ..name = 'emoticons'
+                      ..named = true
+                      ..toThis = true
+                      ..required = true,
+                  ),
+                  Parameter(
+                    (b) => b
+                      ..name = 'shortcodes'
+                      ..named = true
+                      ..toThis = true
+                      ..required = true,
+                  ),
+                  Parameter(
+                    (b) => b
+                      ..name = 'animated'
+                      ..named = true
+                      ..toThis = true
+                      ..required = true,
+                  ),
+                ]),
+            ),
+          )
+          ..fields.addAll([
+            Field(
+              (b) => b
+                ..name = 'base'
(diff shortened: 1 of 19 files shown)


provokateurin:
> what about making this a separate package

I thought about this as well, but right now the code is still quite simple.
Would you just try to separate the code or also publish the package at some point?
There are already a lot of emoji picker packages out there, so naming would be a problem.

> Why not use the officil unicode cldr

For what? I don't understand where you would want to use it.

Leptopoda:
> For what? I don't understand where you would want to use it.

I assume that the emoji category names are currently hand translated.
The unicode cldr already contains official translations for all languages supported by unicode.
These could be used to also generate arb files for all languages making the emoji picker more complete.

>  Would you just try to separate the code or also publish the package at some point?

Once we have translations and search strings for all emojis/languages the app size might increase quite significantly.
Because the emoji picker will not be used by all apps I'd rather have separate arb files for them that are only inserted into the widget tree for the widgets in question.

Even if you we keep it in the neon_framework package, I'd like to have this separation.

provokateurin:
> I assume that the emoji category names are currently hand translated.

Yes

>  The unicode cldr already contains official translations for all languages supported by unicode.

It contains translations for all the symbols, but not for the category names as they are not unicode symbols themselves.

> Once we have translations and search strings for all emojis/languages the app size might increase quite significantly.

Don't forget it's all just string data. 
Also we don't need any translations for the emojis, as translation software and screenreaders do this already internally.
The only translations we need are the group names and there are not many so it's insignificant.

I just did a size analysis of the APK and emojis.dart is 0.1KB and the NeonEmojiPickerDialog is 39.5KB.
I think this is totally acceptable and we have other places that should be optimized first before doing this.
I also did a comparison of the current main branch and this branch and the overall size shrunk by 0.6KB (less than I expected, but it's not more so a win nevertheless).

provokateurin:
We might want to ship and use an Emoji font like https://github.com/googlefonts/noto-emoji directly, to ensure all platforms render them correctly (e.g. different Android versions support different Unicode standards) and the same way.
I think only on iOS and macOS it could go "wrong", because they normally use a different font and not Noto Emoji.

Leptopoda:
I don't know if it is worth the apk size bump.
Maybe we can use the appcompat lib on android:
https://developer.android.com/develop/ui/views/text-and-emoji/emoji2

It's supported since android 11 (~80% of active devices).

provokateurin:
Ah nice I didn't know about that. Unfortunately it says this:
> Your app automatically displays backward-compatible emoji on all devices that provide an emoji2-compatible downloadable fonts provider, such as devices powered by [Google Play services](https://developers.google.com/android).

And my guess would be that it won't work on devices without Google Play services (I would have to test it though).

provokateurin:
> I don't know if it is worth the apk size bump.

Hm yeah it's 10.2 MB, that is 1/3 of our current APK size and it doesn't give us that much benefits except for less broken emojis.

Maybe the better way to deal with this is to dynamically decide at runtime which emojis to keep, so checking the Android API level and then deciding which Unicode version we can assume to be available.

provokateurin:
We could also use https://pub.dev/packages/google_fonts and have the emoji font downloaded at runtime, but there we run into other issues again as we have to explicitly use the font everywhere (see https://github.com/material-foundation/flutter-packages/issues/155)

provokateurin:
Found the CLDR data for the emoji groups: https://github.com/unicode-org/cldr-json/blob/main/cldr-json/cldr-misc-full/main/de/characterLabels.json.
Let's use them in the future and keep the manual translations for now.

packages/neon_framework/generate_emojis.dart
```diff
@@ -0,0 +1,218 @@
...
+  final emitter = DartEmitter(
+    orderDirectives: true,
+    useNullSafetySyntax: true,
+  );
+
+  final inputFile = File('../../external/emoji-metadata/emoji_${unicodeVersion}_ordering.json');
+  final input = json.decode(inputFile.readAsStringSync());
+
+  final library = Library((b) {
+    b.ignoreForFile.add('public_member_api_docs');
```
Leptopoda, line 24: I disagree. Please document the `Emoji` class

packages/neon_framework/generate_emojis.dart
```diff
@@ -0,0 +1,218 @@
...
+                  ..type = refer('List<String>')
+                  ..modifier = FieldModifier.final$,
+              ),
+              Field(
+                (b) => b
+                  ..name = 'animated'
+                  ..type = refer('bool')
+                  ..modifier = FieldModifier.final$,
+              ),
+            ],
```
Leptopoda, line 105:
```suggestion
            ]
```

Leptopoda (changes requested):

.cspell/misc.txt
Leptopoda, line 1: please just ignore the file

Leptopoda (approved):

PR #2634, opened 2024-11-06
fix(ci): speedup fdroid nightly by using move instatt of copy (and use less memory in runner)

wrenix:
hope schedule works on runner again (with nightly clone and so on) ... 

the default Linux runner has just 14 GB Disk: https://docs.github.com/en/actions/writing-workflows/choosing-where-your-workflow-runs/choosing-the-runner-for-a-job#standard-github-hosted-runners-for-public-repositories

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -59,9 +59,9 @@ jobs:
 
       - name: F-Droid nightly
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -1,6 +1,7 @@
 name: Publish
 
 on:
+  push:
```
provokateurin, line 4: Please remove this

provokateurin (changes requested):
Sure this is going to save some disk space, but I don't think disk space is the reason the workflow is failing?

PR #2647, opened 2024-11-11
feat(neon_rich_text): Render markdown

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1929
Just like the emoji picker I thought it would be much harder, but it was not too bad (but still quite a lot of effort, especially the final newline handling as you will see).

This is a fully custom rendering implementation for markdown including rich objects.
Using https://pub.dev/packages/flutter_markdown was not an option because it does not expose the necessary low level APIs to implement everything we need.
I mostly tested this manually with the example from https://markdown-it.github.io.
This likely implements some things that are actually not supported in Talk right now,

I did not write proper tests for the rendering itself (but there is on Talk test that ensures markdown is rendered when requested). There are multiple viable sources we can use to cover the tests:
1. Dart markdown package tests: https://github.com/dart-lang/markdown/tree/master/test
2. Github flavored markdown tests: https://github.com/github/cmark-gfm/tree/master/test
3. nextcloud-vue tests for NcRichText (basically the equivalent Vue component responsible for rendering markdown and rich objects): https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/cypress/component/richtext.cy.ts

IMO option 3 is the best right now, because it will strike a good balance between effort and coverage. Anything not covered by these tests should not ever appear in a Talk message so we should be good.

While writing this I realized I never tested check boxes and they are not implemented either. In Talk there is this cool feature where a click on a checkbox will toggle it and actually edit the message. Implementing that will probably not be easy, so I would like to skip it for now and open a new issue for it once this is merged.

This isn't finished completely, because I first wanted to discuss the general direction before putting even more effort into this.


Leptopoda:
So we don't forget it.
I think that this is at a point that it would benefit from being a separate package.
Do you agree? In that case, I'd prefer to have this done first.

provokateurin:
Yeah let's do that, this feature is getting too big by now.

provokateurin:
Moving to a separate package and the cleanups got to big, so I split that off into https://github.com/nextcloud/neon/pull/2682.

provokateurin:
I didn't add any tests yet and the code in this PR also has to be refactored before a proper review can take place.

provokateurin:
Just a rebase for neon_rich_text, nothing new or tests added.

provokateurin:
I implemented all tests from https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/cypress/component/richtext.cy.ts plus a few that are simply missing in those tests but are working in the web interface and are implemented by us.

I think this is the perfect use-case for using flutter goldens to ensure the rendering is exactly as expected. Proper fonts had to be used in order to make sure the results are correct. Let's hope the font rendering is consistent...

Leptopoda:
> I think this is the perfect use-case for using flutter goldens to ensure the rendering is exactly as expected.

I agree.

Can you please make a separate PR for the two fixes? This is already a quite big one.

provokateurin:
> Can you please make a separate PR for the two fixes?

Done, but the second fix is required for this PR so I based this PR on it.

provokateurin:
(Put it in draft to avoid accidentally merging it into the other branch)

provokateurin:
This likely needs a rebase for https://github.com/nextcloud/neon/pull/2703 as the rendering might have changed slightly.

provokateurin:
@Leptopoda I'm going to merge this now to drastically improve the usability of the Talk app. You can still leave feedback later and we can change things if needed.

PR #2659, opened 2024-11-17
feat(notifications_push_repository): Verify push notification signature

provokateurin:
Closes https://github.com/nextcloud/neon/issues/213

We don't need to catch the exception, as we can not do anything about it and it should never happen, unless the Nextcloud server or the UnifiedPush provider are compromised.


packages/neon_framework/packages/notifications_push_repository/lib/src/models/push_notification.dart
```diff
@@ -19,16 +21,28 @@ abstract class PushNotification implements Built<PushNotification, PushNotificat
...
-    final subject = notifications.DecryptedSubject.fromEncrypted(privateKey, json['subject'] as String);
+    final subject = json['subject'] as String;
+    final signature = json['signature'] as String;
+
+    final valid = userPublicKey.verifySHA512Signature(
+      base64.decode(subject),
+      base64.decode(signature),
+    );
+    if (!valid) {
+      throw Exception('Failed to verify push notification signature!');
```
Leptopoda, line 35: Maybe we should add a custom `PushNotificationException` instead of throwing a generic one

PR #2664, opened 2024-11-19
feat(neon_framework): Implement remote wipe

provokateurin:
Closes https://github.com/nextcloud/neon/issues/43.
Not all data is deleted yet, but that will be fixed in https://github.com/nextcloud/neon/issues/2114.

~Tests for the changes in AccountsBloc still have to be written, but there are no tests for it at all so far, so it is a slightly bigger task than I anticipated.~


provokateurin:
Note that this implementation is not per docs, as we only check if the account needs to be wiped once per app start and not for every 401/403 response we receive.
This should still be decently safe and definitely more safe than not implementing it at all.

PR #2667, opened 2024-11-20
fix(notifications_push_repository): Delete push subscription when account is deleted

provokateurin:
Part of https://github.com/nextcloud/neon/issues/2114

### DIFF
--- a/packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
+++ b/packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
@@ -14,6 +14,9 @@ import 'package:rxdart/rxdart.dart';
 
 part 'account_storage.dart';
 
+/// Signature of a callback that executed right before the [account] is logged out.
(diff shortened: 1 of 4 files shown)


packages/neon_framework/packages/account_repository/lib/src/account_repository.dart
```diff
@@ -299,8 +315,12 @@ class AccountRepository {
 
     _accounts.add((active: active, accounts: accounts));
 
+    for (final callback in _beforeLogOutCallbacks) {
+      await callback(account!);
+    }
```
Leptopoda, line 320: I think that running these callbacks in parallel should be ok.
The account_repo can't guarantee execution order anyway.

Please use a `Future.wait`

Leptopoda (changes requested):
I still don't like interlocking the repositories.
I think that the account_bloc should handle this. Once a `LogoutEvent` is emitted, the bloc should unregister the notifications and afterward remove the account from the repository.

PR #2682, opened 2024-11-27
feat(neon_rich_text): Init

provokateurin:
Prerequisite for https://github.com/nextcloud/neon/pull/2647.
Just moved everything to it's own package and refactored the code and tests a bit to be easier to work with.


packages/neon_framework/lib/widgets.dart
```diff
@@ -7,7 +7,7 @@ export 'package:neon_framework/src/widgets/image.dart' hide NeonImage;
 export 'package:neon_framework/src/widgets/linear_progress_indicator.dart';
 export 'package:neon_framework/src/widgets/list_view.dart';
 export 'package:neon_framework/src/widgets/relative_time.dart';
-export 'package:neon_framework/src/widgets/rich_text/rich_text.dart';
 export 'package:neon_framework/src/widgets/server_icon.dart';
 export 'package:neon_framework/src/widgets/user_avatar.dart' hide NeonUserStatusIndicator;
 export 'package:neon_framework/src/widgets/user_status_icon.dart';
+export 'package:neon_rich_text/neon_rich_text.dart';
```
Leptopoda, line 13: Should we really re-export this package?
What stands against just letting the consumers import it?

packages/neon_framework/packages/neon_rich_text/lib/neon_rich_text.dart
Leptopoda, line 1: I'm not a big fan of re-exporting the libs from the `rich_objects.dart`
This makes the export of `'src/rich_objects/*.dart'` implicit and thus less obvious.

Why not leave the exports as they are and let `rich_objects.dart` import `neon_rich_text.dart`?

Leptopoda (commented):
This was a needed cleanup. Thanks a lot.

I just have some structuring questions.

packages/neon_framework/lib/widgets.dart
provokateurin: I mainly did it because we also export parts of account_repository and I felt this makes it easier for consumers, but in the end it is better to not do this to avoid that we accidentally create a cyclic dependency at some point.

Leptopoda (approved):

PR #2684, opened 2024-11-28
ci: Check generated assets are up-to-date

provokateurin:
Closes https://github.com/nextcloud/neon/issues/1430


provokateurin:
I fear this idea is not possible right now, as different versions are used and even the precompiled SVGs are different (although I don't understand why).
With a nix flake it should be possible to achieve this though, as then everything should be the exact same.

PR #2704, opened 2024-12-12
build!: Migrate to pub workspace

provokateurin:
I'm not sure it will work right away, but let's see :rocket:


.github/workflows/conventional_commits.yaml
```diff
@@ -15,9 +15,9 @@ jobs:
       - name: Install dart
         uses: dart-lang/setup-dart@e630b99d28a3b71860378cafdc2a067c71107f94 # v1
         with:
-          sdk: 3.5
+          sdk: 3.6
       - name: Install commitlint_cli
-        run: dart pub get
+        run: dart pub global activate commitlint_cli
```
Leptopoda, line 20: I'm not a big fan of this.
I'd still like to pin the version we use in CI

packages/cookie_store/packages/cookie_store_conformance_tests/pubspec.yaml
```diff
@@ -3,19 +3,17 @@ description: >-
...
+    path: ../..
   meta: ^1.0.0
   test: ^1.21.2
 
 dev_dependencies:
   neon_lints:
-    git:
-      url: https://github.com/nextcloud/neon
-      path: packages/neon_lints
+    path: ../../../neon_lints
```
Leptopoda, line 19: We should have it as a dev_dependency of the workspace package and just use a global `analysis_options.yaml`.

Every package may still alter the rule set but this would make our lints the default (also for newly added packages).

tool/setup.sh
```diff
@@ -2,16 +2,16 @@
...
 cd "$(dirname "$0")/.."
 
-for package in $(yq -r ".dev_dependencies | keys | .[]" pubspec.yaml); do
-  version="$(yq -r ".dev_dependencies.$package" pubspec.yaml)"
-  dart pub global activate "$package" "$version"
-done
+dart pub global activate coverage
+dart pub global activate custom_lint
+dart pub global activate fvm
+dart pub global activate melos
```
Leptopoda, line 8: I'm not a big fan of this.
Maybe we should first migrate this repo to a nix flake before we merge this one

Leptopoda (commented):
some initial feedback

packages/cookie_store/packages/cookie_store_conformance_tests/pubspec.yaml
provokateurin: I didn't think about this, but yes it makes sense. How are we going to handle custom_lint though? We need to disable some of the rules for specific packages?
OTOH this brings up the point of removing custom_lint entirely, given that it mostly causes headaches and hasn't caught so many problems so far.

tool/setup.sh
provokateurin: I will pin them to exact versions for now, but yes the flake will be better than this.
Unfortunately it needs more work so I would like to have this PR merged without waiting for it.

provokateurin:
For the tests with downgraded dependencies we probably need `test: ^1.25.12`: https://github.com/dart-lang/test/blob/master/pkgs/test/CHANGELOG.md#12512
Unfortunately we are currently stuck: https://github.com/nextcloud/neon/pull/2680

packages/cookie_store/packages/cookie_store_conformance_tests/pubspec.yaml
provokateurin: I put this into a separate issue as it's not really a blocker and we really need this PR to make everything simpler: https://github.com/nextcloud/neon/issues/2925

PR #2705, opened 2024-12-12
refactor: Use spacing from Row/Column instead of intersperse

provokateurin:


Leptopoda (approved):

Leptopoda:
Are you sure the first commit should be a fix?

provokateurin:
I think so because https://github.com/nextcloud/neon/pull/2703/files#diff-8862ea315b55c6cc198dea41818bd6c431000b38225f5b290ce58f1d4a6dced4 should be broken on lower flutter versions.

PR #3162, opened 2026-02-08
feat: add Flatpak configuration

PureTryOut:
This uses https://github.com/TheAppgineer/flatpak-flutter to generate a Flatpak manifest (de.provokateurin.neon.yaml) that can build from source without network access.
The input for that is the flatpak-flutter.yaml. When submitting the application to Flathub de.provokateurin.neon.yaml with the "generated" folder is what would be submitted to them.

~~Note that I have trouble testing this as during the building of `de.provokateurin.neon.yaml` Flatpak executes some commands on the host which then fails as I'm running postmarketOS (and bare Alpine Linux on some other systems) which uses a different libc than the Flatpak sandbox (musl vs glibc). Shouldn't be a problem on distributions using glibc though (most of them).~~ My system was just missing a binary on the host, I've built it successfully now.

You can build the application yourself as follows (instructions taken from https://docs.flatpak.org/en/latest/first-build.html):
```sh
$ flatpak remote-add --if-not-exists --user flathub https://dl.flathub.org/repo/flathub.flatpakrepo
$ flatpak-builder --force-clean --user --install-deps-from=flathub --repo=repo --install builddir de.provokateurin.neon.yaml
```

Please test it before accepting :smile:


provokateurin:
This is basically why I started the project. I wanted to have a nice Nextcloud client for Linux on phones. One main problem is that this is a lot of work, even though this project introduced techniques like OpenAPI that reduced a lot of code that would have to be written manually. This part was also adopted by the whole Nextcloud project, but implementing so much functionality is simply not possible for just 2 people. Like you said using one codebase also has a major advantage over the existing clients that duplicate features and bugs across codebases.

You can basically think of two people, competing with ~20 people from Nextcloud GmbH who have been doing this for years and full time, competing with probably thousands of people from Google and Microsoft who have been doing this even longer also on full time.

I'd be happy to help revive this project, if there is real interest and people who want to continue it.

provokateurin:
> has to build completely from source

I see, I didn't know about that.

> I'm interested, and I'd love to contribute. But I doubt me being interested is enough for you 😅

I'd be up for reviewing the occasional pull request, but at the moment I don't feel like starting to pick up work myself again. However I might have some more time again in a few months, when I have cleared up some personal stuff.

> Maybe you're biting off to much to chew at once

I know, we were always aware of that. Still it felt great to be able to achieve so much with so "little" work.

> Instead you have to start small and get basic functionality in there, for example perhaps just file syncing (since that's the core of Nextcloud after all) and get that working well and tested

There is a prototype for file sync: https://github.com/nextcloud/neon/pull/600

> So why go for such a feature in this one while it doesn't fully compete with the functionality of the main app yet? You're trying to catch up to a rocket going to the moon by shooting yourself to Mars.

Originally I really liked the idea of having multiple Nextcloud apps in a single app, but over time we also realized that it's better to have them split again at some point, such that they only use common libraries, but don't interact directly. OTOH IMO one of the great benefits of Nextcloud is that these apps integrate with each other and are not completely isolated, so...

> I think the basic goals of this framework and app are great, but you have to keep it manageable for the amount of people working on it

This was always a problem, because there were so many ideas, but we were lacking the amount of people that is required to be able to actually implement a fair share of them.

> This will get you users and through that probably also (some) developers, and at the very least it'll get you bug reports which are very useful on itself of course.

I always wished that would happen, but I was probably to afraid of making a first public release for people to try. Also I'm really not good at marketing my own work. This is clearly something this project needs help with, if it is revived.

> Once that's there, slowly expand the scope

I totally agree. Once the basic underlying structure is there, most things become super easy. Like some stuff we implemented for new versions took us only a few hours, while the developers of the official clients needed weeks...

> I don't know, perhaps I'm dreaming to much, but I think the goals are worthwhile going for

Absolutely not! I still admire the goals I once had with this, when I started it.

---

Thank you for your kind words! :heart: Maybe you'd be interested in chatting a little more? We should switch to a different channel though, this PR is not the best place :sweat_smile:

If there is any way to bring in new people and revive this project, I'm absolutely here for it and I'd appreciate any help you or anyone else could give.

provokateurin (commented):
Builds fine for me. It seemed to have no internet access, so I wasn't able to log in, but that could also be my fault. Did it work for you?

I think it's great to have such a setup to be able to build the flatpak from source, it should live somewhere else. Adding the ability to build from path instead of a git commit is not enough, because the generated files need be kept up-to-date and I don't want to add a CI that ensures this is the case.

I'm not sure if we really need this. You have successfully shown that this app can be built as a flatpak from source, so I'd suggest we just keep this PR in draft and once we want to publish to flathub we copy the files there and update them in the dedicated repo.

PureTryOut:
> It seemed to have no internet access

Fixed that. I only ran it to the setup screen, I didn't really test any further so missed this. Sorry for that!

> I think it's great to have such a setup to be able to build the flatpak from source, it should live somewhere else. Adding the ability to build from path instead of a git commit is not enough, because the generated files need be kept up-to-date and I don't want to add a CI that ensures this is the case.

I disagree, I think it's very useful to have automatic Flatpak builds for every commit in git as a "nightly" version that people can easily test out so we get bug reports quickly. There is not much to keep up-to-date either, the file as it is now will basically be it for a long long time and if that changes it's easy to just update it.
I suppose we can keep the current manifest already in the repo in `packages/neon_framework/example` to do this but that then has to be modified too to make it "nightly", and since it's different from what will be on Flathub will not be representative for what people actually get once there is a new release.

PR #3184, opened 2026-03-14
feat(neon_framework): queue for request manager to limit parallel req…

### LINKED ISSUE #1124: Use a queue for all requests in the request manager

vauvenal5:
Limits the amount of parallel requests to the server to not overload it with preview requests when opening a big gallery and scrolling through it.

As agreed in https://github.com/nextcloud/neon/issues/1124 this is a first step.

### DIFF
--- a/packages/neon_framework/lib/src/utils/request_manager.dart
+++ b/packages/neon_framework/lib/src/utils/request_manager.dart
@@ -11,6 +11,7 @@ import 'package:neon_framework/src/bloc/result.dart';
 import 'package:neon_framework/storage.dart';
 import 'package:neon_http_client/neon_http_client.dart';
 import 'package:nextcloud/utils.dart';
+import 'package:queue/queue.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:timezone/timezone.dart' as tz;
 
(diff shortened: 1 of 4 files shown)


provokateurin (changes requested):

packages/neon_framework/lib/src/utils/request_manager.dart
```diff
@@ -79,15 +85,51 @@ class RequestManager {
...
       return;
     }
+
     if (!subject.hasValue) {
       subject.add(Result.loading());
     }
 
     var request = getRequest();
 
+    final shouldRequest = await _shouldRequest(
```
provokateurin: To me this method sounds side effect free, but it actually loads and emits cached data.

I think you can just inline this logic into the existing functionality in the place where you're returning true.

provokateurin (changes requested):

provokateurin (approved):
Cool, thanks!

provokateurin:
Can you check the test results and try to reproduce locally? I think they could be related.

vauvenal5:
I have to look deeper into this but I believe it is a timing issue. The bigger the delay in the test the more refresh calls pass and the expected amount does not match anymore. I guess my changes lead the more calls passing but not sure yet of the details.

vauvenal5:
Okay I believe I have narrowed it down. There are actually 5 events which call `talkBloc.updateRoom` not 4. 
1) BehaviorSubject.seeded(Result.success(room))
2+3) From the refresh in the constructor
4+5) From the refresh in the test

Due to the timing it does not show up in the test until you set a sleep also after the the refresh call in the test:

```
// The delay is necessary to avoid a race condition with loading twice at the same time
await Future<void>.delayed(const Duration(milliseconds: 1));
await roomBloc.refresh();
await Future<void>.delayed(const Duration(milliseconds: 1));

verify(() => talkBloc.updateRoom(any())).called(4);
```
My change seems to every slightly switch timing making all calls finish before the verify gets triggered in the test.

I refactored the test a bit making the behavior more deterministic.

PR #3185, opened 2026-03-14
feat(neon_framework): optimize blur hash handling for being performan…

vauvenal5:
This PR does multiple blur hash improvements:
* moves blur hash decoding out of the building path because it leads to UI freezes when opening a gallery with a lot of blur hashes
* it moves decoding to a separate scheduler to allow Flutter to decide when to decode
* it adds a result cache for future use to be able to reuse calculated blur hashes, this is especially important when scrolling up and down through a gallery
* fixes UI flickering when replacing blur hash with actual preview result

### DIFF
--- a/packages/neon_framework/lib/src/blocs/blur.dart
+++ b/packages/neon_framework/lib/src/blocs/blur.dart
@@ -1,0 +1,122 @@
+import 'dart:async';
+import 'dart:ui' as ui;
+
+import 'package:cacherine/cacherine.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/scheduler.dart';
+import 'package:flutter_blurhash/flutter_blurhash.dart';
+import 'package:logging/logging.dart';
+import 'package:neon_framework/blocs.dart';
+
+/// A bloc that manages the decoding of blur hashes into images.
+/// It uses a [compute] to offload the decoding process to a separate isolate, as it can be CPU intensive.
+/// Results are cached in a [SimpleLRUCache] so that they can be reused without re-decoding with a limit of 1000 entries.
+class BlurBloc extends Bloc {
+  final _blurHashCash = SimpleLRUCache<String, BlurHashDecodeTask>(1000);
+
+  @override
+  void dispose() {
+    _blurHashCash.clear();
+  }
+
+  @override
+  Logger get log => Logger('BlurBloc');
+
+  /// Gets the decoded image for the given [blurHash] and [size].
+  /// If the image is already cached, it returns the cached [BlurHashDecodeTask].
+  /// If the image is not cached, it creates a new [BlurHashDecodeTask] to decode the blur hash.
+  /// The result of the decoding process is stored in a [ValueNotifier] within the [BlurHashDecodeTask].
+  BlurHashDecodeTask getBlurHash(String blurHash, ui.Size size) {
+    final key = '$blurHash-${size.width}x${size.height}';
+
+    final cachedTask = _blurHashCash.get(key);
+    if (cachedTask != null) {
+      return cachedTask;
+    }
+
+    final task = BlurHashDecodeTask(blurHash: blurHash, size: size);
+    unawaited(task.execute());
+    _blurHashCash.set(key, task);
+    return task;
+  }
+}
+
+/// A task to decode a blur hash into an image. The result is stored in a [ValueNotifier].
+class BlurHashDecodeTask {
+  /// Creates a new [BlurHashDecodeTask] with the given [blurHash] and [size].
+  /// The result is stored in a [ValueNotifier] so that cashed blurs can be displayed without any flickering.
+  BlurHashDecodeTask({
+    required String blurHash,
+    required ui.Size size,
+  }) : _blurHashMeta = _BlurHashComputeTask(
+          blurHash: blurHash,
+          width: size.width.toInt(),
+          height: size.height.toInt(),
+        );
+
+  /// The blur hash to decode.
+  final _BlurHashComputeTask _blurHashMeta;
+
+  /// The result of the decoding process, stored in a [ValueNotifier] so that cashed blurs can be displayed without any flickering.
+  ValueNotifier<ui.Image?> blur = ValueNotifier<ui.Image?>(null);
+
+  /// Executes the task by computing the blur in an isolate and then decoding the resulting pixels into an image.
+  Future<void> execute() async {
+    final pixels = await _computePixels();
+
+    // We are using the scheduler to decode the image just to be on the safe side.
+    await SchedulerBinding.instance.scheduleTask(
+      () => ui.decodeImageFromPixels(
+        pixels,
+        _blurHashMeta.width,
+        _blurHashMeta.height,
+        ui.PixelFormat.rgba8888,
+        (image) => blur.value = image,
+      ),
+      Priority.animation,
+    );
+  }
+
+  /// Computes the pixels for the blur image either by using [SchedulerBinding] for web or [compute] for other platforms
+  /// to offload the decoding process to a separate isolate,
+  /// as it can be CPU intensive and we don't want to block the UI thread.
+  Future<Uint8List> _computePixels() async {
+    if (kIsWeb) {
+      return SchedulerBinding.instance.scheduleTask(
+        () => optimizedBlurHashDecode(
+          blurHash: _blurHashMeta.blurHash,
+          width: _blurHashMeta.width,
+          height: _blurHashMeta.height,
+        ),
+        Priority.animation,
+      );
+    }
+
+    return compute(
+      (blurHashMeta) => optimizedBlurHashDecode(
+        blurHash: blurHashMeta.blurHash,
+        width: blurHashMeta.width,
+        height: blurHashMeta.height,
+      ),
+      _blurHashMeta,
+    );
+  }
+}
+
+/// Helper object to carry the necessary data into the compute isolate.
+class _BlurHashComputeTask {
+  _BlurHashComputeTask({
+    required this.blurHash,
+    required this.width,
+    required this.height,
+  });
+
+  // The blur hash to decode.
+  final String blurHash;
+
+  // The width of the resulting image.
+  final int width;
+
+  // The height of the resulting image.
+  final int height;
+}
--- a/packages/neon_framework/lib/src/widgets/image.dart
+++ b/packages/neon_framework/lib/src/widgets/image.dart
@@ -3,14 +3,15 @@ import 'dart:convert';
 import 'dart:typed_data';
 
 import 'package:flutter/material.dart';
-import 'package:flutter_blurhash/flutter_blurhash.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:http/http.dart' as http;
 import 'package:meta/meta.dart';
 import 'package:neon_framework/models.dart';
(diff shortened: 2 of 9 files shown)


packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,98 @@
...
+import 'package:flutter_blurhash/flutter_blurhash.dart';
+import 'package:logging/logging.dart';
+import 'package:neon_framework/blocs.dart';
+import 'package:queue/queue.dart';
+
+/// A bloc that manages the decoding of blur hashes into images. 
+/// It uses a [Queue] to limit the maximum number of concurrent decoding tasks, 
+/// and a map to cache the results of previously decoded blur hashes so that they can be reused without re-decoding.
+///
+/// Please note that cache clean up is the responsibility of the caller, as the cache is not automatically cleared.
```
provokateurin: Maybe it should be an LRU to avoid excessive memory usage if you have a lot of blurhashes. The app could be open without a restart for hours, days or even weeks, so there must be some safeguard. It could still be some high limit like 1000.

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,98 @@
...
+
+  /// Removes the cached image for the given [blurHash] and [size].
+  void remove(String? blurHash, ui.Size size) {
+    if (blurHash == null) {
+      return;
+    }
+
+    final task = BlurHashDecodeTask(blurHash: blurHash, size: size);
+    if (_blurHashCash[task.key] != null) {
+      _blurHashCash.remove(task.key);
```
provokateurin: Feels a bit overkill to build the task just to get the key from it. Maybe this can be a separate method that doesn't require the task to be built?

packages/neon_framework/lib/src/widgets/image.dart
```diff
@@ -96,44 +98,77 @@ class NeonImage extends StatelessWidget {
...
+
+  /// Replacing the blurhash with the actual image leads to UI flickering when scrolling very fast.
+  /// To mitigate this, we keep the blurhash underneath the actual image.
+  Widget _buildImageWithBlur(BuildContext context, {required Widget child, bool isLoading = true}) => Stack(
+        fit: StackFit.passthrough,
+        children: [
+          _buildBlur(context, isLoading: isLoading),
+          child,
+        ],
+      );
```
provokateurin: Good idea, but sounds like it could be a performance problem at some point. Is there no way we could remove the stack and only render the blurhash if the image is currently invisible?

provokateurin (changes requested):
This is a really nice improvement, thanks!

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:logging/logging.dart';
+import 'package:neon_framework/blocs.dart';
+import 'package:queue/queue.dart';
+
+/// A bloc that manages the decoding of blur hashes into images.
+/// It uses a [Queue] to limit the maximum number of concurrent decoding tasks,
+/// and a LRU-Cache to cache the results of previously decoded blur hashes so that they can be reused without re-decoding.
+class BlurBloc extends Bloc {
+  final _blurHashQueue = Queue(parallel: 10);
+  final _blurHashCash = SimpleLRUCache<String, BlurHashDecodeTask>(1000);
```
provokateurin: What's the amount of memory used per cache entry? I'm not sure if there is a benefit to making the cache any larger, but if there is then finding a good value based on the minimum hardware requirements (which are not defined, but basically some low-end Android device) would be nice.

provokateurin (changes requested):

provokateurin:
> And will improve my workflow, sorry for the chaos.

No problem at all! Usually it's not feasible to run all tests locally for all changes, so catching test failures in CI is totally fine :)

packages/neon_framework/lib/src/blocs/blur.dart
vauvenal5: Honestly, I did not look into it. Similar like with the value above I believe this could be refined down the line, if necessary.

When scrolling though a view with more then 1000 images, like a timeline, I believe it quit reasonable if at some point we start clearing old hashes.

At this point it might be more useful to think along the line of predictive loading instead of adding more to the cache, not sure though.

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,84 @@
...
+import 'package:flutter_blurhash/flutter_blurhash.dart';
+import 'package:logging/logging.dart';
+import 'package:neon_framework/blocs.dart';
+import 'package:queue/queue.dart';
+
+/// A bloc that manages the decoding of blur hashes into images.
+/// It uses a [Queue] to limit the maximum number of concurrent decoding tasks,
+/// and a LRU-Cache to cache the results of previously decoded blur hashes so that they can be reused without re-decoding.
+class BlurBloc extends Bloc {
+  final _blurHashQueue = Queue(parallel: 10);
```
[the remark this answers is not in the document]
vauvenal5: I did test it on a Pixel 6 where it behaves nicely. Not sure though if this qualifies as low-end. I actually kept the count deliberately low to not spam the server.

This could be made configurable down the line.

packages/neon_framework/lib/src/widgets/image.dart
```diff
@@ -108,32 +110,65 @@ class NeonImage extends StatelessWidget {
...
+      key: ValueKey(blurHash),
+      // We are not caching the blurHash result because we do not want to take care of cleanup in here.
+      // If pre-caching is required, the encapsulating widget should take care of it and also of the cleanup.
+      future: blurHash != null
+          ? blurBloc.getBlurHash(
+              blurHash!,
+              size ?? const Size.square(32),
+              cache: false,
+            )
+          : null,
```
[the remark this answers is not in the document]
vauvenal5: I refactored it a little, hope that's better.

packages/neon_framework/lib/src/blocs/blur.dart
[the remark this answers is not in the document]
provokateurin: > I actually kept the count deliberately low to not spam the server.

You're probably confusing topics? This is CPU only and not network or server ;)

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,79 @@
...
+    _blurHashCash.clear();
+  }
+
+  @override
+  Logger get log => Logger('BlurBloc');
+
+  /// Gets the decoded image for the given [blurHash] and [size].
+  /// If the image is already cached, it returns the cached image.
+  /// Otherwise, it creates a new [BlurHashDecodeTask] to decode the blur hash, and returns the result of that task.
+  /// If [cache] is `true`, the result will be cached. [cache] does not have any effect on lookup.
```
provokateurin:
```suggestion
```

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,79 @@
...
+  Future<ui.Image> getBlurHash(String blurHash, ui.Size size) {
+    final key = '$blurHash-${size.width}x${size.height}';
+
+    if (_blurHashCash.get(key) != null) {
+      return _blurHashCash.get(key)!.result.future;
```
provokateurin: Don't call `get()` twice, save it to a variable instead.

provokateurin (changes requested):
Nice! Only two small things to fix and then we can merge it :)

packages/neon_framework/lib/src/blocs/blur.dart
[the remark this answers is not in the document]
vauvenal5: You are absolutely right. 🤦‍♂️ Must have been thinking about the request queue.

Well the testing part stands... I tested it on a Pixel 6 😂 .. I can do a retest with a higher number, probably would be best to make it configurable via settings.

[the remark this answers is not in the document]
provokateurin: I don't think such things should be configurable by the user, so just hardcoding it is fine.

vauvenal5:
So I went back into complete re-test last week. Device is a Pixel 6 with roughly 1200 images in the view. Major changes are:

- BlurHash computation is now done with `compute` which moves it into an isolate.
- FutureBuilder was replaced with a `ValueListenableBuilder` to avoid flickering on fast scrolling
- Queue is not needed anymore.
- Cashing 1000 BlurHashes results ~30KB memory. Measured on 208x208 images in the timeline view with the DEV-Tools.

Result is that scrolling is now smooth :ok_hand: Zero notice when switching from cached to loading.

packages/neon_framework/lib/src/blocs/blur.dart
vauvenal5: I measured it actually with the dev tools, for 1000 entries it burns ~30KB of memory.

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,107 @@
...
+  final _BlurHashComputeTask _blurHashMeta;
+
+  // The result of the decoding process, stored in a [ValueNotifier] so that cashed blurs can be displayed without any flickering.
+  ValueNotifier<ui.Image?> blur = ValueNotifier<ui.Image?>(null);
+
+  // Executes the task by computing the blur in an isolate and then decoding the resulting pixels into an image.
+  Future<void> execute() async {
+    // We are using compute to offload the decoding process to a separate isolate,
+    // as it can be CPU intensive and we don't want to block the UI thread.
+    final pixels = await compute(
```
vauvenal5: Only disadvantage is, `compute` does not work on web platforms. So depending on how hard web support is desired, I would need to add a check here and run it on the SchedulerBinding instead when on web.

Problem with SchedulerBinding on mobile was, that when scrolling you can feel/see a slow down.

provokateurin: Yeah web support should be retained. Either use SchedulerBinding or just to it synchronously as it was done before.

packages/neon_framework/lib/src/blocs/blur.dart
```diff
@@ -0,0 +1,107 @@
...
+          (blurHashMeta) => optimizedBlurHashDecode(
+        blurHash: blurHashMeta.blurHash,
+        width: blurHashMeta.width,
+        height: blurHashMeta.height,
+      ),
+      _blurHashMeta,
+    );
+
+    // We are using the schedular to decode the image just to be on the safe side.
```
provokateurin:
```suggestion
    // We are using the scheduler to decode the image just to be on the safe side.
```

provokateurin (changes requested):

packages/neon_framework/lib/src/blocs/blur.dart
vauvenal5: I moved it into the SchedulerBinding for better performance. By default compute would run it on the main loop on web.

vauvenal5:
Codecov is failing because of a cert issue. I need to look into this. Or, is it some known behavior?

provokateurin:
Never seen before, but we can just ignore it.

provokateurin (approved):

PR #3187, opened 2026-03-15
Fix podman rootless

provokateurin:


provokateurin:
> Should I try if it still works with docker? (would need to wait until tomorrow)

Yes please! I think the CI would fail anyway, but it's still better if you can validate it as well.

provokateurin:
Failing tests were already fixed in https://github.com/nextcloud/neon/pull/3188, this branch was just not rebased.

PR #3194, opened 2026-03-27
chore: Remove nextcloud package

provokateurin:
Moved to https://github.com/provokateurin/nextcloud.dart

PR #3195, opened 2026-03-28
wip: remove cookie_store package

Leptopoda:
similar to: #3194 this removes the cookie_store and related packages from the neon codebase.
The new home is: https://github.com/foldland/cookie_store

This is still a WIP as dependency resolution will not work yet. 
`package:cookie_store` imports `package:cookie_store_conformance_test` from a local path causing dart to hick up.
I'll create tags and use the tag prefix versioning to allow a better dependency resolution.


Leptopoda:
I've used a global override for now because tag patterns require dart 3.9 which is still far away for us :(

provokateurin:
The patch for the downgrade checks needs to be adjusted.

provokateurin (changes requested):
The CI with downgraded dependencies can be removed in this PR now, since all of the packages left in this monorepo won't be published.

